PR #17, opened 2022-09-05
Clean fix static binaries

msooseth:
Fixes static binary creation. The biggest change here is adding BOTH:

```
                "--extra-lib-dirs=${pkgs.glibc}/lib"
                "--extra-lib-dirs=${pkgs.glibc.static}/lib"
```

i.e. both dynamic and static, otherwise some intermediate compiles fail. The final executable will be static, and using the static glibc, of course. But ghci builds an intermediate dynamic library that fails if dynamic glibc is not here. Another important change is adding static library generation NIX file for libsecp256k1 -- the original NIX package does not allow for a static compilation, it's missing an option to add `--enable-static` to the `./configure`.

NOTE: This diff is missing the changes to `hevm-cli.hs`, namely [1], because it's not relevant to the change, though HEVM will not compile without it (currently, compile is broken)

[1]

```
diff --git a/src/hevm/hevm-cli/hevm-cli.hs b/src/hevm/hevm-cli/hevm-cli.hs
index 17237d55..f8e8a1b8 100644
--- a/src/hevm/hevm-cli/hevm-cli.hs
+++ b/src/hevm/hevm-cli/hevm-cli.hs
@@ -506,7 +506,7 @@ assert cmd = do
   preState <- symvmFromCommand cmd
   let errCodes = fromMaybe defaultPanicCodes (assertions cmd)
   withSolvers EVM.SMT.Z3 4 $ \solvers -> do
-    res <- verify solvers preState (maxIterations cmd) (askSmtIterations cmd) rpcinfo Nothing (Just $ checkAssertions errCodes)
+    res <- verify solvers preState (maxIterations cmd) (askSmtIterations cmd) rpcinfo (Just $ checkAssertions errCodes)
     case res of
       [Qed _] -> putStrLn "QED: No reachable property violations discovered"
       cexs -> do
@@ -650,7 +650,8 @@ launchExec cmd = do
   vm <- vmFromCommand cmd
   case optsMode cmd of
     Run -> do
-      vm' <- execStateT (EVM.Stepper.interpret fetcher . void $ EVM.Stepper.execFully) vm
+      vm' <- undefined
+      --vm' <- execStateT (EVM.Stepper.interpret fetcher . void $ EVM.Stepper.execFully) vm
       --when (trace cmd) $ hPutStr stderr (showTraceTree dapp vm')
       case view EVM.result vm' of
         Nothing ->
@@ -682,7 +683,7 @@ launchExec cmd = do
     --Debug -> void $ EVM.TTY.runFromVM Nothing dapp fetcher vm
     --JsonTrace -> void $ execStateT (interpretWithTrace fetcher EVM.Stepper.runFully) vm
     _ -> error "TODO"
-   where fetcher = maybe EVM.Fetch.zero (EVM.Fetch.http block') (rpc cmd)
+   where fetcher = undefined -- maybe undefined (EVM.Fetch.http block') (rpc cmd)
          block'  = maybe EVM.Fetch.Latest EVM.Fetch.BlockNumber (block cmd)
 
 data Testcase = Testcase {
@@ -864,7 +865,7 @@ symvmFromCommand cmd = do
     (Nothing, Just sig') -> do
       method' <- functionAbi sig'
       let typs = snd <$> view methodInputs method'
-      pure $ symCalldata (view methodSignature method') typs (arg cmd) EmptyBuf
+      pure $  undefined -- symCalldata (view methodSignature method') typs (arg cmd) EmptyBuf
     _ -> error "incompatible options: calldata and abi"
 
   -- TODO: rework this, ConcreteS not needed anymore
```


arcz:
@msooseth is there any benefit to copying the nix files instead of overriding the existing derivation to add extra static flags?

msooseth:
@arcz Actually, no -- but I could not figure out how to override, sorry! It's in `stdenv.mkDerivation` so I heard it needs some special magic? I don't seem to have the magic :magic_wand: :disappointed:  However, if you have the :magic_wand: , then I'd be very happy to just do that, instead of what's above :) Let me know if you know how to add the `    "--enable-static"` to the `configureFlags` -- I would actually be really curious how to do that!

flake.nix
```diff
@@ -22,6 +22,7 @@
         # --- packages ----
 
         packages = flake-utils.lib.flattenTree {
+          libsecp256k1 = pkgs.callPackage (import ./nix/libsecp256k1.nix) {};
```
arcz, line 25: `libsecp256k1 = pkgs.secp256k1.overrideAttrs (attrs: { configureFlags = attrs.configureFlags ++ ["--enable-static"]; })`

msooseth: Ah, thanks! Let me try that out :)

arcz:
No problem 😄 I've added a comment on how to do it. I think I'll upstream the option to enable the static build for those libs to nixpkgs so it's easier to use.

msooseth:
Thanks! I did think about upstreaming an option to add that flag, but it's literally the first time I touched nix, so I was like... maybe I don't understand Nix conventions well enough to do that. Upstreaming an option would probably be the right thing to do, though.

flake.nix
msooseth: Done, thanks!

msooseth:
Thanks for the fix @arcz ! I now re-built and validated that build, it indeed work as expected, yay! Thanks, this simplifies the diff quite a bit :)

nix/libff.nix
```diff
@@ -12,7 +12,7 @@ stdenv.mkDerivation rec {
     fetchSubmodules = true;
   };
 
-  cmakeFlags = [ "-DWITH_PROCPS=Off" ];
```
arcz, line 15: hmm, I've checked and this flag doesn't seem to influence anything.

nix/libff.nix
```diff
@@ -12,7 +12,7 @@ stdenv.mkDerivation rec {
...
+  cmakeFlags = [ "-DWITH_PROCPS=OFF -DBUILD_SHARED_LIBS=OFF" ];
```
arcz, line 15: hmm, I've checked and this flag doesn't seem to influence anything.

msooseth: Which flag, actually? I did this change out of caution, and it may not be needed.

The `Off` almost certainly should be changed to `OFF`, not only because of convention, but because of how `if` [works in cmake](https://cmake.org/cmake/help/latest/command/if.html#basic-expressions).

Regarding `-DBUILD_SHARED_LIBS=OFF` -- you re-built it without this change, and the built `hevm` binary is static? Often, in my experience, `-DBUILD_SHARED_LIBS=OFF` helps build static libraries & binaries.

arcz: I meant `-DBUILD_SHARED_LIBS=OFF`

msooseth: I have just checked, (took about 15 mins of build & preparation time, due to `nix-build` restarting from scratch, and I wanted to be sure), but this statement seems incorrect. Removing `-DBUILD_SHARED_LIBS=OFF` makes this reproducibly incorrect on my end, we get a build error.

In case I am wrong, please post full build log here, so I can check.  To get full build long, apply the DIFF in the 1st comment so Main compiles, run `nix-build`, check the `nix log`, take the option set as per `configureFlags:`, remove the `$...` parts, run it as `cabal build ...` under `nix shell` and don't forget to change `--verbose` to `--verbose=3`. Then post the whole build log here. I believe you should get the same build error I am getting, about linking errors.

d-xo (approved):
lgtm

msooseth:
@arcz Hey! Do you have any other feedback, or have you managed to make it compile without `-DBUILD_SHARED_LIBS=OFF`? It doesn't seem to compile for me without :disappointed:

arcz:
@msooseth I didn't have time to check so far, this is not a blocker though. I can take a look into that after it's merged

PR #18, opened 2022-09-05
Simp rules

msooseth:
## Description

These are likely uncontroversial except for the `ReadWord`/`WriteWord` -- however, for the moment, that really makes the generated IR a _lot_ nicer. So I'd like to have it in until I can understand & improve `simplifyReads`.

What do you think?

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -367,6 +367,7 @@ simplify e = if (mapExpr go e == e)
       | a == (Lit 0) = v
       | otherwise = o
     go o@(And a (And b c))
+      | a == c = (And a b)
       | a == b = (And b c)
       | otherwise = o
     go o@(Add a b)
@@ -386,12 +387,6 @@ simplify e = if (mapExpr go e == e)
     -- we write at least 32, so if x < 32, it's TRUE
     go o@(EVM.Types.GT (BufLength (WriteWord {})) x)
       | x < Lit 0x32 = Lit 0x1
-      | otherwise = o
-    go o@(SLT (Lit a) (Lit b))
-      | (a < 128)  && (b < 128)  && (a < b)  = Lit 1
-      | (a < 128)  && (b < 128)  && (a >= b) = Lit 0
-      | (a >= 128) && (b >= 128) && (a < b)  = Lit 0
-      | (a >= 128) && (b >= 128) && (a >= b) = Lit 1
       | otherwise = o
     go o@(Sub a b)
       | a == b = Lit 0
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -379,11 +379,11 @@ simplify e = if (mapExpr go e == e)
       | otherwise = Lit 0
     -- we write at least 32, so if x <= 32, it's FALSE
     go o@(EVM.Types.LT (BufLength (WriteWord {})) x)
-      | x <= Lit 0x32 = Lit 0x0
+      | x <= Lit 32 = Lit 0x0
       | otherwise = o
     -- we write at least 32, so if x < 32, it's TRUE
     go o@(EVM.Types.GT (BufLength (WriteWord {})) x)
-      | x < Lit 0x32 = Lit 0x1
+      | x < Lit 32 = Lit 0x1
       | otherwise = o
     go o@(Sub a b)
       | a == b = Lit 0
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/SymExec.hs
```diff
@@ -364,8 +367,31 @@ simplify e = if (mapExpr go e == e)
...
+    go o@(Add a b)
+      | b == (Lit 0) = a
+      | a == (Lit 0) = b
+      | otherwise = o
+    go o@(EVM.Types.LT (Lit a) (Lit b))
+      | (a < b) = Lit 1
+      | otherwise = Lit 0
+    go o@(IsZero (CallValue x))
+      | x == 0 = Lit 0x1
+      | otherwise = Lit 0x0
```
d-xo, line 382: The `x` in callvalue refers to the index of the frame for that particular instance of `CallValue`, not the value itself. We can probably tidy things up so the frame index isn't required, but as it stands this rule is wrong.

src/hevm/src/EVM/SymExec.hs
```diff
@@ -364,8 +367,31 @@ simplify e = if (mapExpr go e == e)
...
+    -- we write at least 32, so if x <= 32, it's FALSE
+    go o@(EVM.Types.LT (BufLength (WriteWord {})) x)
+      | x <= Lit 0x32 = Lit 0x0
+      | otherwise = o
+    -- we write at least 32, so if x < 32, it's TRUE
+    go o@(EVM.Types.GT (BufLength (WriteWord {})) x)
+      | x < Lit 0x32 = Lit 0x1
```
d-xo, line 389: Shouldn't it be as follows? Or am I misunderstanding?

```suggestion
      | x < Lit 0x32 = Lit 0x0
```

src/hevm/src/EVM/SymExec.hs
msooseth: OOOps, thanks!

src/hevm/src/EVM/SymExec.hs
msooseth: Let's do this step by step so I also understand a bit better :)

1. We write at least 32 bits. So we have ` o@(EVM.Types.GT y x)` where `y`>=32.
2. Simplify to `GT y x`, where `y`>=32.
3. Let's say we have `x` < 32, say 31 (largest `x` possible), and y=32 (smallest `y` possible). Then: `GT 32 31`.
4. Since 32>31, we have `true`, i.e. `0x01` as a result

So I believe this makes this rule correct? I am assuming that `GT a b` is operating as `a>b`.

msooseth:
I have now removed the simplification rule that you noticed is wrong, sorry!

src/hevm/src/EVM/SymExec.hs
d-xo: ah yes of course, I was confused. 

However, `0x32` is not the same as `32`, so the rule is still wrong but for a different reason :sweat_smile:

msooseth: Oh waaaaait!! OOooops. Good catch :100:

msooseth: OK, fixed to 32!

msooseth:
Fixed to 32! Can you please review @d-xo ? I hope there are more issues...

msooseth:
Based on what @ zoep said, actually some of these rules were incorrect, wow. So now they should be fine. This comparison thing is not great indeed, insidious!

src/hevm/src/EVM/SymExec.hs
```diff
@@ -364,8 +367,28 @@ simplify e = if (mapExpr go e == e)
...
+    go o@(Add a b)
+      | b == (Lit 0) = a
+      | a == (Lit 0) = b
+      | otherwise = o
+    go o@(EVM.Types.LT (Lit a) (Lit b))
+      | (a < b) = Lit 1
+      | otherwise = Lit 0
+    -- we write at least 32, so if x <= 32, it's FALSE
+    go o@(EVM.Types.LT (BufLength (WriteWord {})) (Lit x))
+      | x <= 32 = Lit 0x0
```
d-xo, line 382:
```suggestion
      | x <= 32 = Lit 0
```

Just for consistency...

src/hevm/src/EVM/SymExec.hs
```diff
@@ -364,8 +367,28 @@ simplify e = if (mapExpr go e == e)
...
+      | otherwise = o
+    -- we write at least 32, so if x < 32, it's TRUE
+    go o@(EVM.Types.GT (BufLength (WriteWord {})) (Lit x))
+      | x < 32 = Lit 0x1
```
d-xo, line 386:
```suggestion
      | x < 32 = Lit 1
```

d-xo (approved):
fine to merge once CI is green

msooseth:
This cannot be made to work due to being on top of an old `main`. Closing and opening a new one, identical, but on top of new `main`.

PR #19, opened 2022-09-05
Makig formatting possible to use in SymExec

msooseth:
This is not an amazing diff, but it fixes the issue of not being able to use the function `formatExpr` in `SymExec.hs`, which was quite a pain before. I guess this should be fixed better... maybe you can give some advice?

### DIFF
--- a/src/hevm/src/EVM/Format.hs
+++ b/src/hevm/src/EVM/Format.hs
@@ -491,5 +491,3 @@ formatExpr = go
           , ")"
           ]
       a -> show a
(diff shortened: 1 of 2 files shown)


msooseth:
Yep, good point. Let me try to break the cycle instead :)

msooseth:
Oh man, this was trivial, there was no cycle, just had to remove the import.... Sorry. Trivial change now.

d-xo (approved):

PR #23, opened 2022-09-06
Fix builds on macos

d-xo:
This isn't a full fix, but at least it gets rid of the missing libraries errors that we were seeing before. I'm still running into this issue though when building the tests: https://github.com/NixOS/nixpkgs/issues/149937.


d-xo:
ok so I *think* this should fix using cabal on macos for local development inside of the nix-shell. `nix-build` is still not working, but I think we can save that for another PR.

PR #28, opened 2022-09-08
Fixing PNeg p -> p and also a grammar typo in the same function

msooseth:
## Description
Fixes bug with PNeg being forgotten, and also fixes a grammar typo.


d-xo (approved):
thanks!

PR #30, opened 2022-09-08
Better name for parameter

msooseth:
Better name for parameter. It's actually `[(script, models)]` and gets unpacked into `(s, ms)`.


d-xo:
hmmmm, don't love the name, it's longer and the second part isn't really even a model, more like "names of variables that we should extract a model for"

msooseth:
Fair enough :) I'll close!

PR #32, opened 2022-09-08
Fixing up P issues

msooseth:
This fixes what I believe are bugs and makes it easier to make sure this rewrite set is correct.

I believe this is a bug:

```
 PGT l r -> if l > r
             then PBool True
             else PEq l r
```

Notice the `PEq` instead of `PGT` ? I mean we do check if it's bigger syntactically, but we may not be able to determine that, right? Also, it could be smaller, not just equal? So we should return ` PGT l r` right?

Similarly for `PLT` and`PLEq`. Maybe my understanding is wrong though :S Please check.

### DIFF
--- a/src/hevm/test/test.hs
+++ b/src/hevm/test/test.hs
@@ -464,9 +464,10 @@ tests = testGroup "hevm"
             |]
           [Qed res] <- withSolvers Z3 1 $ \s -> checkAssert s defaultPanicCodes c (Just ("deposit(uint8)", [AbiUIntType 8])) []
           putStrLn $ "successfully explored: " <> show (Expr.numBranches res) <> " paths"
-          {-bs <- runSMT $ query $ do
-            (Cex, -) <- checkAssert allPanicCodes c (Just ("deposit(uint8)", [AbiUIntType 8])) []
-              case view (state . calldata . _1) vm of
+          {-
+          bs <- runSMT $ query $ do
+            (Cex _, vm) <- checkAssert allPanicCodes c (Just ("deposit(uint8)", [AbiUIntType 8])) []
+            case view (state . calldata . _1) vm of
               SymbolicBuffer bs -> BS.pack <$> mapM (getValue.fromSized) bs
               ConcreteBuffer _ -> error "unexpected"
 
--- a/src/hevm/test/test.hs
+++ b/src/hevm/test/test.hs
@@ -412,7 +412,7 @@ tests = testGroup "hevm"
                     break;
                   }
                  deposit_count = deposit_count >> 1;
-                }
+                 }
                 assert(found);
               }
              }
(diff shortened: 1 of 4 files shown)


msooseth:
Note, the tests fail because of unexpected successes :laughing:

zoep:
The fix seems right to me, but, more generally, are the semantics of Haskell's ">" what we mean to do here? 

My understanding is that here we want to check if the operands have concrete values and if these values satisfy the predicate. 

This check seems to work: 
```
ghci>  Lit 5 > Lit 3
True
```
but these do not seem like the desired behaviour:
```
ghci> (Add (Lit 0) (Lit 0)) > Lit 3
True
ghci> Var "a" > Lit 3
True
```
Maybe we should unwrap the inner literal and check for inequalities then?

msooseth:
Hi,

Oh wow, what an amazing observation, actually! We should work on fixing it! I think it could be fixed by not relying on `<`. It seem to be defined at `Types.hs`, with `deriving Ord`, and later with `instance Ord Prop where [...]`. It's apparently needed for some `map` usage here and there. If I take it out, I get a bunch of:

```
      In the expression:
        do traceShowM e
           s <- get
           let (_, bs) = bufs s
           case Map.lookup e bs of
             Just (n, _) -> pure . T.pack $ "buf" <> show n
             Nothing -> do ...
    |
502 |     case Map.lookup e bs of
```
However, it's kinda dangerous, actually. I wonder how to _really_ fix it. Maybe we could have a version that derives `Ord` for specifically the places where `map` is used, and everywhere else, such comparison would not be allowed. That would error out in all cases where we compare incorrectly, e.g. the instance you shown and some instances where I think it's definitely misused. That probably is the long-term solutions?

I personally would prefer to have this PR pushed in, and then fix the `Ord` issue, though, rather than making a single big PR for both problems in one go? I see them as separate issues, and the `Ord` one may take longer to figure out, I think? What do you think?

zoep:
Yes, let's fix this in a separate PR! 

Regarding the instance visibility, I looked it up a bit, and it seems that instances are automatically imported and exported so we can't restrict the scope of  `Ord` in modules that don't use `map`. One solution (from here https://www.reddit.com/r/haskell/comments/cc271p/hiding_typeclass_instances/) would be to declare a `newtype` that is a wrapper for `Expr` and does not derive `Ord`. But it not very clean as it requires one more level of pattern matching. Or maybe we should just be careful to avoid the use of `Ord` for `Expr` in the code. (Though I agree that it would be nice to have some way of statically enforcing that)

d-xo:
very nice catch @zoep. I wonder if we could define a custom ord instance that does what we want? Otherwise I agree it seems to be too big of a footgun to leave in, and think we should figure out how to live without being able to key maps with Expr instances....

d-xo:
I'm fine to merge once CI is green here (looks like we just need to remove an `expectFail` from the tests).

src/hevm/test/test.hs
```diff
@@ -398,33 +398,7 @@ tests = testGroup "hevm"
           [Cex _, Cex _] <- withSolvers Z3 1 $ \s -> checkAssert s defaultPanicCodes c (Just ("fun(uint256)", [AbiUIntType 256, AbiUIntType 256])) []
           putStrLn "expected 2 counterexamples found"
         ,
-        expectFail $ testCase "Deposit contract loop (z3)" $ do
```
msooseth, line 401: NOTE: I am removing this because it's a copy-paste of the one below, which has now been fixed, except that this had everything in the code commented out.

PR #36, opened 2022-09-14
Visual Debugger & Solidity Unit Testing Framework

d-xo:
This fixes / comments out a bunch of type errors in UnitTest.hs and TTY.hs meaning that we now have a (semi) working visual debugger and solidity unit testing framework.

There is still quite some work to do to get this working properly, but this is a nice base that we can start to use to debug the various concrete execution bugs introduced during the refactor.

### DIFF
--- a/src/hevm/src/EVM/Dapp.hs
+++ b/src/hevm/src/EVM/Dapp.hs
@@ -62,6 +62,10 @@ makeLenses ''DappContext
 instance Show Test where
   show t = unpack $ extractSig t
 
+isSymbolic :: Test -> Bool
+isSymbolic (SymbolicTest _) = True
+isSymbolic _ = False
+
 dappInfo
   :: FilePath -> Map Text SolcContract -> SourceCache -> DappInfo
 dappInfo root solcByName sources =
--- a/src/hevm/src/EVM/Dapp.hs
+++ b/src/hevm/src/EVM/Dapp.hs
@@ -62,10 +62,6 @@ makeLenses ''DappContext
 instance Show Test where
   show t = unpack $ extractSig t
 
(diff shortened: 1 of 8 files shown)


src/hevm/src/EVM/TTY.hs
```diff
@@ -725,21 +719,21 @@ drawVmBrowser ui =
               (\selected (k, c') ->
                  withHighlight selected . txt . mconcat $
                    [ fromMaybe "<unknown contract>" . flip preview dapp' $
-                       ( dappSolcByHash . ix (view codehash c')
+                       ( dappSolcByHash . ix (maybeHash c)
```
zoep, line 722: Is `c` or `c'` the right contract to use here? Since this is inside the closure folding over `c'`s I would suspect `c'`

src/hevm/src/EVM/Dapp.hs
```diff
@@ -62,6 +62,10 @@ makeLenses ''DappContext
 instance Show Test where
   show t = unpack $ extractSig t
 
+isSymbolic :: Test -> Bool
+isSymbolic (SymbolicTest _) = True
+isSymbolic _ = False
+
```
msooseth, line 68: I don't believe `isSymbolic` is used anywhere? Maybe it's not needed?

src/hevm/src/EVM/TTY.hs
d-xo: nice catch :)

src/hevm/src/EVM/Dapp.hs
d-xo: nice catch :)

PR #37, opened 2022-09-16
Removed dependency to Ord for Expr and minor fix in SMT encoding

zoep:
* As discussed in #32 , the automatically derived Ord does not have the expected semantics. In `evalProp` we used to do
  ```
  o@(PGT l r) -> if l > r
                 then PBool True
                 else o
  ```
  But
  ```
  ghci> (Add (Lit 0) (Lit 0)) > Lit 3
  True
  ```
  therefore that would result to `(PGT (Add (Lit 0) (Lit 0)) (Lit 3))` evaluating to `PBool True`.
  
  Turned these to direct comparisons of the inner literals.
  
  * Removed what seems to be redundant stores in the SMT encoding of `writeWord`.


msooseth (approved):

d-xo (approved):
awesome 😎

PR #38, opened 2022-09-16
Simp rules -- now on top of new `main`

msooseth:
## Description

These are likely uncontroversial except for the `ReadWord`/`WriteWord` -- however, for the moment, that really makes the generated IR a _lot_ nicer. So I'd like to have it in until I can understand & improve `simplifyReads`.

This new PR is needed because the old PR (#18 ) was on top of an old `main` and would never give us a green light.


msooseth:
Actually, this is not needed

d-xo (approved):
nice :)

PR #39, opened 2022-09-16
Panic code tests

msooseth:
## Description
This PR adds tests for each panic code. All seem to work, except for things that cannot currently be handled by HEVM due to missing symbolic representation: too much memory allocation, and function variable calls.

Hence, this tests all panic codes except `0x22: If you access a storage byte array that is incorrectly encoded.`. This is marked as a todo.

Let me know what you think! I think it's a valuable "unit" test for each of the panic codes.


d-xo (approved):
nice!

PR #41, opened 2022-09-19
SMT encoding bug fixes

zoep:
Fixes a couple of bugs that was causing `safeAdd` to return spurious counter examples.

- `writeWord` stored the least significant bytes of a word in small addresses whereas `readWord` expected the most significant bytes of words in small addresses. For example,  `writeWord  offset (_ bv3 256) buf)` would store `0x03` in `buff[offset]` and `readWord offset buf` would have `0x03` as the most significant byte. As a result, `assert (not (= val (readWord (_ bv0 256) (writeWord (_ bv0 256) val buf)))))` would be `sat`. 

  Now `writeWord` stores the most significant bytes in the smaller addresses.
  
- The encoding of `copySlice srcOff dstOff size src dst` was mixing up `srcOff` and `dstOff` parameters.
  
- The encoding of `copySlice` had on off-by-one error in the treating of the size parameter. For example, copying 32 bytes `CopySlice (Lit 0x80) (Lit 0x0) (Lit 0x20)  buf1 buf2` would generate 33 `store`s and `select`s.
  
 
After these changes the `pop-empty-array` tests loops when it is interpreted to `Expr`. This test has a loop that depends on a symbolic input. This is likely because the interpreter was doing some smt queries that now return a different result.


src/hevm/test/test.hs
```diff
@@ -260,27 +260,28 @@ tests = testGroup "hevm"
             |]
         [Cex _] <- withSolvers Z3 1 $ \s -> checkAssert s [0x21] c (Just ("fun(uint256)", [AbiUIntType 256])) []
         putStrLn "expected counterexample found"
-     ,
+     -- ,
      -- TODO 0x22 is missing: "0x22: If you access a storage byte array that is incorrectly encoded."
      -- TODO below should NOT fail
-     testCase "pop-empty-array" $ do
```
d-xo, line 266: can we use `expectFail` here instead of commenting

d-xo (approved):
Just a lil nit, but apart from that looks awesome. Really nice finds ❤️

src/hevm/test/test.hs
zoep: Because the test doesn't not fail but it loops during `interpret`, expectFail also never terminates

d-xo: I think we can use `ignoreTest` in this case instead. It's a bit nice to keep even ignored tests typechecking at least

zoep: Oh nice! I'll do that!

d-xo (approved):
lgtm! :sparkles:

PR #43, opened 2022-09-22
Code simplifications in EVM.SMT

arcz:
## Description
Just a quick rewrite of `concatBytes` and `writeBytes` in `EVM.SMT` to get rid of `error` calls.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/SMT.hs
+++ b/src/hevm/src/EVM/SMT.hs
@@ -25,10 +25,12 @@ import Data.Char (isSpace)
 import Data.Containers.ListUtils (nubOrd)
 import Control.Monad.State.Strict
 
+import qualified Data.ByteString as BS
 import qualified Data.List as List
+import Data.List.NonEmpty (NonEmpty((:|)))
+import qualified Data.List.NonEmpty as NE
 import Data.String.Here
 import Data.Map (Map)
-import Data.Maybe
 import qualified Data.Map as Map
 import Data.Text (Text)
 import qualified Data.Text as T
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/SMT.hs
```diff
@@ -824,24 +824,20 @@ expandExp base expnt
...
-  eBuf <- exprToSMT buf
-  eIdx <- exprToSMT idx
-  eByte <- exprToSMT b
-  pure $ "(store " <> eBuf `sp` eIdx `sp` eByte <> ")"
-writeBytes idx (hd : tl) buf = do
-  eHd <- exprToSMT hd
-  eIdx <- exprToSMT idx
-  eTl <- writeBytes (add idx (Lit 1)) tl buf
-  pure $ "(store " <> eTl `sp` eIdx `sp` eHd <> ")"
+writeBytes :: [Expr Byte] -> State BuilderState Text
```
d-xo, line 835: Not sure if I'm reading this write, but it seems less general than the old `writeBytes`, which could write a region of bytes at an arbitrary index onto an arbitrary buffer? This version seems to just write them onto an `EmptyBuf`?

arcz: Right, I didn't know how general this should be. I can add an argument back to `writeBytes` and pass an `EmptyBuf` at the call site.

arcz: Changed

d-xo (approved):

d-xo:
Thanks!

PR #47, opened 2022-10-04
Enable more tests

zoep:
This enables all remaining commented tests, modulo the  Equivalence checking testing group. Most of the tests don't work so they require `expectFailure`.  The code that is inspecting counterexamples in few tests is still commented out since we don't have proper cex parsing yet. 

Also, tests that are duplicates for different solvers are deleted, since we plan to run tests with multiple solvers later.


zoep:
Closing to create a new one on top of new main.

PR #50, opened 2022-10-04
Moving to cvc5 and fixing it up

msooseth:
## Description
This fixes using cvc, now using cvc5. There was a bug in the `prelude` -- lines cannot be cut by spec, but Z3 allows them to be cut, so this bug was not caught. The new `cvc5` requires a different option (`--no-interactive`) to get a "clean" interactive shell. Tests are NOT fixed up in this PR, because that would conflict with @zoep 's work on fixing up tests.

NOTE: you *must* `rm -rf dist-newstyle` in order for your `cabal v2-repl` to correctly find `cvc5`. Unfortunately, `cabal v2-repl` caches your `$PATH` and if it changes, it will not be reflected next time your run `cabal v2-repl`. So `cvc5` will not be found. See https://github.com/haskell/cabal/issues/2015 for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Available options:
                            60000)
   --max-iterations INTEGER Number of times we may revisit a particular branching
                            point
-  --solver TEXT            Used SMT solver: z3 (default) or cvc4
+  --solver TEXT            Used SMT solver: z3 (default) or cvc5
   --smtdebug               Print smt queries sent to the solver
   --assertions [WORD256]   Comma seperated list of solc panic codes to check for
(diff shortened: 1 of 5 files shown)


d-xo:
nice! Our encoding is still not compatible with cvc5, but this is good by itself

msooseth:
True, but it's not 100% unusable -- some things do work! :) I removed a test from here because it would conflict with the work of Zoe, but some tests do pass with CVC5. Once Zoe is done with her test fix-up, I can add some CVC5 tests, too, to demonstrate where it works.

PR #53, opened 2022-10-05
[WIP] Add LSB-MSB i.e. endianness tests

msooseth:
## Description
These tests, once Cex generation is enabled, should allow us to understand LSB/MSB issues


msooseth:
@d-xo what do you think, should we merge this?

PR #54, opened 2022-10-05
IDX cannot be larger than 31 here, make it undefined

msooseth:
## Description
Since `idx` should never be more than 31 here, let's make it `undefined`, so we get an error in case of some problems.

### DIFF
--- a/src/hevm/src/EVM/Expr.hs
+++ b/src/hevm/src/EVM/Expr.hs
@@ -521,7 +521,7 @@ indexWord (Lit idx) (JoinBytes zero        one        two       three
   | idx == 29 = twentynine
   | idx == 30 = thirty
   | idx == 31 = thirtyone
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/Expr.hs
```diff
@@ -521,7 +521,7 @@ indexWord (Lit idx) (JoinBytes zero        one        two       three
   | idx == 29 = twentynine
   | idx == 30 = thirty
   | idx == 31 = thirtyone
-  | otherwise = LitByte 0
+  | otherwise = undefined
```
d-xo, line 524: Lets use `error` instead here. Smth like this:

```suggestion
  | otherwise = error $ "Internal error: idx must be <= 31 (actual: " <> (show . num $ idx) <> ")"
```

PR #57, opened 2022-10-18
Minor typo in safemath 

zoep:
Fixed contract name in safemath test and now the test passes. On a side note, the test takes a lot of time to prove (~85'' in my computer) and the expr IR has 46 branches.


d-xo (approved):
Nice catch. Do you know how many branches are being checked for assertion violations? I wonder if we can get a speed up by checking multiple branches in a single query?

PR #58, opened 2022-10-19
Dapp tests

d-xo:
## Description

Adds the test cases for the ds-test runner into the tasty testing framework. Based on top of https://github.com/ethereum/hevm/pull/56 since there were a lot of nice fixes there that made a bunch of tests pass.

Still some failures to deal with here tho...


d-xo:
This PR is getting pretty huge, and has a ton of useful little fixes in it by now. I suggest merging and we can fix the remaining failures in seperate smaller PRs.

src/hevm/test/test.hs
```diff
@@ -1269,3 +1319,115 @@ bothM f (a, a') = do
...
+    , EVM.UnitTest.testParams = params
+    , EVM.UnitTest.dapp = srcInfo
+    , EVM.UnitTest.ffiAllowed = True
+    }
+
+compileWithDSTest :: FilePath -> IO (Text, Text)
+compileWithDSTest src =
+  withSystemTempFile "input.json" $ \file handle -> do
+    hClose handle
+    dsTest <- readFile =<< Paths.getDataFileName "test/contracts/lib/test.sol"
```
d-xo, line 1382: right now we use a vendored version of ds-test. It would be nice if we could take a path via env var or something that would let us test against an arbitrary ds-test commit...

src/hevm/src/EVM/Types.hs
```diff
@@ -244,21 +236,18 @@ data Expr (a :: EType) where
 
   -- logs
 
-  EmptyLog       :: Expr Logs
-
-  Log            :: Expr EWord         -- address
+  LogEntry       :: Expr EWord         -- address
```
d-xo, line 239: I made a small change to the way that we represent logs here, so we just use a normal haskell list of `LogEntry`s, which made working with the logs in `Format.hs` much easier.

d-xo:
Summary of changes:

- Integrate the solidity code from dapp-tests into the tasty test runner
- add a repl helper for debugging these tests into test.hs
- fix the SAR instruction for concrete cases
- fix a few off by one / argument ordering errors in Expr.hs
- "fix" the calculation of the `v` paramater when signing (or just document better the hack we were using)
- add simplifier rules to deal with function selector checks
- simplify the smt encoding for GEq and LEq
- some qol / devx improvements in the debugger
- add constraints on the max possible size for calldata
- uncomment & adapt a bunch of log formatting code

src/hevm/src/EVM/SymExec.hs
```diff
@@ -347,41 +351,68 @@ simplify e = if (mapExpr go e == e)
...
+    -- concrete LT / GT
+    go (EVM.Types.LT (Lit a) (Lit b))
+      | a < b = Lit 1
+      | otherwise = Lit 0
+    go (EVM.Types.GT (Lit a) (Lit b))
+      | a > b = Lit 1
+      | otherwise = Lit 0
+
+    -- syntactic Eq reduction
+    go (Eq (Lit a) (Lit b))
```
msooseth, line 373: I would think/hope this is redundant given the one below? But maybe there is a reason why it's needed, sorry, it's a bit unclear to me. I thought the stuff below should catch this?

src/hevm/src/EVM/TTY.hs
```diff
@@ -945,27 +946,36 @@ drawTracePane s =
...
+        <=> hBorderWithLabel (txt "Returndata")
+        <=> ourWrap (prettyIfConcrete (view (state . returndata) vm))
+        <=> hBorderWithLabel (txt "Output")
+        <=> ourWrap (maybe "" show (view result vm))
+        <=> hBorderWithLabel (txt "Cache")
+        <=> ourWrap (show (view (cache . path) vm))
+        <=> hBorderWithLabel (txt "Path Conditions")
+        <=> (ourWrap $ show $ view constraints vm)
+        <=> hBorderWithLabel (txt "Memory")
+        <=> (ourWrap (prettyIfConcrete (view (state . memory) vm)))
```
msooseth, line 961: On this note, we should merge Zoe's pending changes and auto-indent everything :) Seems like there is a mix of indentation going on sometimes.

src/hevm/src/EVM/Types.hs
```diff
@@ -571,7 +559,7 @@ foldExpr f acc expr = acc <> (go expr)
         <> (go b)
         <> (go c)
         <> (go d)
-        <> (go g)
+        <> (foldl (<>) mempty (fmap go g))
```
msooseth, line 562: It'd be nice to know what the parameters to `Create` are. Right now, `a b c d g h` is a bit obscure, so the change on `g` is obscure :S Can we change `Create a b c ...` to  `Create offs sz mem logs sto addr` ? Then I'd get the change on `logs` :)

src/hevm/test/contracts/pass/dsProvePass.sol
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    function prove_add(uint x, uint y) public {
+        unchecked {
+            if (x + y < x) return; // no overflow
+            assertTrue(x + y >= x);
+        }
+    }
+
+    //function prove_balance(address usr, uint amt) public {
```
msooseth, line 34: Were these committed commented-out by accident? It's useful to comment out during debugging, so I'm thinking this may have been accidental?

src/hevm/src/EVM/Expr.hs
```diff
@@ -153,9 +153,8 @@ shr = op2 SHR (\x y -> if x > 256 then 0 else shiftR y (fromIntegral x))
 
 sar :: Expr EWord -> Expr EWord -> Expr EWord
 sar = op2 SAR (\x y ->
-  let msb = testBit y 255
-      shifted = if x > 256 then 0 else shiftR y (fromIntegral x)
-  in if msb then setBit shifted 255 else shifted)
+  let asSigned = (fromIntegral y) :: Int256
+  in fromIntegral $ shiftR asSigned (fromIntegral x))
```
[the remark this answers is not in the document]
d-xo: The `shiftR` from `Data.Bits` does a logical shift on unsigned types and an arithmetic shift on signed types, so I just convert `y` to be an `Int256` instead of a `Word256` and use the SAR implementation built in to the `Data.Bits` package.

src/hevm/src/EVM/Types.hs
d-xo: Yeah these traversals are a pretty obscure in general, but I'm not sure how important it is to assign a distinct name to the params here, since it's more about the structure of the expressions than the actual meaning of each variable?

src/hevm/src/EVM/SymExec.hs
d-xo: So the one below applies to all `Eq` expressions, and in that case we don't do any simplifications in the case where `a` is not the exact same expression as `b` (since doing so would require some smt calls or custom decision procedure). 

The case here applies to `Lit`s only and then we can simplify unequal `Lit`s down to `Lit 0`.

src/hevm/src/EVM/Expr.hs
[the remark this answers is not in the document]
msooseth: Ah, I see. Sorry.

src/hevm/src/EVM/SymExec.hs
msooseth: Ahhh, it's for `Lit 0`. I see, sorry.

src/hevm/src/EVM/Types.hs
msooseth: Yeah, it's true that it would just be a maintenance nightmare, and it's pretty easy to look it up.

src/hevm/test/contracts/pass/dsProvePass.sol
msooseth: Ah, I see. Sorry, I was afraid this was unintentional.

msooseth:
So this can be merged I guess?

zoep (approved):

PR #59, opened 2022-10-20
Expr: correct oob case for symbolic branch in indexWord

d-xo:
## Description

Returning 0 on oob is actually the correct behaviour here: https://www.evm.codes/#1a?fork=arrowGlacier

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/Expr.hs
+++ b/src/hevm/src/EVM/Expr.hs
@@ -532,7 +532,7 @@ indexWord (Lit idx) (JoinBytes zero        one        two       three
   | idx == 29 = twentynine
   | idx == 30 = thirty
   | idx == 31 = thirtyone
-  | otherwise = error $ "Internal error: idx must be <= 31 (actual: " <> (show . num $ idx) <> ")"
+  | otherwise = LitByte 0
 indexWord idx w = IndexWord idx w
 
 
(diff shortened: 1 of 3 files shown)


msooseth:
I'd put a test case in just in case. BYTE should return 0, so we could have a test case like:

```
        testCase "check-asm-byte-oob" $ do
          Just c <- solcRuntime "C"
            [i|
            contract C {
              function foo(uint256 x) external pure {
                uint256 y;
                assembly { y := byte(32,x) }
                assert(y == 0);
              }
            }
            |]
          [Qed res] <- withSolvers Z3 1 $ \s -> checkAssert s defaultPanicCodes c (Just ("foo(uint256)", [AbiUIntType 256])) []
          putStrLn $ "Yes, we only get 0 back for any input"
        ,
```
We could even go fancy, and have `foo(uint256 x, uint256 x2)` and have an `if (x2 >= 32) { assembly { y := byte(x2,x)...}`, thereby proving it's correct for all oob access.

What do you think?

msooseth (changes requested):
NICE CATCH! I'd like to have a test case as I mentioned in the comments, if that's OK? It should just copy-pate and work out of the box, I think? Let me know what you think!

zoep (approved):

d-xo:
Added a few more tests and also included an SMT encoding for symbolic word indexing. I let the existing concrete code paths in place since they resulted in a more direct encoding in SMT, but maybe it's worth simplifying the encoding logic at the expense of a more complex encoding.

msooseth (approved):
Looking good!

d-xo:
I think tests should pass here once the shl pr is merged

PR #64, opened 2022-11-01
Adding SIGNEXTEND

msooseth:
## Description
Adding sign extension OP. I am using a macro, as discussed. However, we could still use structural hashing here, to extract out the value and use it only once.

As per: https://www.evm.codes/#0b?fork=arrowGlacier

## Checklist

- [x] tested locally
- [x]  added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (commented):
looks good! just the indentation in the tests....

msooseth:
Ayyy sorry! I didn't realize what was wrong at first and fixed a minor indentation issue. I think I now realized what you meant. Can you check again? Sorry :(

src/hevm/src/EVM/SMT.hs
```diff
@@ -254,6 +254,41 @@ prelude =  (flip SMT2) mempty $ fmap (T.drop 2) . T.lines $ [i|
   (declare-const chainid Word)
   (declare-const basefee Word)
 
+  ; macros
+  (define-fun signext ( (sz (_ BitVec 256)) (val (_ BitVec 256))) (_ BitVec 256) (
+     ite (= sz (_ bv0  256)) (_ bv0 256)
```
d-xo, line 259: This seems incorrect according to whast we observed on the evm.codes playground

src/hevm/src/EVM/SMT.hs
```diff
@@ -254,6 +254,41 @@ prelude =  (flip SMT2) mempty $ fmap (T.drop 2) . T.lines $ [i|
   (declare-const chainid Word)
   (declare-const basefee Word)
 
+  ; macros
+  (define-fun signext ( (sz (_ BitVec 256)) (val (_ BitVec 256))) (_ BitVec 256) (
```
d-xo, line 258: We actually should be extending by `sz + 1` (weirdly)....

msooseth: Sorry. Fixed :) Chris said the reason is: "we extend byte X" I.e. it's not "size", but "the byte to extend". I now renamed it to "b" rather than "size", since it's more appropriate.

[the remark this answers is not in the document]
msooseth: Thanks, this has now been fixed :)

d-xo (approved):
Looks good!

PR #66, opened 2022-11-03
Fixing SHL and adding a test case

msooseth:
## Description
The arguments to `SHL` were reversed I believe. I added a test case that demonstrates this and fixed the order. The test fails with the old setup.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/test/test.hs
+++ b/src/hevm/test/test.hs
@@ -485,7 +485,25 @@ tests = testGroup "hevm"
             |]
         [Qed _] <- withSolvers Z3 1 $ \s -> checkAssert s defaultPanicCodes c (Just ("fun(int256,int256)", [AbiIntType 256, AbiIntType 256])) []
         putStrLn "SAR works as expected"
-      ,
+     ,
+     testCase "opcode-shl" $ do
+        Just c <- solcRuntime "MyContract"
+            [i|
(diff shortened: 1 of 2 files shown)


src/hevm/test/test.hs
```diff
@@ -485,7 +485,25 @@ tests = testGroup "hevm"
...
+              require(val < (1<<16));
+              require(shift_by < 16);
+              require(shift_by >= 0);
+              uint256 out;
+              assembly {
+                out := shl(shift_by,val)
+              }
+              assert (out >= val);
+              }
+             }
```
d-xo, line 503: indentation

src/hevm/test/test.hs
```diff
@@ -485,7 +485,25 @@ tests = testGroup "hevm"
...
-      ,
+     ,
+     testCase "opcode-shl" $ do
+        Just c <- solcRuntime "MyContract"
+            [i|
+            contract MyContract {
+              function fun(uint256 shift_by, uint256 val) external pure {
+              require(val < (1<<16));
+              require(shift_by < 16);
+              require(shift_by >= 0);
```
d-xo, line 496: I think this is redundant?

PR #70, opened 2022-11-09
Common subexpression elimination pass for Expr

zoep:
Add a common subexpression elimination pass (CSE) for buffer and storage expression at the Expr level and removes the corresponding pass from the SMT encoding. 

- Add a `GVar a s` node to the `Expr` AST that represents global variables referring to a global environment. `a` is a unique id and `s` is a string that represents the (unique) variable name that gets printed out when we generate the encoding. 
- Add `mapExprM` and `mapPropM` generic traversals that map monadic computations over `Expr` and `Prop`. These are moved to a new Traversals.hs file along with preexisting traversals.
- Common subexpression elimination pass is expressed using the new traversal functions.
- Plug CSE into symbolic executor. Because of some static evaluation that happens during this phase (e.g., `simplify` and evaluation of pre and post condition), CSE has to happen at the very last step before SMT encoding, after evaluation of pre and postcondition. 
- [This](https://gist.github.com/zoep/6b89b9b722f14debf4919b4893d21676) is an example of CSE in the flattened expression of the safeAdd test.


src/hevm/src/EVM/Types.hs
```diff
@@ -123,13 +123,18 @@ data EType
...
+  deriving (Show, Eq, Ord)
+
 -- add type level list of constraints
 data Expr (a :: EType) where
 
   -- identifiers
 
   Lit            :: W256 -> Expr EWord
   Var            :: Text -> Expr EWord
+  GVar           :: GVar a -> Text -> Expr a
```
d-xo, line 137: curious why we need an id and a stringified name here?

src/hevm/src/EVM/Traversals.hs
```diff
@@ -0,0 +1,846 @@
...
+  PGT a b -> PGT (mapExpr f (f a)) (mapExpr f (f b))
+  PLEq a b -> PLEq (mapExpr f (f a)) (mapExpr f (f b))
+  PGEq a b -> PGEq (mapExpr f (f a)) (mapExpr f (f b))
+  PNeg a -> PNeg (mapProp f a)
+  PAnd a b -> PAnd (mapProp f a) (mapProp f b)
+  POr a b -> POr (mapProp f a) (mapProp f b)
+
+-- | Recursively applies a given function to every node in a given expr instance
+-- Recursion schemes do this & a lot more, but defining them over GADT's isn't worth the hassle
+mapExpr :: (forall a . Expr a -> Expr a) -> Expr b -> Expr b
```
d-xo, line 249: Did you play around with implementing this in terms of `mapExprM`?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -406,7 +408,7 @@ reachableQueries = go []
           (go (PEq (Lit 1) c : pcs) t)
           (go (PEq (Lit 0) c : pcs) f)
         pure (tres <> fres)
-      _ -> pure [assertProps pcs]
+      _ -> pure [assertProps pcs mempty mempty]
```
d-xo, line 411: It's a little clunky to have to pass `mempty` twice to `checkSat'` / `assertProps` and friends. I wonder if there's some helper we can define that we expose as the main API from `EVM.SMT`?

src/hevm/src/EVM/CSE.hs
```diff
@@ -0,0 +1,110 @@
...
+-- | Common subexpression elimination pass for Expr
+eliminateExpr' :: Expr a -> State BuilderState (Expr a)
+eliminateExpr' e = mapExprM go e
+
+eliminateExpr :: Expr a -> (Expr a, BufEnv, StoreEnv)
+eliminateExpr e =
+  let (e', st) = runState (eliminateExpr' e) initState in
+  (e', invertKeyVal (snd (bufs st)), invertKeyVal (snd (stores st)))
+  where
+    invertKeyVal =  Map.fromList . map (\(x, y) -> (Id y, x)) . Map.toList
```
d-xo, line 91: maybe worth pulling this out into a top level function since we have it duplicated a few times in the `where` blocks?

src/hevm/src/EVM/Types.hs
d-xo: Yeah I think I would prefer either of the above two. I think both are probably fine. I think you  should be able to use [`eqT`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Typeable.html#v:eqT)  for type level reflection. 

There is an example in act [here](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Typeable.html#v:eqT)

src/hevm/src/EVM/CSE.hs
```diff
@@ -0,0 +1,110 @@
...
+initState = BuilderState
+  { bufs = (0, Map.empty)
+  , stores = (0, Map.empty)
+  }
+
+
+go :: Expr a -> State BuilderState (Expr a)
+go = \case
+  -- buffers
+  e@(WriteWord i v b) -> do
```
msooseth, line 41: According to GHC, this can be replaced with `WriteWord {}`. Similarly below for `WriteByte,` `CopySlice`, and `SStore`. The `v` here shadows the `v` below in `Just v`, which confused me at first.

zoep: Good point!

src/hevm/src/EVM/SymExec.hs
zoep: Done!

src/hevm/src/EVM/CSE.hs
```diff
@@ -0,0 +1,110 @@
+{-# Language DataKinds #-}
+
+{- |
+    Module: EVM.CSE
+    Description: Common subexpression elimination for Expr ast
+-}
+
+module EVM.CSE where
```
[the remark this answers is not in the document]
zoep: Done!

src/hevm/src/EVM/CSE.hs
zoep: Done!

d-xo (approved):
Lookgs good!

msooseth (approved):
I only had these two small nitpicks. Also, reading this I realized we have no tests on bitwise XOR. I'll add that now :)

src/hevm/src/EVM/Types.hs
zoep: I ended up implementing the second bullet

PR #73, opened 2022-11-09
Change gas to Word64 to improve performance

arcz:
## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/hevm/src/EVM.hs
```diff
@@ -1822,20 +1828,16 @@ notStatic continue = do
 -- | Burn gas, failing if insufficient gas is available
 -- We use the `Integer` type to avoid overflows in intermediate
 -- calculations and throw if the value won't fit into a uint64
```
d-xo, line 1830: comment is outdated re: Integer type

d-xo (commented):
looks good, one tiny nit about an outdated comment.

Curious to see numbers on the perf improvement here...

PR #74, opened 2022-11-11
Options for `checkAssert` + Test for `CopySlice` and bitwise `OR`

msooseth:
## Description
Added a test for `CopySlice` that required an option to be added to `checkAssert` so it can accept an option for `simplify` being disabled. I also refactored the related functions to extract out `maxIter` and `askSmtIters` into the option to be passed in.

I tested the `debug` flag manually to print debug information. Note that `maxIters` and `askSmtIters` were actually not used by `verify`, weirdly enough.

I'd really appreciate feedback on this PR, because I'm new to Haskell so I lack knowledge of best-practices. In particular, I'm not so sure about `If'`and how the default options should be passed. 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -273,7 +273,7 @@ type Postcondition = VM -> Expr End -> Prop
 
 
 checkAssert :: SolverGroup -> [Word256] -> ByteString -> Maybe (Text, [AbiType]) -> [String] -> VeriOpts -> IO [VerifyResult]
-checkAssert solvers errs c signature' concreteArgs opts = verifyContract' solvers c signature' concreteArgs opts SymbolicS Nothing (Just $ checkAssertions errs)
+checkAssert solvers errs c signature' concreteArgs opts = verifyContract solvers c signature' concreteArgs opts SymbolicS Nothing (Just $ checkAssertions errs)
 
 {- |Checks if an assertion violation has been encountered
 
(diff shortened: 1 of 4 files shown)


src/hevm/src/EVM/SymExec.hs
```diff
@@ -261,7 +272,10 @@ type Precondition = VM -> Prop
 type Postcondition = VM -> Expr End -> Prop
 
 checkAssert :: SolverGroup -> [Word256] -> ByteString -> Maybe (Text, [AbiType]) -> [String] -> IO [VerifyResult]
-checkAssert solvers errs c signature' concreteArgs = verifyContract solvers c signature' concreteArgs SymbolicS Nothing (Just $ checkAssertions errs)
+checkAssert a b c d e  = checkAssert' a b c d e defaultVeriOpts
```
d-xo, line 275: can we  keep the meaningful arg names here?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -559,14 +577,21 @@ evalProp = \case
                    _ -> o
   o -> o
 
+if' :: Bool -> IO () -> IO ()
```
d-xo, line 580: I think we can just use `when` instead of this?

src/hevm/src/EVM/SymExec.hs
d-xo: Do we even need this variant at all? Can we just construct a `VeriOpts` and pass that in whereever we call `checkAssert`?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -300,9 +314,13 @@ panicMsg :: Word256 -> ByteString
 panicMsg err = (selector "Panic(uint256)") <> (encodeAbiValue $ AbiUInt 256 err)
 
 verifyContract :: SolverGroup -> ByteString -> Maybe (Text, [AbiType]) -> [String] -> StorageModel -> Maybe Precondition -> Maybe Postcondition -> IO [VerifyResult]
-verifyContract solvers theCode signature' concreteArgs storagemodel maybepre maybepost = do
+verifyContract solvers theCode signature' concreteArgs storagemodel maybepre maybepost =
```
d-xo, line 317: Same question as for `checkAssert`. Do we need this variant?

src/hevm/src/EVM/SymExec.hs
msooseth: Good point, actually. Fixed now :)

src/hevm/src/EVM/SymExec.hs
msooseth: Good point, sorry. I didn't want to change all the lines of code in `test.hs`, but it was not a good strategy. We can fix them up differently if they are ugly, e.g. by writing a helper function in `test.hs`.

msooseth:
I have now also added a test for bitwise `OR`

d-xo (approved):
Looks good!

msooseth:
Thanks :heart:  I merged!

PR #75, opened 2022-11-14
Adding LEq and GEq to simplify, rewriting `GT` and `GEq`, improving `ITE`

msooseth:
## Description
I was surprised that `simplify (LEq (Lit 6) (Lit 7))` didn't simplify down to `Lit 1` while `simplify (LT (Lit 5) (Lit 6))` did. So I added the required simplification rules.

I am also changing the logic for `GT` -- I don't think we should not have the same logic for both `LT` and `GT`. I think it's cleaner to just do `LT`, and then rewrite `GT` to `LT`. This is what I have seen in other rewriting systems. If we come up with some cool rewrite for some form of `LT`, we won't have to do it for `GT`, it'll work automatically. Same for `GEq`. In the long run, this could save us bugs and lines of code, I hope?

Further, `ITE` is TRUE for any `Lit x` where `x!=0`. In other words, not only `Lit 1` is True but also e.g. `Lit 5`. This should allow for more simplifications.

TODO: `SLT` ` and `SGT`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/hevm/src/EVM/SymExec.hs
```diff
@@ -365,9 +365,11 @@ simplify e = if (mapExpr go e == e)
     go (EVM.Types.LT (Lit a) (Lit b))
       | a < b = Lit 1
       | otherwise = Lit 0
-    go (EVM.Types.GT (Lit a) (Lit b))
-      | a > b = Lit 1
+    go (EVM.Types.GT a b) = EVM.Types.LT b a
```
d-xo, line 368: Shouldn't this rewrite to `LEq`?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -365,9 +365,11 @@ simplify e = if (mapExpr go e == e)
...
+    go (EVM.Types.LEq (Lit a) (Lit b))
+      | a <= b = Lit 1
       | otherwise = Lit 0
+    go (EVM.Types.GEq a b) = EVM.Types.LEq b a
```
d-xo, line 372: Shouldn't this rewrite to `LT`?

src/hevm/src/EVM/SymExec.hs
msooseth: What I am trying to say I think is that `a < b` is the same as `b > a`. But actually...
```
a<b == !(!(a<b))
a<b == !(a>=b)
```

So then we could rewrite all 4 to just `LT`. All of `LT`, `GT`, `LEq`, `GEq` would become LT. Let me do that, actually. Maybe more sane? Then we'd have:

* `GT`->`LT` via symmetry
* `GEq`->`LT` via negation (as above)
* `LEq`->`GEq` via symmetry

What do you think?

d-xo: ah yes, I wasn't thinking properly. I agree that rewriting everything to LT seems nice.

msooseth: No, it was great, now it's all LT, yay :)

src/hevm/src/EVM/SymExec.hs
```diff
@@ -413,15 +416,42 @@ simplify e = if (mapExpr go e == e)
...
+      | otherwise = t
+    go (ITE (Or a b@(Lit _)) t f) = ITE (Or b a) t f
+
     -- we write at least 32, so if x <= 32, it's FALSE
     go o@(EVM.Types.LT (BufLength (WriteWord {})) (Lit x))
       | x <= 32 = Lit 0
       | otherwise = o
     -- we write at least 32, so if x < 32, it's TRUE
-    go o@(EVM.Types.GT (BufLength (WriteWord {})) (Lit x))
+    go o@(EVM.Types.LT (Lit x) (BufLength (WriteWord {})))
```
msooseth, line 448: Note: we replace all LT with GT, so this simplification rule had to be adjusted.

src/hevm/src/EVM/SymExec.hs
```diff
@@ -414,12 +412,26 @@ simplify e = if (mapExpr go e == e)
       | a == b = (And b c)
       | otherwise = o
 
+    -- go (And (Lit x) b)
+    --   | x == 0 = Lit 0
+    --   | otherwise = b
+    -- go (And a (Lit x))
+    --   | x == 0 = Lit 0
+    --   | otherwise = a
```
d-xo: Can we remove this commented code?

msooseth: Hmmm I fixed this yesterday, maybe I forgot to push? It's fixed now :)

src/hevm/test/test.hs
```diff
@@ -467,6 +467,46 @@ tests = testGroup "hevm"
...
+             uint256 a = (x | 8);
+             bool w;
+             assembly {
+                 w:=a
+             }
+             assert(w);
+           }
+         }
+         |]
+       -- should find a counterexample
```
d-xo, line 507: comment is wrong here

src/hevm/src/EVM/SymExec.hs
```diff
@@ -430,11 +430,8 @@ simplify e = if (mapExpr go e == e)
       | x == 0 = a
       | otherwise = o
 
-    -- Non-bitwise AND & OR -- no need to preserve bitwise equivalence.
-    go (ITE (And (Lit x) a) t f)
-      | x == 0 = f
-      | otherwise = ITE a t f
-    go (ITE (And a b@(Lit _)) t f) = ITE (And b a) t f
+    -- Non-bitwise OR -- no need to preserve bitwise equivalence.
```
d-xo, line 433: I actually find the comment here a little confusing, the Or is still bitwise. Maybe smth like this?
```suggestion
    -- If x is ever non zero the Or will always evaluate to some non zero value and the false branch will be unreachable
```

src/hevm/test/test.hs
msooseth: :S Fixed!

PR #76, opened 2022-11-15
ConcreteStore encoding & readStorage fix

d-xo:
Just a few small fixes:

1. Adds an SMT encoding for `ConcreteStore`
2. Makes `readStorage` a little smarter so that it can handle more kinds of fully concrete reads from semi abstract stores

### DIFF
--- a/src/hevm/src/EVM/Expr.hs
+++ b/src/hevm/src/EVM/Expr.hs
@@ -447,9 +447,14 @@ readStorage addr loc store@(ConcreteStore s) = case (addr, loc) of
     pure $ Lit val
   _ -> Just $ SLoad addr loc store
 readStorage addr' loc s@AbstractStore = Just $ SLoad addr' loc s
-readStorage addr' loc s@(SStore addr slot val prev) = case (addr, slot, addr', loc) of
-  (Lit _, Lit _, Lit _, Lit _) -> if loc == slot && addr == addr' then Just val else readStorage addr' loc prev
+readStorage addr' loc s@(SStore addr slot val prev) = case addr of
+  Lit _ -> if addr == addr'
+           then case slot of
+                  Lit _ -> if slot == loc then (Just val) else readStorage addr' loc prev
+                  _ -> Just $ SLoad addr' loc s
+           else readStorage addr' loc prev
   _ -> Just $ SLoad addr' loc s
+readStorage _ _ (GVar _) = error "Can't read from a GVar"
 
 readStorage' :: Expr EWord -> Expr EWord -> Expr Storage -> Expr EWord
 readStorage' addr loc store = case readStorage addr loc store of
--- a/src/hevm/src/EVM/Expr.hs
+++ b/src/hevm/src/EVM/Expr.hs
@@ -447,12 +447,13 @@ readStorage addr loc store@(ConcreteStore s) = case (addr, loc) of
     pure $ Lit val
   _ -> Just $ SLoad addr loc store
 readStorage addr' loc s@AbstractStore = Just $ SLoad addr' loc s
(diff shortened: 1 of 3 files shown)


src/hevm/src/EVM/Expr.hs
```diff
@@ -441,9 +441,14 @@ readStorage addr loc store@(ConcreteStore s) = case (addr, loc) of
...
   _ -> Just $ SLoad addr loc store
 readStorage addr' loc s@AbstractStore = Just $ SLoad addr' loc s
-readStorage addr' loc s@(SStore addr slot val prev) = case (addr, slot, addr', loc) of
-  (Lit _, Lit _, Lit _, Lit _) -> if loc == slot && addr == addr' then Just val else readStorage addr' loc prev
+readStorage addr' loc s@(SStore addr slot val prev) = case addr of
+  Lit _ -> if addr == addr'
+           then case slot of
+                  Lit _ -> if slot == loc then (Just val) else readStorage addr' loc prev
+                  _ -> Just $ SLoad addr' loc s
+           else readStorage addr' loc prev
```
zoep, line 449: I feel that you can return `readStorage add' loc prev` only if both addresses are different literals. The above would return `readStorage addr' loc prev` skipping a potential relevant store if the addresses are syntactically different but semantically the same (for example one of them is `Lit 0x0` and the other `ReadWord idx (WriteWord idx (Lit 0x0) buf)` or some symbolic variable). The same goes for slot.

d-xo: nice catch, should be fixed now I think.

zoep (approved):

msooseth (changes requested):
I'd add a test case for `encodeConcreteStore`:

```
, testCase "encodeConcreteStore-overwrite" $
      let
        w :: Int -> W256
        w x = W256 $ EVM.Types.word256 $ BS.pack [fromIntegral x]
      in
      assertEqual ""
        (EVM.SMT.encodeConcreteStore $
          Map.fromList [(w 1, (Map.fromList [(w 2, w 99), (w 2, w 100)]))])
        "(sstore (_ bv1 256) (_ bv2 256) (_ bv100 256) emptyStore)"
```

It seems to miss a test currently, the above should address it. I wrote it because I couldn't quite understand what that code did :S

msooseth:
Here's the full diff:

```
diff --git a/src/hevm/test/test.hs b/src/hevm/test/test.hs
index da8a0229..8bc574ea 100644
--- a/src/hevm/test/test.hs
+++ b/src/hevm/test/test.hs
@@ -11,6 +11,7 @@ import System.IO.Temp
 import System.Process (readProcess)
 import GHC.IO.Handle (hClose)
 import Control.Monad
+import Data.Char (ord)
 
 import Prelude hiding (fail)
 
@@ -146,6 +147,15 @@ tests = testGroup "hevm"
         -- same as above, but with offset 2
         (LitByte 0xbb)
         (Expr.indexWord (Lit 2) (Lit 0xff22bb4455667788990011223344556677889900112233445566778899001122))
+    , testCase "encodeConcreteStore-overwrite" $
+      let
+        w :: Int -> W256
+        w x = W256 $ EVM.Types.word256 $ BS.pack [fromIntegral x]
+      in
+      assertEqual ""
+        (EVM.SMT.encodeConcreteStore $
+          Map.fromList [(w 1, (Map.fromList [(w 2, w 99), (w 2, w 100)]))])
+        "(sstore (_ bv1 256) (_ bv2 256) (_ bv100 256) emptyStore)"
     , testCase "indexword-oob-sym" $ assertEqual ""
         -- indexWord should return 0 for oob access
         (LitByte 0x0)
```

msooseth (approved):
Looks good to me :)

PR #77, opened 2022-11-16
CSE: declare all extracted vars in dependency order

d-xo:
Orders all intermediate buffers and stores by dependency order (i.e. if a store references a buffer that references a store, they will all be declared in the correct order in the resulting SMT query).

This is achieved by:

1. using a shared counter for both buffers and stores in CSE
2. conducting CSE from the leaves upwards (ensuring subexpressions have lower indentifiers)
3. using these identifiers to order the declaration of the intermediate bufs / stores in the produced SMT expressions


zoep (approved):
Looks great!

PR #80, opened 2022-11-17
Rewrites for `Read/WriteWord`, `Read/WriteByte`, and `CopySlice`

msooseth:
## Description
Rewrites for `Read/WriteWord`, `Read/WriteByte`, and `CopySlice`. Also, I moved `octet*` to `Types.hs`, it belongs there  more, I think, plus I need it :)

I am not attached to this PR, I'd just like get some feedback :)

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/Expr.hs
+++ b/src/hevm/src/EVM/Expr.hs
@@ -392,8 +392,8 @@ instance Monoid (Expr Buf) where
 -- | Removes any irrelevant writes when reading from a buffer
 simplifyReads :: Expr a -> Expr a
 simplifyReads = \case
-  ReadWord (Lit idx) b -> ReadWord (Lit idx) (stripWrites idx (idx + 31) b)
-  ReadByte (Lit idx) b -> ReadByte (Lit idx) (stripWrites idx idx b)
+  ReadWord (Lit idx) b -> readWord (Lit idx) (stripWrites idx (idx + 31) b)
+  ReadByte (Lit idx) b -> readByte (Lit idx) (stripWrites idx idx b)
   a -> a
 
 -- | Strips writes from the buffer that can be statically determined to be out of range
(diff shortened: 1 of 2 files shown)


src/hevm/src/EVM/Expr.hs
```diff
@@ -392,8 +392,18 @@ instance Monoid (Expr Buf) where
...
+    in case a of
+      (ReadWord (Lit idx2) (ConcreteBuf buf)) ->
+        let unpacked = Data.List.take 32 $ Data.List.drop (fromIntegral $ toInteger idx2) $ BS.unpack buf
+        in Lit (W256 $ word256 $ BS.pack unpacked)
+      x -> x
+  ReadByte (Lit idx) b ->
+    let a = ReadByte (Lit idx) (stripWrites idx idx b)
+    in case a of
+      (ReadByte (Lit idx2) (ConcreteBuf buf)) -> LitByte (BS.unpack buf !! (fromIntegral $ toInteger idx2))
+      x -> x
```
zoep, line 406: This simplification of reading bytes and words from concrete buffs is also achieved by `readBytes` and `readWords`. Maybe we could apply these instead?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -369,6 +371,31 @@ simplify e = if (mapExpr go e == e)
       | otherwise = o
     go o@(ReadWord (Lit _) _) = Expr.simplifyReads o
     go o@(ReadByte (Lit _) _) = Expr.simplifyReads o
+    go (WriteWord (Lit locLit) (Lit valLit) (ConcreteBuf buf)) = ConcreteBuf (BS.pack $ begin ++ middle ++ end)
```
zoep, line 374: Could we reuse `writeWord`, `writeByte` and `copySlice` from Expr.hs here?

msooseth: Oh damn, good point, I didn't notice these...

d-xo (approved):
Looks good!

src/hevm/src/EVM/Expr.hs
msooseth: This has been moved to the things you recommended, thanks!

PR #81, opened 2022-11-17
Simplifier fuzzing

d-xo:
## Description

Adds some small fuzzing pipelines for the simplification engine. This works as follows:

1. generate a random Expr
2. run the simplifier on this Expr
3. compare the two for equivalence using an SMT solver

This is a draft for now because we need to add some solver timeouts before we can actually start to run this in CI. Still the pipelines seem pretty useful, I already found quite a few small bugs with them...

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/SMT.hs
+++ b/src/hevm/src/EVM/SMT.hs
@@ -651,7 +651,7 @@ data CheckSatResult
   | Unsat
   | Unknown
   | Error Text
-  deriving (Show)
+  deriving (Show, Eq)
 
 isSat :: CheckSatResult -> Bool
 isSat (Sat _) = True
--- a/src/hevm/src/EVM/SMT.hs
+++ b/src/hevm/src/EVM/SMT.hs
@@ -24,7 +24,9 @@ import Control.Monad.State.Strict
 import Language.SMT2.Parser (getValueRes, parseFileMsg)
 import Data.Either
 import Data.Maybe
+import Data.Word
 import Numeric (readHex)
+import Data.ByteString (ByteString)
 
 import qualified Data.ByteString as BS
 import qualified Data.List as List
@@ -533,7 +535,7 @@ exprToSMT = \case
   ReadByte idx src -> op2 "select" src idx
 
   ConcreteBuf "" -> "emptyBuf"
-  ConcreteBuf bs -> writeBytes (LitByte <$> BS.unpack bs) mempty
+  ConcreteBuf bs -> writeBytes bs mempty
   AbstractBuf s -> s
   ReadWord idx prev -> op2 "readWord" idx prev
   BufLength b -> op1 "bufLength" b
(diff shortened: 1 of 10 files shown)


src/hevm/src/EVM/SMT.hs
```diff
@@ -465,7 +465,7 @@ exprToSMT = \case
   Min a b ->
     let aenc = exprToSMT a
         benc = exprToSMT b in
-    "(ite (<= " <> aenc `sp` benc <> ") " <> aenc `sp` benc <> ")"
+    "(ite (bvule " <> aenc `sp` benc <> ") " <> aenc `sp` benc <> ")"
```
d-xo, line 468: encoding here was using integer Leq instead of the bitvector variant, produced a type error when passed to the smt solver

src/hevm/src/EVM/SMT.hs
```diff
@@ -740,6 +740,9 @@ withSolvers solver count cont = do
                   mempty (calldataV cexvars)
               pure $ Sat $ SMTCex
                 { calldata = calldatamodels
+                , storage = mempty
```
d-xo, line 743: these missing fields were triggering a runtime exception if we got `sat` back.

src/hevm/src/EVM/SymExec.hs
```diff
@@ -377,7 +377,7 @@ simplify e = if (mapExpr go e == e)
       | otherwise = Lit 0
     go (EVM.Types.GT a b) = EVM.Types.LT b a
     go (EVM.Types.GEq a b) = EVM.Types.LEq b a
-    go (EVM.Types.LEq a b) = EVM.Types.Not (EVM.Types.GT a b)
+    go (EVM.Types.LEq a b) = EVM.Types.IsZero (EVM.Types.GT a b)
```
d-xo, line 380: Bitwise not is not the same as logical not. This rule was incorrect.

src/hevm/src/EVM/Traversals.hs
```diff
@@ -128,7 +128,7 @@ foldExpr f acc expr = acc <> (go expr)
       e@(GasLimit) -> f e
       e@(ChainId) -> f e
       e@(BaseFee) -> f e
-      e@(BlockHash _) -> f e
+      e@(BlockHash a) -> f e <> (go a)
```
d-xo, line 131: we didn't recurse into the arguments for `BlockHash` in `foldExpr`

src/hevm/test/test.hs
```diff
@@ -1468,6 +1497,182 @@ instance Arbitrary RLPData where
...
+
+genNat :: Gen Int
+genNat = fmap fromIntegral (arbitrary :: Gen Natural)
+
+genName :: Gen Text
+genName = fmap T.pack $ listOf1 (oneof . (fmap pure) $ ['a'..'z'] <> ['A'..'Z'])
+
+genWord :: Int -> Gen (Expr EWord)
+genWord 0 = frequency
+  [ (1, fmap Lit arbitrary)
```
d-xo, line 1539: word generation is biased to generate a `Lit` more often than other constructors. I think this better reflects the kind of expressions we actually build when executing EVM, and makes the simplifier more likely to fire, so the tests aren't just a no-op....

src/hevm/src/EVM/SMT.hs
msooseth: Sorry, my mistake.

src/hevm/src/EVM/SymExec.hs
msooseth: Woooooow. WHAT JUST HAPPENED. Mind blown. So sorry!

d-xo: haha, I've made that same mistake many times before

msooseth (approved):
Looks good to me! :)

PR #86, opened 2022-11-18
SymExec: rework verify debugging output

d-xo:
## Description

Tweaks the debugging output in verify a little so it outputs everything as a file, which makes it a little easier to work with (especially if there are many smt queries output).

I also removed a few pieces of output that I generally don't find super useful (`canViolate`, and `leaves`). I'm happy to add them back if you want @msooseth @zoep ?

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -57,6 +57,9 @@ data VeriOpts = VeriOpts
 defaultVeriOpts :: VeriOpts
 defaultVeriOpts = VeriOpts { simp = True, debug = False, maxIter = Nothing, askSmtIters = Nothing }
 
+debugVeriOpts :: VeriOpts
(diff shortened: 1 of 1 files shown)


src/hevm/src/EVM/SymExec.hs
```diff
@@ -57,6 +57,9 @@ data VeriOpts = VeriOpts
 defaultVeriOpts :: VeriOpts
 defaultVeriOpts = VeriOpts { simp = True, debug = False, maxIter = Nothing, askSmtIters = Nothing }
 
+debugVeriOpts :: VeriOpts
+debugVeriOpts = VeriOpts { simp = True, debug = True, maxIter = Nothing, askSmtIters = Nothing }
```
msooseth, line 61: Sorry, should have added that already.

msooseth (approved):
lgtm

PR #87, opened 2022-11-18
Keccak Min

d-xo:
## Description

This adds assumptions that constrain the output of each keccak invocation to be greater than 50. This is a probabalistic assumption, and it is needed to avoid spurious counterexamples when exploring contracts containing mappings or dynamic arrays in storage. If keccak would ever return e.g. 0, then a read from `balanceOf` in the following contract could overlap with `totalSupply` (and vice versa). 

I tested this against `keccak-inj` with `main` merged on and almost all of the tests in `dsProvePass` are working once these assumptions are added.

```solidity
contract Token {
    uint public totalSupply;
    mapping public (address => uint) balanceOf;
}
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
sorry this should be a draft, will rebase on main and merge once https://github.com/ethereum/hevm/pull/85 is in...

src/hevm/src/EVM/Keccak.hs
```diff
@@ -0,0 +1,81 @@
...
+  let (_, st) = runState (findKeccakPropsExprs ps bufs stores) initState in
+  fmap minProp (Set.toList (keccaks st))
+  where
+    minProp :: Expr EWord -> Prop
+    minProp k@(Keccak _) = PGT k (Lit 50)
+    minProp _ = error "Internal error: expected keccak expression"
+
+-- | Takes a list of Props, finds all Keccak occurences and generates
+-- Keccak injectivity assumptions for all unique pairs of Keccak calls
+keccakInj :: forall a. [Prop] -> [Expr Buf]  -> [Expr Storage] -> [Prop]
```
zoep, line 73: Maybe we can have a function that generates both keccak injectivity and min assertions to avoid computing the set of keccak calls twice

zoep (commented):
Nice addition on top of injectivity!

src/hevm/src/EVM/Keccak.hs
d-xo: done

zoep (approved):
Looks great!

PR #90, opened 2022-11-23
Handle symbolic constructor arguments and immutables in ds-test

d-xo:
## Description

A bunch of small fixes and optimizations that were needed to get the `prove_constructorArgs` and `prove_transfer` tests in `dsProvePass.sol` to pass.

I also made a small refactor and moved the simplification logic into `Expr`. Originally I did this because I thought that we needed to run some intermediate simplifciations in `EVM.hs` (and moved it to avoid a cyclic module dependency). In the end this wasn't needed, but I kinda prefer it there, so I left it in.

I also added a fuzz pipeline for `toList` and `fromList` in `Expr` (`toList` was completely broken :see_no_evil:).

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/hevm/src/EVM/Expr.hs
```diff
@@ -503,11 +616,19 @@ isLitByte :: Expr Byte -> Bool
...
 
+isPower2 :: W256 -> Bool
+isPower2 n = n .&. (n-1) == 0
+
 -- | Returns the byte at idx from the given word.
 indexWord :: Expr EWord -> Expr EWord -> Expr Byte
+indexWord (Lit idx) (And (Lit mask) w)
+  | mask == (2 ^ (256 :: W256)) - 1 = indexWord (Lit idx) w -- we need this case here to avoid overflow below
+  | isPower2 (mask + 1), 2 ^ (idx + 1) <= mask = indexWord (Lit idx) w
+  | isPower2 (mask + 1), 2 ^ (idx + 1) > mask = LitByte 0
```
d-xo, line 627: These let us read into bitmasked words.

src/hevm/src/EVM/Expr.hs
```diff
@@ -221,6 +223,12 @@ readBytes (Prelude.min 32 -> n) idx buf
 
 -- | Reads the word starting at idx from the given buf
 readWord :: Expr EWord -> Expr Buf -> Expr EWord
+readWord idx (WriteWord idx' val buf)
+  | idx == idx' = val
+  | otherwise = readWord idx buf
```
zoep: maybe we should check if `idx < idx' || idx > idx' + 31` in order to entirely skip this write? If addresses are not word-aligned `idx` could be in the middle of the word being written

src/hevm/src/EVM/Expr.hs
```diff
@@ -265,6 +273,9 @@ copySlice (Lit srcOffset) (Lit dstOffset) (Lit size) (ConcreteBuf src) (Concrete
         tl = BS.drop (num dstOffset + num size) dst
     in ConcreteBuf $ hd <> sl <> tl
 
+-- copying 32 bytes can be rewritten to a WriteWord on dst (e.g. CODECOPY of args during constructors)
+copySlice srcOffset dstOffset (Lit 32) src dst = WriteWord dstOffset (readWord srcOffset src) dst
```
zoep, line 296: We can also return `writeWord dstOffset (readWord srcOffset src) dst` here, that will perform some concrete execution if possible. When `copySlice` is used `simplify` it does matter, because this will get simplified in an other iteration, but it may matter if we ever use a standalone `copySlice`

zoep (commented):
Looks great! Just a couple of comments in Expr

d-xo:
This branch includes a few commits that it shouldn't, I'll rebase to fix....

src/hevm/src/EVM/Expr.hs
d-xo: done

PR #91, opened 2022-11-24
Keccak equations for concrete values

zoep:
## Description
When there is a Keccak call with concrete arguments in the source code, it gets computed during symbolic evaluation and we can't tell where the value came from. 

For example, in the encoding of this contract
```
contract C {
  function kecc(uint x) public pure {
    if (x == 0) {
      assert(keccak256(abi.encode(x)) == keccak256(abi.encode(0)));
    }
  }
}
```
We can't prove that these two values are equal, because keccak is an uninterpreted function and the `keccak256(abi.encode(0)))` has been replace with its actual value.

To avoid this, we track all concrete keccak computations and we accumulate equations of the form `someConcreteHashValue = Keccak(ConcreteBuf bs)`, so that we also know what keccak call produced this value. Then we can prove correctness of the above contract.  

## Checklist

- [X] tested locally
- [X] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
Looks great!

PR #92, opened 2022-11-29
Optimize SMT Generation

d-xo:
## Description

Concatenation of strict Text instances is O(n+m) in time, and additionally requires allocation of an entirely new piece of memory for each concat operation. This could be observed during profiling, where text concatenation during smt generation often dominated total runtime.

This change optimizes smt generation by switching to lazt `Text` in `SMT.hs`, and using the [Builder API](https://hackage.haskell.org/package/text-2.0.1/docs/Data-Text-Lazy-Builder.html). Concatenation of Builders is `O(1)` and requires significantly less allocation. There is a single `O(n)` pass to convert the `Builder` into a lazy `Text` instance once the full SMT expression has been constructed.

I also tweaked a little our solver interaction code to use the IO facilities built into the Text library and avoid any unescesary `unpack` operations.

I benchmarked this change by running the `buffer-simplificaiton` fuzz pipeline (with a fixed seed & max `size` param in copySlice of 1000) and observed the following change:

**new**:

46.3s
87% z3
13% hevm

**old**:

166.5s
77% hevm
23% z3

When benchmarking with even larger maximum values for the CopySlice `size` parameter (10,000 bytes), I noticed that the bottlneck now appears to be AST construction within the solver itself (~25% of runtime), suggesting that optimization for larger `CopySlice` expressions would require linking against the sovler and constructing SMT AST instances directly within hevm (https://github.com/ethereum/hevm/issues/82).

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/Dev.hs
+++ b/src/hevm/src/EVM/Dev.hs
@@ -15,6 +15,8 @@ import System.Directory
 import Data.String.Here
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
+import qualified Data.Text.Lazy as TL
+import qualified Data.Text.Lazy.IO as TL
 
 import EVM
 import EVM.SMT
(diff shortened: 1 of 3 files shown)


msooseth (approved):
Looks like a pretty clean change, it's amazing it gives so much performance boost! Very cool!

d-xo:
There is another interesting looking package [here](https://hackage.haskell.org/package/text-builder-linear) that should be even faster than the `Builder` interface built into text, but it's currently marked as broken in nixpkgs, and it doesn't seem worth the effort to fix the build in nixpkgs right now.

arcz:
@d-xo I fixed `text-builder-linear` https://github.com/NixOS/nixpkgs/pull/203590

d-xo:
oh, that's awesome. tysm ❤️

PR #106, opened 2022-12-08
Equivalence Check + Few other things

msooseth:
## Description
This is a rather big diff. The changes included are:
* Equivalence check working. Returns `[]` if nothing failed. Otherwise, returns a set of timed out queries and `Cex`-es. So you are only sure all is good if return is `[]`. We could return the `Qed`-s too. Let me know what you think.
* Tests for equivalence checks via YUL tests from Solidity. This necessitated adding Solidity as a submodule
* Tests for equivalence checks via hand-written tests
* Fixing `Div` -- division by zero is zero, but in SMT it's 0xfff...
* Adding `Mod`, `MulMod`, `SMod`, `SDiv` (division by zero thing applies) in SMT. Note that `MulMod` is expensive.
* Adding tests for `MulMod`, `Div`
* Adding end states for: StackLimitExceeded, InvalidMemoryAccess, BadJumpDestination. TODO PLEASE REVIEW. I have no clue whether these are valid end states!
* Fixing timeout for `CVC5`. It doesn't accept the SMT option in the input file.
* Fixing SMT encoding for `*`
* Fixed `concatbytes`. It didn't have a "last node", so concatBytes with 2 bytes was `(concat X1 (concat X2))` instead of `(concat X1 X2)`. CVC5 complained of wrong SMT syntax, Z3 somehow worked (but not sure if correctly).
* Removed 2 unused iteration counting code lines (was confusing)
* Added a note about symbolic exponents -- it's possible but needs symbolic gas
* Fixing cabal file so it uses 6 cores when available. This should speed up compile times significantly. I chose 6, because most people have at least 4 cores, and it will not run out of memory.
* Added `time` and `regex` Haskell modules, so I can time the YUL examples (useful), and generate them via Haskell (instead of `sed` via `nix`)

A LOT of this work was thanks to @d-xo helping me

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/Traversals.hs
+++ b/src/hevm/src/EVM/Traversals.hs
@@ -70,6 +70,9 @@ foldExpr f acc expr = acc <> (go expr)
       -- control flow
 
       e@(Invalid a) -> f e <> (foldl (foldProp f) mempty a)
+      e@(StackLimitExceeded a) -> f e <> (foldl (foldProp f) mempty a)
+      e@(InvalidMemoryAccess a) -> f e <> (foldl (foldProp f) mempty a)
+      e@(BadJumpDestination a) -> f e <> (foldl (foldProp f) mempty a)
       e@(SelfDestruct a) -> f e <> (foldl (foldProp f) mempty a)
       e@(IllegalOverflow a) -> f e <> (foldl (foldProp f) mempty a)
       e@(Revert a b) -> f e <> (foldl (foldProp f) mempty a) <> (go b)


src/hevm/hevm.cabal
```diff
@@ -108,7 +108,7 @@ library
   autogen-modules:
     Paths_hevm
   ghc-options:
-    -Wall -Wno-deprecations -Wno-unticked-promoted-constructors
```
arcz, line 111: Why 6?

src/hevm/hevm.cabal
```diff
@@ -108,7 +108,7 @@ library
...
+    -Wall -Wno-deprecations -Wno-unticked-promoted-constructors -j6
```
arcz, line 111: Why 6?

d-xo: I think we can just use `-j` and then it will figure out how many cores to use based on the machine that it's executing on?

src/hevm/hevm-cli/hevm-cli.hs
```diff
@@ -389,31 +389,30 @@ findJsonFile Nothing = do
...
+      bytecodeB = hexByteString "--code" . strip0x $ codeB cmd
+      veriOpts = VeriOpts { simp = True
+                            , debug = False
+                            , maxIter = maxIterations cmd
+                            , askSmtIters = askSmtIterations cmd
+                          }
+  -- maybeSignature <- case sig cmd of
+  --   Nothing -> return Nothing
+  --   Just sig' -> do method' <- functionAbi sig'
+                    -- return $ Just (view methodSignature method', snd <$> view methodInputs method')
```
d-xo, line 403: do we need this?

src/hevm/hevm.cabal
msooseth: Ah, new feature! Finally they don't refuse to guess the number of processors :D This has been bugging me for about 10 years, but finally, they did fix it! It used to be that `-j` launched as many processes as there were compile items... so it'd crash immediately on large projects. Finally fixed! Yep, OK, I am moving to `-j`, apparently it now works:

`
Using the [-j[⟨n⟩]](https://downloads.haskell.org/ghc/latest/docs/users_guide/using.html#ghc-flag--j[%E2%9F%A8n%E2%9F%A9]) flag, you can compile modules in parallel. Specify -j ⟨n⟩ to compile ⟨n⟩ jobs in parallel. If ⟨n⟩ is omitted, then it defaults to the number of processors.
`
OK, I'm fixing this :)

src/hevm/src/EVM/SymExec.hs
```diff
@@ -36,14 +37,16 @@ import qualified Data.Text.Lazy.IO as TL
...
-  deriving (Show)
+  deriving (Show, Eq)
 type VerifyResult = ProofResult (Expr End) (Expr End, SMTCex) (Expr End)
-type EquivalenceResult = ProofResult ([VM], [VM]) VM ()
 
 isQed :: ProofResult a b c -> Bool
 isQed (Qed _) = True
 isQed _ = False
 
+containsA :: ProofResult () () () -> [(a , b, ProofResult () () ())] -> Bool
```
d-xo, line 47:
```suggestion
containsA :: ProofResult a b c -> [(d, e, ProofResult a b c)] -> Bool
```

I think this can be made a bit more general?

src/hevm/hevm.cabal
msooseth: Fixed to `-j`. Thanks for the feedback!

src/hevm/test/test.hs
```diff
@@ -1550,36 +1662,401 @@ tests = testGroup "hevm"
...
           }
           |]
-        runSMTWith z3 $ query $ do
-          Cex _ <- equivalenceCheck aPrgm bPrgm Nothing Nothing Nothing
+        withSolvers Z3 3 Nothing $ \s -> do
+          a <- equivalenceCheck s aPrgm bPrgm defaultVeriOpts Nothing
+          assertBool "Must have a difference" (not (null a))
+      ,
+      testCase "eq-sol-exp-qed" $ do
+        -- These yul programs are not equivalent: (try --calldata $(seth --to-uint256 2) for example)
```
d-xo, line 1695: This comment is wrong I think?

src/hevm/test/test.hs
```diff
@@ -1550,36 +1662,401 @@ tests = testGroup "hevm"
...
+          |]
+        withSolvers Z3 3 Nothing $ \s -> do
+          a <- equivalenceCheck s aPrgm bPrgm defaultVeriOpts Nothing
+          assertEqual "Must have no difference" [] a
           return ()
-          -}
-
+      ,
+      testCase "eq-sol-exp-cex" $ do
+        -- These yul programs are not equivalent: (try --calldata $(seth --to-uint256 2) for example)
```
d-xo, line 1722: this one too

src/hevm/test/test.hs
```diff
@@ -1550,36 +1662,401 @@ tests = testGroup "hevm"
...
+            False -> do
+              print $ "OK. Took " <> (show $ diffUTCTime end start) <> " seconds"
+              let timeouts = filter (\(_, _, c) -> c == EVM.SymExec.Timeout()) res
+              unless (null timeouts) $ putStrLn $ "But " <> (show $ length timeouts) <> " timeout(s) occurred"
+            True -> do
+              putStrLn $ "Not OK: " <> show f <> " Got: " <> show res
+              error "Was NOT equivalent, error"
+         )
+    -- TODO run, or see how to run https://github.com/ethereum/hevm/blob/main/nix/hevm-tests/yul-equivalence.nix
+    -- convert from nix to Haskell
```
d-xo, line 2059: I think we can remove this comment?

d-xo (commented):
Looks awesome! Just a few little nits.

src/hevm/hevm-cli/hevm-cli.hs
msooseth: I actually checked this, and I think it was some kind of reminder that we didn't implement something and I though it may be important to keep around. I likely misremembered. I am removing.

src/hevm/test/test.hs
msooseth: Yep, thanks, removing!

src/hevm/test/test.hs
msooseth: Oooops, sorry.

src/hevm/test/test.hs
msooseth: yep, removing.

src/hevm/src/EVM/SymExec.hs
msooseth: Good point, thanks, fixed!

src/hevm/src/EVM/Traversals.hs
```diff
@@ -323,6 +323,15 @@ mapExprM f expr = case expr of
   IllegalOverflow a -> do
     a' <- mapM (mapPropM f) a
     f (IllegalOverflow a')
+  InvalidMemoryAccess a -> do
```
zoep, line 326: Shall we add these cases to `foldExpr` in the same file?

msooseth: Ah, good catch, I didn't see it! Fixed in new push :)

d-xo (approved):
looks good!

Issue #100, opened 2022-11-29
Fix `prove_loop` in `dsProvePass`


d-xo:
This test will only pass with some `maxIter` set. We should hook up the `maxIter` setting and make sure it is respected when executing prove tests.

PR #108, opened 2022-12-08
Propagate `maxIter` parameter and enable `prove_loop`

zoep:
This is a fix for #100. Summary of changes:
- Propagate the `UnitTestOptions` to VeriOpts in SymExec.hs.
- Add debug option in `UnitTestOptions`
- Move `prove_loop` in a separate file
- Run `prove_loop` with two different `maxIter` params (passing and failing)


src/hevm/src/EVM/UnitTest.hs
```diff
@@ -74,6 +75,7 @@ data UnitTestOptions = UnitTestOptions
   , verbose     :: Maybe Int
   , maxIter     :: Maybe Integer
   , askSmtIters :: Maybe Integer
+  , debug       :: Bool
```
d-xo, line 78: could we call this smtdebug? just to make the distinction between dumping the smt queries and running the visual debugger clear?

d-xo (commented):
Looks great! just a little nit re: naming

src/hevm/src/EVM/UnitTest.hs
zoep: Sure! Do you think we should change the `bebug` option in `VeriOpts` as well or only this one?

zoep: Left `VeriOpts` as is since it prints nonSMT stuff as well

PR #116, opened 2022-12-09
Support post merge hardfork changes

d-xo:
- Renames DIFFICULTY -> PREVRANDAO
- Supports pasring prevRandao out of post merge blocks
- Integrates prevRandao into the rpc machinery

I need to test a bit more around this before we merge...


d-xo:
PR blew up a a bit again :see_no_evil: 

When testing around this change I noticed that we didn't pull the rpcInfo into `verify` when executing solidity unit tests. This required a small refactor in `UnitTest.hs` that changed how we constructed Fetchers. Previously we constructed a single fetcher and passed it in via `UnitTestOptions`, but `verify` expected to be able to construct it's `Fetcher` itself. So I changed `UnitTestOptions` to take a `SolverGroup` and `RpcInfo` and constructed the fetchers inline where they were needed in `UnitTest.hs`.

I'm also working on setting up a proper automated test harness for remote state tests here (since we currently do not test the rpc features at all :scream_cat:), and ended up writing a small assembler for EVM so we can have a relatively high level syntax for defining tests using bytecode directly.

PR #126, opened 2022-12-15
Group error cases in Expr

zoep:
Instead of having an `Expr` constructor for each EVM failure case, this groups failures under the same constructor with the help of an auxiliary data type. This helps avoiding repetition when pattern matching against `Expr` and makes it easier to add new error cases. 

I'm not very attached to it, just thought it would be cleaner and wanted to try it out.

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -346,14 +346,14 @@ runExpr = do
     Nothing -> error "Internal Error: vm in intermediate state after call to runFully"
     Just (VMSuccess buf) -> Return asserts buf (view (env . EVM.storage) vm)
     Just (VMFailure e) -> case e of
-      UnrecognizedOpcode _ -> Invalid asserts
-      SelfDestruction -> SelfDestruct asserts
-      EVM.StackLimitExceeded -> EVM.Types.StackLimitExceeded asserts
-      EVM.IllegalOverflow -> EVM.Types.IllegalOverflow asserts
+      UnrecognizedOpcode _ -> Failure asserts Invalid
+      SelfDestruction -> Failure asserts SelfDestruct
+      EVM.StackLimitExceeded -> Failure asserts EVM.Types.StackLimitExceeded
+      EVM.IllegalOverflow -> Failure asserts EVM.Types.IllegalOverflow
       EVM.Revert buf -> EVM.Types.Revert asserts buf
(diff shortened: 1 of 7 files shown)


src/hevm/src/EVM/SymExec.hs
```diff
@@ -583,14 +572,9 @@ equivalenceCheck solvers bytecodeA bytecodeB opts signature' = do
...
             (_, Revert _ _) -> PBool True
-            (Invalid _, Invalid _) -> PBool False
-            (Invalid _, _ ) -> PBool True
-            (_, Invalid _) -> PBool True
-            (EVM.Types.StackLimitExceeded _, EVM.Types.StackLimitExceeded _ ) -> PBool False
-            (EVM.Types.StackLimitExceeded _, _) -> PBool True
-            (_, EVM.Types.StackLimitExceeded _) -> PBool True
-            (a, b) -> if a == b then PBool False
-                                else error $ "Unimplemented, see TODO about EVM failures and TmpExpr. Left: " <> show a <> " Right: " <> show b
+            (Failure erra _, Failure errb _) -> if erra==errb then PBool False else PBool True
```
zoep, line 575: This has slightly different semantics than the previous equivalence check. Previously, if end results were different and both of them were one of the unhandled error cases, then it would raise the unimplemented exception. Now all error are handled in the same way. Note that `TmpError` cannot happen here since `flattenExpr` catches it and raises an exception.

d-xo (approved):
looks good

src/hevm/src/EVM/SymExec.hs
msooseth: This looks correct to me. However, for the sake of defensive coding, I'd try to check for `TmpError` here, and throw an `error`. Maybe in the future someone will use this function for something else, where `TmpError` hasn't been filtered yet.

msooseth (approved):
Looks good, but I'd change it a bit to do the checking for `TmpError`. It's up to you though :)

src/hevm/src/EVM/SymExec.hs
zoep: Do we want to return True if both programs return the same TmpError as before? Or raise an error if there is an unhandled error in either of the two programs, regardless of wether the error is the same?

msooseth: I'd just raise an `error.` I think `TmpError` is effectively an internal error, rather than a EVM error.

zoep: I agree! Fixed

PR #127, opened 2022-12-15
Fix verify rpc

d-xo:
## Description

This fixes symbolic execution against remote state (via rpc calls into an ethereum node) when executing via a `ds-test` harness. Prior to this change we didn't pass the required rpc info into `verify` in the `symRun` function in `UnitTest.hs`. To fix this I changed how we store the rpc info in `UnitTestOptions`. We used to store a fully constructed `Fetcher` in `UnitTestOptions`, but `verify` takes a raw `RpcInfo` and then constructs it's `Fetcher` internally. 

This meant we either had to change `verify` to take a fully constructed `Fetcher`, or change `UnitTestOptions` so it holds a raw `RpcInfo`. After playing around a bit the first seemed to produce a cleaner diff than the second so I went with that. 

I also added tests to `rpc.hs` that exercise this functionality, and modularised some of the test utils so that they can be shared between the `rpc-tests` binary, and the main `test` binary.

If you want to test around this change a bit, you can add tests to the `rpc.sol` file and run them from the rpc-tests repl (cabal  repl rpc-tests) with `:main -p "dapp-test"`.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -52,14 +52,24 @@ data VeriOpts = VeriOpts
   , debug :: Bool
   , maxIter :: Maybe Integer
   , askSmtIters :: Maybe Integer
+  , rpcInfo :: Fetch.RpcInfo
   }
   deriving (Eq, Show)
 
 defaultVeriOpts :: VeriOpts
-defaultVeriOpts = VeriOpts { simp = True, debug = False, maxIter = Nothing, askSmtIters = Nothing }
+defaultVeriOpts = VeriOpts
+  { simp = True
+  , debug = False
+  , maxIter = Nothing
+  , askSmtIters = Nothing
+  , rpcInfo = Nothing
+  }
+
+rpcVeriOpts :: (Fetch.BlockNumber, Text) -> VeriOpts
+rpcVeriOpts info = defaultVeriOpts { rpcInfo = Just info }
 
 debugVeriOpts :: VeriOpts
-debugVeriOpts = VeriOpts { simp = True, debug = True, maxIter = Nothing, askSmtIters = Nothing }
+debugVeriOpts = defaultVeriOpts { debug = True }
 
 extractCex :: VerifyResult -> Maybe (Expr End, SMTCex)
 extractCex (Cex c) = Just c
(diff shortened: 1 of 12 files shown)


src/hevm/test/rpc.hs
```diff
@@ -43,11 +54,94 @@ tests = testGroup "rpc"
...
         assertEqual "basefee" 22163046690 basefee
         assertEqual "prevRan" 0x2267531ab030ed32fd5f2ef51f81427332d0becbd74fe7f4cd5684ddf4b287e0 prevRan
     ]
+  , testGroup "execution with remote state"
+    -- execute against remote state from a ds-test harness
+    [ testCase "dapp-test" $ do
+        let testFile = "test/contracts/pass/rpc.sol"
+        runDappTestCustom testFile ".*" Nothing False testRpcInfo >>= assertEqual "test result" True
+
+    -- concretely exec "transfer" on WETH9 using remote rpc
```
msooseth, line 67: Perhaps we can reference `https://etherscan.io/token/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2#code` here somewhere?

src/hevm/src/EVM/SymExec.hs
```diff
@@ -453,11 +463,11 @@ extractProps = \case
 
 
 -- | Symbolically execute the VM and check all endstates against the postcondition, if available.
-verify :: SolverGroup -> VeriOpts -> VM -> Maybe (Fetch.BlockNumber, Text) -> Maybe Postcondition -> IO (Expr End, [VerifyResult])
-verify solvers opts preState rpcinfo maybepost = do
+verify :: SolverGroup -> VeriOpts -> VM -> Maybe Postcondition -> IO (Expr End, [VerifyResult])
```
msooseth, line 466: This is actually cleaner. I like passing it as part of `VeriOpts`. I was always confused what is what

msooseth (commented):
Looks good, but I think some files are referenced but missing somehow from the diff and the codebase? Maybe I'm doing something wrong.

src/hevm/test/rpc.hs
d-xo: done.

src/hevm/test/EVM/TestUtils.hs
```diff
@@ -0,0 +1,136 @@
...
+    dsTest <- readFile =<< Paths.getDataFileName "test/contracts/lib/test.sol"
+    erc20 <- readFile =<< Paths.getDataFileName "test/contracts/lib/erc20.sol"
+    testFilePath <- Paths.getDataFileName src
+    testFile <- readFile testFilePath
+    T.writeFile file
+      [i|
+      {
+        "language": "Solidity",
+        "sources": {
+          "ds-test/test.sol": {
```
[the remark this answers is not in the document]
msooseth: Sorry, that's a virtual filesystem. Closing.

msooseth (approved):
My last confusion has been resolved :)

PR #129, opened 2022-12-16
Warnings

zoep:
Remove warnings. Most of them where redundant imports, unused variables and code, nonexhaustive matches, and shadowing bindings.


msooseth (approved):
Looking good!

PR #143, opened 2022-12-19
Fixing QuickCheck generation so it doesn't generate SMT2 syntax errors

msooseth:
## Description
Fixes name generation for quickcheck fuzzer

Closes https://github.com/ethereum/hevm/issues/139

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/hevm/test/test.hs
+++ b/src/hevm/test/test.hs
@@ -2239,6 +2239,7 @@ genNat :: Gen Int
 genNat = fmap fromIntegral (arbitrary :: Gen Natural)
 
 genName :: Gen Text
+-- In order not to generate SMT reserved words, we prepend with "esc_"
 genName = fmap (T.pack . ("esc_" <> )) $ listOf1 (oneof . (fmap pure) $ ['a'..'z'] <> ['A'..'Z'])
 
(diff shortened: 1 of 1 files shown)


src/hevm/test/test.hs
```diff
@@ -2239,7 +2239,7 @@ genNat :: Gen Int
 genNat = fmap fromIntegral (arbitrary :: Gen Natural)
 
 genName :: Gen Text
-genName = fmap T.pack $ listOf1 (oneof . (fmap pure) $ ['a'..'z'] <> ['A'..'Z'])
+genName = fmap (T.pack . ("esc_" <> )) $ listOf1 (oneof . (fmap pure) $ ['a'..'z'] <> ['A'..'Z'])
```
d-xo, line 2242: maybe worth a comment here explaining what the `esc_` is for?

d-xo (approved):
Looks good!

src/hevm/test/test.hs
msooseth: Good point! I add an explanation :)

zoep (approved):

PR #148, opened 2022-12-20
Less SMT queries for equivalence checking through query caching

msooseth:
## Description
This is a rough PR, and needs some help with how I can make this more understandable. So advice welcome how to make this cleaner.

The idea is the following. If we have:

```
if a then if x then y else y
     else z
```

Then if we proved that `a` is UNSAT, then there is no point in checking `a & x`, because `a&x` is UNSAT for sure. Hence, we should cache that `a` was UNSAT. 

We want to have `a` checked before `a&x`, so there should be some order in how we check. I order them according to size, which is not optimal but it's not terrible.

Next we want to have this parallel, because we have K processors. So we need some synchronization for the cache, which is achieved via `STM`. Finally, we want the threads to go through the job in an orderly fashion, not just randomly -- which is what `mapConcurrently` used to do.

Thanks to d-xo  for all the fixes to this, it's WAY cleaner AND better now!

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/hevm/hevm.cabal
```diff
@@ -176,7 +176,9 @@ library
     smt2-parser                       >= 0.1.0.1,
     word-wrap                         >= 0.5 && < 0.6,
     spool                             >= 0.1 && < 0.2,
-    parsec
+    parsec,
+    stm,
+    spawn
```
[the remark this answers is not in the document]
msooseth: Oh, I see! I didn't understand. OK, it's been fixed, along with `parsec` now!

d-xo:
yes I think this is good to go once merge conflicts are resolved

d-xo:
Oh, also just a little note in the changelog would be great here

msooseth:
@d-xo I had to do a few changes to make sure it works with the latest changes. Can you please do a quick review? I'm sorry, but there were really a lot of changes since and so I had to adjust things. I tried my very best to change the least possible.

hevm.cabal
```diff
@@ -263,5 +263,55 @@ common test-base
...
     vector,
     smt2-parser >= 0.1.0.1,
-    stm,
-    spawn
+    stm >= 2.5.0,
+    spawn >= 0.3,
+    witherable,
+    smt2-parser >= 0.1.0.1
+
+library test-utils
```
d-xo, line 271: I don't understand why this is showing up in the diff?

test/test.hs
```diff
@@ -1795,14 +1796,13 @@ tests = testGroup "hevm"
...
       , testCase "eq-all-yul-optimization-tests" $ do
-        let myVeriOpts = VeriOpts{ simp = True, debug = False, maxIter = Just 5, askSmtIters = Just 20 }
-            ignoredTests =
-                    --[
-                      --"controlFlowSimplifier/terminating_for_nested.yul"
-                    --, "controlFlowSimplifier/terminating_for_nested_reversed.yul"
+        let myVeriOpts = VeriOpts{ simp = True, debug = False, maxIter = Just 5, askSmtIters = Just 20, rpcInfo = Nothing }
+            ignoredTests = [
+                    -- "controlFlowSimplifier/terminating_for_nested.yul"
+                    -- , "controlFlowSimplifier/terminating_for_nested_reversed.yul"
```
d-xo, line 1802: why is this commented out code here?

test/test.hs
```diff
@@ -2001,7 +2001,7 @@ tests = testGroup "hevm"
                     , "fullSuite/extcodelength.yul" -- extcodecopy bug?
                     , "loadResolver/keccak_short.yul" -- keccak bug
                     , "reasoningBasedSimplifier/signed_division.yul" -- ACTUAL bug, SDIV I think?
-                    ]
+                           ]
```
d-xo, line 2004: indentation

hevm.cabal
msooseth: Sorry, was a merge mistake :( Thanks for catching it!

msooseth: Fixed :)

test/test.hs
msooseth: Sorry for this coment ending up in there. With the improvements we made, these actually can be checked. So some got commented out from the list of things that were ignored, but I forgot to actually delete them. Deleted now!

test/test.hs
msooseth: Hm, sorry, this was weird -- GHC complained about the indentation not matching the starting (`[`) bracket. So I moved it to the same column. But now it's happy to move it back. Weird, I'm still confused sometimes about indentation.

msooseth: Fixed!

PR #149, opened 2022-12-21
Tidy less queries

d-xo:
## Description

- Tidies & refactors #148 a bit
- Enables a few more yul equivalence tests

### DIFF
--- a/src/hevm/src/EVM/SymExec.hs
+++ b/src/hevm/src/EVM/SymExec.hs
@@ -6,7 +6,7 @@ module EVM.SymExec where
 import Prelude hiding (Word)
 
 import Control.Lens hiding (pre)
-import EVM hiding (Query, Revert, push)
+import EVM hiding (Query, Revert, push, bytecode, cache)
 import qualified EVM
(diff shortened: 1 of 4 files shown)


src/hevm/src/EVM/SymExec.hs
```diff
@@ -558,34 +563,112 @@ verify solvers opts preState rpcinfo maybepost = do
...
+      res <- if subsetAny props ku
+             then pure (True, Unsat)
+             else (fmap ((False),) (checkSat solvers smt))
+      case res of
+        (_, Sat x) -> pure (Cex x, False)
+        (quick, Unsat) -> case quick of
+                            True  -> pure (Qed (), quick)
+                            False -> do
+                              -- nb: we might end up with duplicates here due to a
+                              -- potential race, but it doesn't matter for correctness
```
msooseth, line 641: That's a good idea to explain, thanks, should  have done that myself :S

msooseth (approved):
Thanks for the fix! Much better & faster!

PR #152, opened 2022-12-27
Optimize concrete runtime contract code

arcz:
## Description
This PR optimizes runtime contract code that is known to be concrete. It resulted in a significant performance bump when executing ethereum-tests suite on my machine down to 76s from 86s (~13%).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -15,7 +15,7 @@ import EVM.Types hiding (IllegalOverflow, Error)
 import EVM.Solidity
 import EVM.Concrete (createAddress, create2Address)
 import EVM.Op
-import EVM.Expr (readStorage, writeStorage, readByte, readWord, writeWord, writeByte, bufLength, indexWord, litAddr, readBytes, word256At, copySlice, isLitByte)
+import EVM.Expr (readStorage, writeStorage, readByte, readWord, writeWord, writeByte, bufLength, indexWord, litAddr, readBytes, word256At, copySlice)
 import EVM.FeeSchedule (FeeSchedule (..))
 import Options.Generic as Options
 import qualified EVM.Precompiled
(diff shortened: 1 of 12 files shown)


src/EVM.hs
```diff
@@ -305,10 +305,15 @@ data SubState = SubState
   hopefully we do not have to deal with dynamic immutable before we get a real data section...
 -}
 data ContractCode
-  = InitCode ByteString (Expr Buf)     -- ^ "Constructor" code, during contract creation
-  | RuntimeCode (V.Vector (Expr Byte)) -- ^ "Instance" code, after contract creation
+  = InitCode ByteString (Expr Buf) -- ^ "Constructor" code, during contract creation
+  | RuntimeCode RuntimeCode -- ^ "Instance" code, after contract creation
   deriving (Show)
 
+data RuntimeCode
```
d-xo, line 312:
```suggestion
-- | We have two variants here to optimize the fully concrete case.
-- ConcreteRuntimeCode just wraps a ByteString
-- SymbolicRuntimeCode is a fixed length vector of potentially symbolic bytes, which lets us handle symbolic pushdata (e.g. from immutable variables in solidity).
data RuntimeCode
```

src/EVM.hs
```diff
@@ -1075,7 +1079,9 @@ exec1 = do
           case stk of
             (x:xs) ->
               burn g_mid $ forceConcrete x "JUMP: symbolic jumpdest" $ \x' ->
-                checkJump x' xs
```
d-xo, line 1078: why remove the `checkJump` here?

d-xo (commented):
Looks great generally, just a couple of nits & questions.

src/EVM.hs
arcz: it's still there but guarded under the case

d-xo: oh of course... :woman_facepalming:

d-xo (approved):

PR #163, opened 2023-01-04
Fix SAR arithmetic overflow and copySlice regressions

arcz:
## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -330,7 +330,6 @@ copySlice a@(Lit srcOffset) b@(Lit dstOffset) c@(Lit size) d@(ConcreteBuf src) e
             else padRight (num size) $ BS.take (num size) (BS.drop (num srcOffset) src)
           tl = BS.drop (num dstOffset + num size) dst
       in ConcreteBuf $ hd <> sl <> tl
-  | size < maxBytes = ConcreteBuf $ BS.replicate (num size) 0
   | otherwise = CopySlice a b c d e
 
 -- copying 32 bytes can be rewritten to a WriteWord on dst (e.g. CODECOPY of args during constructors)
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -181,8 +181,12 @@ shr = op2
 
 sar :: Expr EWord -> Expr EWord -> Expr EWord
 sar = op2 SAR (\x y ->
-  let asSigned = (fromIntegral y) :: Int256
-  in fromIntegral $ shiftR asSigned (fromIntegral x))
+  let msb = testBit y 255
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -319,14 +323,14 @@ copySlice a@(Lit srcOffset) b@(Lit dstOffset) c@(Lit size) d@(ConcreteBuf src) e
...
   | dstOffset < maxBytes
-  , srcOffset < maxBytes
   , size < maxBytes =
       let hd = padRight (num dstOffset) $ BS.take (num dstOffset) dst
           sl = if srcOffset > num (BS.length src)
             then BS.replicate (num size) 0
             else padRight (num size) $ BS.take (num size) (BS.drop (num srcOffset) src)
           tl = BS.drop (num dstOffset + num size) dst
       in ConcreteBuf $ hd <> sl <> tl
+  | size < maxBytes = ConcreteBuf $ BS.replicate (num size) 0
```
msooseth, line 333: Sorry, I don't understand :( So if `dstOffset >= maxBytes` and `size < maxBytes` then we return all-0? I am not familiar with the EVM, is this somewhere described, or am I misunderstanding this? Just wanna make sure I understand this correctly, sorry :(

msooseth: But we are reading from a sane location, right? We are writing to an insane location, that's true. If we were _reading_ from an insane location (if `srcoffset` was insane), then I understand. But it's the `dstoffset` that's insane here, `srcoffset` may well be correct here.

So what you mean is that since we write to an insane location, it will never be read, hence it'll all be 0? Actually, does it make sense to write to an insane location? Maybe such a write should be skipped? (though then I don't know how we'd keep track that it would have ran out of gas).

msooseth: All good! It's actually a highly non-trivial thing! Thanks so much for walking me through it! Good stuff!

msooseth (approved):

src/EVM/Expr.hs
```diff
@@ -319,7 +323,6 @@ copySlice a@(Lit srcOffset) b@(Lit dstOffset) c@(Lit size) d@(ConcreteBuf src) e
 
 copySlice a@(Lit srcOffset) b@(Lit dstOffset) c@(Lit size) d@(ConcreteBuf src) e@(ConcreteBuf dst)
   | dstOffset < maxBytes
-  , srcOffset < maxBytes
```
d-xo, line 322: why?

d-xo:
A little note in the changelog on what was fixed here would be great too

d-xo (approved):
Thanks!

PR #164, opened 2023-01-05
Minor fixes: missing `next`, `mulmod`, `BaseFee`, etc.

msooseth:
## Description
This is a set of things I have found to be missing/incorrect with my fuzzing adventures. In particular, it:
* Adds `-j -O0` option in case you pass ` -f devel` to your `cabal repl test`. This will _significantly_ improve your compile times. Note that you can't do performance debugging with that. I personally plan on using `-f devel` but you can skip this change by not passing `-f devel` and never notice the difference.
* Adds missing `OpBaseFee`
* Adds `StackUnderrun`
* Test for `addmod`
* Simplify a test `return`
* Missing `next` at the end of `EXTCODESIZE`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -58,7 +58,7 @@ common shared
   if flag(ci)
     ghc-options: -Werror
   if flag(devel)
-    ghc-options: -j --disable-optimization
+    ghc-options: -j -O0
   default-language: GHC2021
   default-extensions:
     LambdaCase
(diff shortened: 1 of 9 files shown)


d-xo (approved):
looks good!

PR #166, opened 2023-01-06
[WIP] Better error handling

msooseth:
## Description
* Rename Error in `EVM` to `EVMError`.
* Rename Error in `EVM.Types` to `ExprError`
* Rename `UnrecognizedOpcode` to `InvalidOpcode` (this make this the same name as in `EVM.Types`)
* Rename `SelfDestruction` to `SelfDestruct` (this make this the same name as in `EVM.Types`)
* Reorder case checks in `runExpr` to match that in `EVM.Types` so it's clear we cover all
* Rename `TmpError` to `WrappedEVMError`, since it's ONLY generated as part of `runExpr`, and only when it cannot be converted into an `EVM.Types.ExprError` or a `Revert`.
* Remove `SMTTimeout` from `EVMError` since it's never used and does not belong there
* Add `SMTError` to `VerifyResult` to account for SMT solver errors
* Return `Either ExprError` from `verify`, `equivalenceCheck` and associated functions, when `Expr` generation fails
* Adding `Assembler.hs` by dxo
* Adding random bytecode generator to do some fuzzing using QuickCheck. This fixes up `JumpDestination`-s after generation and ensures that jumps only happen _forward_ so we never encode any loops.
* When copying a large `CopySlice`, return a `Left` instead of a very-very large SMT file
* Minor English typo corrections

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -64,8 +64,8 @@ import Crypto.PubKey.ECC.Generate (generateQ)
 -- | EVM failure modes
 data EVMError
   = BalanceTooLow W256 W256
-  | UnrecognizedOpcode Word8
-  | SelfDestruction
+  | InvalidOpcode Word8
+  | SelfDestruct
   | StackUnderrun
   | BadJumpDestination
   | Revert (Expr Buf)
@@ -1334,7 +1334,7 @@ exec1 = do
             _ -> underrun
 
         xxx ->
-          vmError (UnrecognizedOpcode xxx)
+          vmError (EVM.InvalidOpcode xxx)
 
 transfer :: Addr -> Addr -> W256 -> EVM ()
 transfer xFrom xTo xValue =
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -62,7 +62,7 @@ import Crypto.PubKey.ECC.Generate (generateQ)
 -- * Data types
 
 -- | EVM failure modes
-data Error
+data EVMError
   = BalanceTooLow W256 W256
   | UnrecognizedOpcode Word8
   | SelfDestruction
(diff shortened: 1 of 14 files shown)


src/EVM.hs
```diff
@@ -62,10 +62,10 @@ import Crypto.PubKey.ECC.Generate (generateQ)
 -- * Data types
 
 -- | EVM failure modes
-data Error
+data EVMError
   = BalanceTooLow W256 W256
-  | UnrecognizedOpcode Word8
-  | SelfDestruction
+  | InvalidOpcode Word8
```
d-xo, line 67: The `Invalid` opcode is (just by convention) a special opcode (`0xfe`), wheras there are many unrecognised opcodes. Prior to solc 0.8, failed `assert` statements would result in the execution of `0xfe` (which would also consume all remaining gas). These days they revert with a specific error code. 

It's not a huge deal, but I somewhwat prefer the older name, since it is more in line with this conventional usage of `Invalid` in the context of the EVM.

src/EVM/SMT.hs
```diff
@@ -1049,7 +1049,8 @@ readSExpr h = go 0 0 []
 
 -- | Stores a region of src into dst
 copySlice :: Expr EWord -> Expr EWord -> Expr EWord -> Builder -> Builder -> Builder
-copySlice srcOffset dstOffset size@(Lit _) src dst
+copySlice srcOffset dstOffset size@(Lit sz) src dst
+  | sz > (1024*16) = error "Too large copySlice size"
```
d-xo, line 1053: why?

src/EVM/SymExec.hs
```diff
@@ -62,6 +60,16 @@ isQed :: ProofResult a b c -> Bool
 isQed (Qed _) = True
 isQed _ = False
 
+sameCnstr :: ProofResult a b c -> ProofResult a b c -> Bool
```
d-xo, line 63: I think we can clean this up like this? 

https://stackoverflow.com/a/10112843

PR #168, opened 2023-01-10
Storage counterexamples

zoep:
## Description

Construction of storage counterexamples based on the model returned by the SMT. The approach is: 

- Keep track of the storage slots that are read by a particular query.
- Parse the storage model returned by the SMT as a Haskell function.
- Query the SMT solver for the concrete value of the slots that we are interested in, and add them to a storage map. The value of the map is the value of the function for the given slot.

## Checklist

- [X] tested locally
- [X] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -803,11 +803,16 @@ getStore inst = do
 
 parse1DArray :: (Map Symbol Term) -> (W256 -> Maybe W256) -> Term -> (W256 -> Maybe W256)
 parse1DArray env f = \case
+  -- variable reference
+  TermQualIdentifier (Unqualified (IdSymbol s)) ->
+    case Map.lookup s env of
+      Just t -> parse1DArray env f t
+      Nothing -> error "Internal error: unknown identifier, cannot parse array"
   -- let (x t') t
   TermLet (VarBinding x t' :| []) t -> parse1DArray (Map.insert x t' env) f t
   TermLet (VarBinding x t' :| lets) t -> parse1DArray (Map.insert x t' env) f (TermLet (NonEmpty.fromList lets) t)
-  -- (as const (Array (_ BitVec 256) (_ BitVec 256))) <constant>
-  TermApplication asconst (TermSpecConstant val :| []) | isAsConst asconst ->
+  -- (as const (Array (_ BitVec 256) (_ BitVec 256))) SpecConstant
+  TermApplication asconst (TermSpecConstant val :| []) | is1DArrConst asconst ->
     \_ -> Just $ parseW256 val
   -- store arr ind val
   TermApplication store (TermSpecConstant ind :| [TermSpecConstant val]) | isStore store ->
@@ -815,7 +820,7 @@ parse1DArray env f = \case
   _ -> error "Internal error: cannot parse array value"
   where
 
-    isAsConst = \case
+    is1DArrConst = \case
       Qualified (IdSymbol "const") sort -> is1DArray sort
       _ -> False
     is1DArray = \case
@@ -829,6 +834,43 @@ parse1DArray env f = \case
       Unqualified (IdSymbol "store") -> True
       _ -> False
 
+
+parse2DArray :: (Map Symbol Term) -> (W256 -> W256 -> Maybe W256) -> Term -> (W256 -> W256 -> Maybe W256)
+parse2DArray env f = \case
+  -- variable reference
+  TermQualIdentifier (Unqualified (IdSymbol s)) ->
+    case Map.lookup s env of
+      Just t -> parse2DArray env f t
+      Nothing -> error "Internal error: unknown identifier, cannot parse array"
+  -- let (x t') t
+  TermLet (VarBinding x t' :| []) t -> parse2DArray (Map.insert x t' env) f t
+  TermLet (VarBinding x t' :| lets) t -> parse2DArray (Map.insert x t' env) f (TermLet (NonEmpty.fromList lets) t)
+  -- (as const (Array (_ BitVec 256) (Array (_ BitVec 256) (_ BitVec 256)))) 1DarrayConstant
+  TermApplication asconst (val :| []) | is2DArrConst asconst ->
+    \x -> parse1DArray env (f x) val
+  -- store arr2D ind arr1D
+  TermApplication store (TermSpecConstant ind :| [val]) | isStore store ->
+    \x -> if x == parseW256 ind then parse1DArray env (f x) val else f x
+  _ -> error "Internal error: cannot parse array value"
+  where
+
+    is2DArrConst = \case
+      Qualified (IdSymbol "const") sort -> is2DArray sort
+      _ -> False
+    is2DArray = \case
+      SortParameter (IdSymbol "Array") (sort1 :| [sort2]) ->
+         isBitVec256 sort1 && is1DArray sort2
+      _ -> False
+    is1DArray = \case
+      SortParameter (IdSymbol "Array") (sort1 :| [sort2]) ->
+         isBitVec256 sort1 && isBitVec256 sort2
+      _ -> False
+    isBitVec256 = \case
+      SortSymbol (IdIndexed "BitVec" (IxNumeral "256" :| [])) -> True
+      _ -> False
+    isStore = \case
+      Unqualified (IdSymbol "store") -> True
+      _ -> False
 
 
 getBufs :: SolverInstance -> [TS.Text] -> IO (Map (Expr Buf) ByteString)
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -829,73 +829,52 @@ getStore inst sreads = do
         r -> parseErr r
 
 
-is1DArrConst :: QualIdentifier -> Bool
-is1DArrConst = \case
-  Qualified (IdSymbol "const") sort -> is1DArray sort
-  _ -> False
 
-is1DArray :: Sort -> Bool
-is1DArray = \case
-  SortParameter (IdSymbol "Array") (sort1 :| [sort2]) ->
-    isBitVec256 sort1 && isBitVec256 sort2
-  _ -> False
+-- | Interpret an N-dimentional array as a value of type a.
+-- Parameterized by an interpretation function for array values.
+interpretNDArray :: (Map Symbol Term -> Term -> a) -> (Map Symbol Term) -> Term -> (W256 -> a)
+interpretNDArray interp env = \case
+  -- variable reference
+  TermQualIdentifier (Unqualified (IdSymbol s)) ->
+    case Map.lookup s env of
+      Just t -> interpretNDArray interp env t
+      Nothing -> error "Internal error: unknown identifier, cannot parse array"
+  -- (let (x t') t)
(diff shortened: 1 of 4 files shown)


src/EVM/SMT.hs
```diff
@@ -362,138 +391,85 @@ declareBufs names = SMT2 ("; buffers" : fmap declareBuf names <> ("; buffer leng
...
-referencedBufs expr = nubOrd (foldExpr go [] expr)
-  where
-    go :: Expr a -> [Builder]
-    go = \case
-      AbstractBuf s -> [fromText s]
-      _ -> []
+referencedBufs expr = nubOrd $ foldExpr referencedBufsGo [] expr
 
 referencedBufs' :: Prop -> [Builder]
-referencedBufs' = \case
```
zoep, line 381: Rewrote a bunch of such functions using `foldProp` and hoisted the `go` helper outside of their `Expr` counterpart to use in both of them.

src/EVM/SMT.hs
```diff
@@ -811,6 +787,117 @@ getVars parseFn inst names = Map.mapKeys parseFn <$> foldM getOne mempty names
...
+  TermApplication asconst (TermSpecConstant val :| []) | is1DArrConst asconst ->
+    \_ -> parseW256 val
+  -- (store arr ind val)
+  TermApplication store (arr :| [TermSpecConstant ind, TermSpecConstant val]) | isStore store ->
+    \x -> if x == parseW256 ind then parseW256 val else interpret1DArray env arr x
+  t -> error $ "Internal error: cannot parse array value. Unexpected term: " <> (show t)
+
+-- | Interpret a 2-dimentional array as a function
+interpret2DArray :: (Map Symbol Term) -> Term -> (W256 -> W256 -> W256)
+interpret2DArray env = \case
```
zoep, line 883: There is some code duplication between `interpret2DArray` and `interpret1DArray`, but in order to generalise to `interpretNDArray` we would need a dependent types for the result, and I'm not sure if it's worth the hustle (though I'd be somewhat curious to try it out).

src/EVM/SMT.hs
```diff
@@ -811,6 +787,96 @@ getVars parseFn inst names = Map.mapKeys parseFn <$> foldM getOne mempty names
...
+interpretNDArray interp env = \case
+  -- variable reference
+  TermQualIdentifier (Unqualified (IdSymbol s)) ->
+    case Map.lookup s env of
+      Just t -> interpretNDArray interp env t
+      Nothing -> error "Internal error: unknown identifier, cannot parse array"
+  -- (let (x t') t)
+  TermLet (VarBinding x t' :| []) t -> interpretNDArray interp (Map.insert x t' env) t
+  TermLet (VarBinding x t' :| lets) t -> interpretNDArray interp (Map.insert x t' env) (TermLet (NonEmpty.fromList lets) t)
+  -- (as const (Array (_ BitVec 256) (_ BitVec 256))) SpecConstant
```
msooseth, line 845: Why is this commented out?

src/EVM/SMT.hs
```diff
@@ -811,6 +787,96 @@ getVars parseFn inst names = Map.mapKeys parseFn <$> foldM getOne mempty names
...
+  TermApplication asconst (val :| []) | isArrConst asconst ->
+    \_ -> interp env val
+  -- (store arr ind val)
```
msooseth, line 848: I don't understand this comment :(

test/test.hs
```diff
@@ -1736,6 +1746,55 @@ tests = testGroup "hevm"
...
+          let a = getVar cex "arg1"
+          let testCex = Map.size storeCex == 1 &&
+                        case Map.lookup addr storeCex of
+                          Just s -> Map.size s == 2 &&
+                                    case (Map.lookup 0 s, Map.lookup (10 + a) s) of
+                                      (Just x, Just y) -> x >= y
+                                      _ -> False
+                          _ -> False
+          assertBool "Did not find expected storage cex" testCex
+          putStrLn "Expected counterexample found"
```
msooseth, line 1797: That's COOL that even such a complicated one can be generated!

msooseth (commented):
Sorry, I only found minor issues :( I'll read through again tomorrow!

src/EVM/SMT.hs
```diff
@@ -107,6 +112,26 @@ declareIntermediates bufs stores =
...
+isAbstractStorage = getAny . foldExpr go (Any False)
+  where
+    go :: Expr a -> Any
+    go AbstractStore = Any True
+    go _ = Any False
+
+-- | This function overapproximates the reads from the abstract
+-- storage. Potentially, it can return a location that has been
+-- overwritten in the abstract store. However, such reads will have
+-- been simplified away, is the source expresion is simplified.
```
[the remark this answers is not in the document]
zoep: Yes, thanks for spotting!

src/EVM/SMT.hs
```diff
@@ -811,6 +787,96 @@ getVars parseFn inst names = Map.mapKeys parseFn <$> foldM getOne mempty names
...
+      case parseCommentFreeFileMsg getValueRes (T.toStrict raw) of
+        Right (ResSpecific (valParsed :| [])) ->
+          case valParsed of
+            (_, TermSpecConstant sc) -> pure $ parseW256 sc
+            _ -> error "Internal Error: cannot parse model for storage index"
+        r -> parseErr r
+
+
+
+-- | Interpret an N-dimentional array as a value of type a.
```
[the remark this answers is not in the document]
zoep: Yes, thanks for spotting this!

src/EVM/SMT.hs
zoep: these are comments denoting what is the SMT expression that this term corresponds too. Any idea how to make it clearer? For instance this node could correspond to a `((as const (Array (_ BitVec 256) (_ BitVec 256)))
  #x0000000000000000000000000000000000000000000000000000000000000001)`

src/EVM/SMT.hs
zoep: Same as above. This is a comment denoting this corresponds to a store operation. Happy to make it clearer, not sure how though

src/EVM/SMT.hs
```diff
@@ -362,138 +391,85 @@ declareBufs names = SMT2 ("; buffers" : fmap declareBuf names <> ("; buffer leng
...
+  _ -> []
 
 referencedBufs :: Expr a -> [Builder]
-referencedBufs expr = nubOrd (foldExpr go [] expr)
-  where
-    go :: Expr a -> [Builder]
-    go = \case
-      AbstractBuf s -> [fromText s]
-      _ -> []
+referencedBufs expr = nubOrd $ foldExpr referencedBufsGo [] expr
```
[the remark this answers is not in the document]
zoep: We can but in a rather ad hoc way: we could declare a typeclass of types that can by folded over a function of type `(forall a . Expr a -> b)` (Note that already existing [foldable class](https://en.wikibooks.org/wiki/Haskell/Foldable) does not quite work, since it's meant for container types like list)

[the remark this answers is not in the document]
msooseth: Oh, okay, then it's more trouble then it's worth. Thanks for the explanation, I learned something!

src/EVM/SMT.hs
msooseth: Ah, I see!!! No, that make sense, I didn't understand, I thought that was Haskell :D Sorry!

src/EVM/SMT.hs
msooseth: All good, sorry, didn't realize it was SMT2

msooseth (approved):
I had more time + slept on it, to check, seems OK to me.

src/EVM/SMT.hs
```diff
@@ -107,6 +112,26 @@ declareIntermediates bufs stores =
...
+isAbstractStorage = getAny . foldExpr go (Any False)
+  where
+    go :: Expr a -> Any
+    go AbstractStore = Any True
+    go _ = Any False
+
+-- | This function overapproximates the reads from the abstract
+-- storage. Potentially, it can return a location that has been
+-- overwritten in the abstract store. However, such reads will have
+-- been simplified away, if the source expression is simplified.
```
d-xo, line 125: I don't quite understand the comment about simplification, can you expand a bit?

src/EVM/SMT.hs
```diff
@@ -107,6 +112,26 @@ declareIntermediates bufs stores =
     encodeStore n expr =
        fromLazyText ("(define-const store" <> (T.pack . show $ n) <> " Storage ") <> exprToSMT expr <> ")"
 
+isAbstractStorage :: Expr Storage -> Bool
+isAbstractStorage = getAny . foldExpr go (Any False)
+  where
+    go :: Expr a -> Any
```
d-xo, line 118: This `Any` is pretty cute, didn't know about it before :)

src/EVM/SMT.hs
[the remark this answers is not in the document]
d-xo: afaict `referencedBufs` is only used inside of `referencedBufs'`. I think we could just inline it as a helper. The fact that we have both is probably just a bit of leftover cruft.

src/EVM/SMT.hs
```diff
@@ -107,6 +112,26 @@ declareIntermediates bufs stores =
...
+-- storage. Potentially, it can return a location that has been
+-- overwritten in the abstract store. However, such reads will have
+-- been simplified away, if the source expression is simplified.
+findStorageReads :: Prop -> [(Expr EWord, Expr EWord)]
+findStorageReads = foldProp go []
+  where
+    go :: Expr a -> [(Expr EWord, Expr EWord)]
+    go = \case
+      SLoad addr slot storage ->
+        if isAbstractStorage storage then [(addr, slot)] else []
```
d-xo, line 132:
```suggestion
      SLoad addr slot storage -> [(addr, slot) | isAbstractStorage storage]
```

src/EVM/SymExec.hs
```diff
@@ -602,7 +602,7 @@ showModel cd (expr, res) = do
 
 
 formatCex :: Expr Buf -> SMTCex -> Text
-formatCex cd m@(SMTCex _ _ blockContext txContext) = T.unlines $
+formatCex cd m@(SMTCex _ _ _ blockContext txContext) = T.unlines $
```
d-xo, line 605: Can we add some pretty printing for storage counterexamples here?

src/EVM/SMT.hs
zoep: Yes! I found out while I was trying to write this function, and was looking for a monoid for Bool. There's an `All` too :)

d-xo (commented):
Looks great generally. 

Would be nice to also test the codepaths where we are executing on top of a concrete store instead of an abatract one (check the `StorageModel` param in `verifyContract`).

src/EVM/SMT.hs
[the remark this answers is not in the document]
zoep: Both are used in `assertProps`, maybe a typeclass would help cleaning up this code a bit. I'm trying this out in my current branch.

d-xo (approved):
looks great!

PR #172, opened 2023-01-15
Enable OverloadedRecordDot, NoFieldSelectors and DuplicateRecordFields

arcz:
## Description
`OverloadedRecordDot` is available from 9.2 and allows for nice records access without lens. I think the next step would be to move to something more modern for lens such as `generic-lens` or `optics-core`. We can then drop the underscores from records and access them through the dot syntax and reach for lens for more complex cases.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -471,72 +471,72 @@ currentContract vm =
 
 makeVm :: VMOpts -> VM
 makeVm o =
-  let txaccessList = vmoptTxAccessList o
-      txorigin = vmoptOrigin o
-      txtoAddr = vmoptAddress o
+  let txaccessList = o.vmoptTxAccessList
+      txorigin = o.vmoptOrigin
+      txtoAddr = o.vmoptAddress
       initialAccessedAddrs = fromList $ [txorigin, txtoAddr] ++ [1..9] ++ (Map.keys txaccessList)
       initialAccessedStorageKeys = fromList $ foldMap (uncurry (map . (,))) (Map.toList txaccessList)
-      touched = if vmoptCreate o then [txorigin] else [txorigin, txtoAddr]
+      touched = if o.vmoptCreate then [txorigin] else [txorigin, txtoAddr]
   in
   VM
   { _result = Nothing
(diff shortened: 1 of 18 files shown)


src/EVM.hs
```diff
@@ -590,7 +590,7 @@ exec1 = do
           copyBytesToMemory (the state calldata) (Lit calldatasize) (Lit 0) (Lit 0)
           executePrecompile self (the state gas) 0 calldatasize 0 0 []
           vmx <- get
-          case view (state.stack) vmx of
```
d-xo, line 593: Can this be `vm.state.stack`?

d-xo:
This is great, awesome to see such advanced technology arrive in haskell after 30 years :sweat_smile: 

Supportive of the longer term refactoring plans to move away from lens. I haven't personally used optics, but I would definitely be interested in trying it out as I have always found the type signatures and documentation in the og lens library extremely hard to parse. We can also probably tidy the field names in quite a few records since duplicates are allowed now. 

One concern I do have is how this will impact any downstream consumers who do not have `OverloadedRecordDot` and friends enabled? I guess at the very least this means we can't compile with ghc < 9.2 (not a blocker, just curious).

src/EVM.hs
arcz: It can be `vmx._state._stack` at the moment. I changed this line because with `OverloadedRecordDot` the function composition dot has to have spaces around it. I made as few changes as possible to enable those extensions in this PR, but I can change the rest of the trivial `view`s in the next PR.

arcz:
If needed, for GHC < 9.2, we could add this for `OverloadedRecordDot` https://hackage.haskell.org/package/record-dot-preprocessor, the `NoFieldSelectors` extension can be simply dropped and `DuplicateRecordFields` is available in the older compilers.

d-xo:
I think having a min dep on 9.2 is fine for now, but good to know that we have options if it becomes problematic. Thanks for the pr!

PR #174, opened 2023-01-18
test: small nits in tests

zoep:
Removes a few redundant comments (some of them possibly propagated because of copy paste) and enables test


d-xo (approved):
lgtm

PR #176, opened 2023-01-23
SMT: remove dependency to SolverInstance from cex parsing functions

zoep:
## Description

Removes the dependency to `SolverInstance` from the counterexample parsing functions, so that the counter example parsing logic can be moved into its own file without any mutual dependecy.

Instead of passing the `SolverInstance` parameter, we can pass a function that returns the raw value of a variable from the SMT model. This function will be the closure `getValue inst`.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
looks good!

PR #178, opened 2023-01-26
Move solver orchestration to separate file

zoep:
## Description

Moves solver execution utilities from SMT.hs to new Solvers.hs file. Line history is preserved in both files.

My initial intention was to also move the Cex parsing logic into its own file, but I didn't because of a mutual dependency to SMT.hs (see comment for possible solutions).

There are no changes to existing code, just some rearrangements in the order of functions within the SMT.hs file.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM/SMT.hs
```diff
@@ -961,216 +874,91 @@ getBufs getVal names = foldM getBuf mempty names
...
+      case Map.lookup addr store of
+        Just m -> Map.insert addr (Map.insert slot (fun addr slot) m) store
+        Nothing -> Map.insert addr (Map.singleton slot (fun addr slot)) store
 
 
--- ** Helpers ** ---------------------------------------------------------------------------------
+    queryValue :: Expr EWord -> IO W256
+    queryValue (Lit w) = pure w
+    queryValue w = do
+      let expr = toLazyText $ exprToSMT w
```
zoep, line 909: I would like to move the Cex datatypes and functions to their own file `Cex.hs`  but because of this call to `exprToSMT`, the Cex parsing depends on SMT encoding and the SMT encoding depends on Cex because the`SMT2` datatype uses the `CexVars` datatype. There are possible workarounds:

1) Keep the `CexVars` datatype in the SMT.hs file. Import SMT from Cex. 
2) Leave the file as is 
3) Encode the value of storage reads when constructing the CexVars records and pass it along with their Expr representation.
4) Pass a callback to `exprToSMT` to the `getStore` function

Thoughts?

d-xo: 1 seems fine to me

msooseth (approved):

PR #179, opened 2023-01-27
cli: symbolic: correct status code when cex are found

d-xo:
## Description

v small bug where we weren't exiting with the correct status code if cexs were found during symbolic execution

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


zoep (approved):
Looks good!

PR #180, opened 2023-01-27
Fix slot fetch cache lookup

arcz:
## Description
I think the slot fetch cache didn't work as intended because it checked `fetchedContracts` instead of `fetchedStorage`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1649,13 +1649,13 @@ accessStorage addr slot continue = do
         Just x ->
           continue x
         Nothing ->
-          if c._external
-          then
-            -- check if the slot is cached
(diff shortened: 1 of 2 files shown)


arcz:
If this PR makes sense I'll add a test for it too

d-xo:
nice catch. fix seems sensible, happy to merge once we have a test in place

arcz:
@d-xo Test added

d-xo (approved):

d-xo:
Thanks!

PR #181, opened 2023-01-28
Cleanup some records

arcz:
## Description
Some records cleanup thanks to the new extensions.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (commented):
Looks generally great. Curious if the plan is to also remove the underscores from the field names for the EVM records, or do we still have a few usages of lens that mean we have to keep them around indefinitely?

arcz:
The end goal is to have no underscores and no prefixes in records. The EVM records require lens for now as there are many state updates, most of which are nested. To migrate that to clean records we have two options. Either we stick with lens for nested updates with something like generic-lens/optics (making sure there is no performance impact) or we make a few plain dedicated functions to perform nested updates. The first option is less boilerplate and could be considered more idiomatic, the second is more straightforward and contributor friendly (no lens knowledge required).

d-xo:
understood, thanks for the clarification :+1:

PR #183, opened 2023-01-30
Implement debug() cheat code

arcz:
## Description
This cheat code provides a way to stop the execution and hand over the current vm state for debugging. An alternative name to consider - `breakpoint()`.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -235,6 +235,9 @@ oracle solvers info q = do
            Nothing ->
              error ("oracle error: " ++ show q)
 
+    EVM.PleaseDebugMe _vm continue ->
+      pure continue
+
 type Fetcher = EVM.Query -> IO (EVM ())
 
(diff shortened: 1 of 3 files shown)


src/EVM/Fetch.hs
```diff
@@ -235,6 +235,9 @@ oracle solvers info q = do
            Nothing ->
              error ("oracle error: " ++ show q)
 
+    EVM.PleaseDebugMe _vm continue ->
+      pure continue
```
arcz, line 239: This cheat code is useful for programs that use hevm as a library. To consider if it should be enabled for hevm here.

d-xo:
Looks good so far, but can you please also modify the interpreter in TTY to respect this cheatcode

arcz:
I'll close this for now, I have to rethink if this is really necessary.

PR #185, opened 2023-01-31
Distributable Binaries For Macos

d-xo:
## Description

This PR lets us build binaries for macos that only link dynamically against libs that we can be sure will be provided by macos itself. All other non system c dependencies (secp256k1, libff, gmp, ...) are linked statically. This is required since full static linking is not supported on macos.

Cabal doesn't seem to offer friendly config options that would make the production of this kind of binary easy, and in the end we are relying on some behaviour that is at best poorly documented (and quite possibly implicit / unintended):

- ghc / cabal allows users to pass library directories to be searched during linking
- if these library dirs contain both static and dynamic versions of some required lib, it will be linked dynamically
- if they contain only static versions of the lib, it will be linked statically

We can therefore get the behaviour we want by ensuring that we pass only static versions of the libs that we wish to be statically linked. Once this is done, we can do a final piece of denixification to rewrite the library paths for the remaining dynamic libs to point to `/usr/lib/` instead of the nix store, giving us "static" binaries that can run without requiring the installation of additional libraries on any x86 macos system.

I tested the updated release pipeline on a fork: https://github.com/d-xo/hevm/actions/runs/4049153909 which produced binaries that can be tested here: https://github.com/d-xo/hevm/releases/tag/release%2F0.50.3.

output of `otool -L` on the resulting macos binary: 

```
	/usr/lib/libstdc++.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libiconv.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,10 +22,12 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
-        secp256k1-static = pkgs.secp256k1.overrideAttrs (attrs: {
+
+        secp256k1-static = stripDylib (pkgs.secp256k1.overrideAttrs (attrs: {
           configureFlags = attrs.configureFlags ++ [ "--enable-static" ];
-        });
-        hevmUnwrapped = (with pkgs; lib.pipe (
+        }));
+
+        hevmUnwrapped = (with pkgs.pkgs; lib.pipe (
           haskellPackages.callCabal2nix "hevm" ./. {
             # Haskell libs with the same names as C libs...
             # Depend on the C libs, not the Haskell libs.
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -65,17 +70,59 @@
...
+          otool = "${pkgs.darwin.binutils.bintools}/bin/otool";
+          install_name_tool = "${pkgs.darwin.binutils.bintools}/bin/install_name_tool";
+        in if pkgs.stdenv.isLinux
+        then pkgs.haskell.lib.dontCheck hevmWrapped
+        else pkgs.runCommand "stripNixRefs" {} ''
+          mkdir -p $out/bin
+          cp ${pkgs.haskell.lib.dontCheck hevmUnwrapped}/bin/hevm $out/bin/
+
+          # get the list of dynamic libs from otool and tidy the output
+          libs=$(${otool} -L $out/bin/hevm | tail -n +2 | sed 's/^[[:space:]]*//' | cut -d' ' -f1)
```
msooseth, line 90: You are a proper command-line haxxor I see :)

msooseth (approved):
LGTM

minimapletinytools:
goodness 😭

PR #187, opened 2023-01-31
Add constraints asserting that reads beyond buffer size are zero

zoep:
## Description

Adds assertions stating that a read beyond the length of a buffer is zero. It works as follows: 

- Finds all indices `idx` that are read from any buffer `buf`. 
- Filters our indices that are provably within range.
- Adds assertions of the form `idx >= bufLength buf => buf[idx] = 0`.

## Checklist

- [X] tested locally
- [X] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -25,6 +25,7 @@ import qualified Data.Map.Strict as Map
 import qualified Data.Vector as V
 import qualified Data.Vector.Storable as VS
 import Data.Vector.Storable.ByteString
+import Data.Semigroup (Any, Any(..), getAny)
 
 
 -- ** Stack Ops ** ---------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


test/test.hs
```diff
@@ -1600,7 +1600,24 @@ tests = testGroup "hevm"
           assertEqual "w==z for hash collision" w z
           putStrLn "expected counterexample found"
         ,
-        expectFail $ testCase "calldata beyond calldatasize is 0 (z3)" $ do
+        -- testCase "symbolic calldataload" $ do
```
d-xo, line 1603: why is this commented out?

src/EVM/Expr.hs
```diff
@@ -936,3 +959,12 @@ numBranches _ = 1
 
 allLit :: [Expr Byte] -> Bool
 allLit = Data.List.and . fmap (isLitByte)
+
+-- | True if the given expression contains any node that satisfies the
+-- input predicate
+containsNode :: (forall a. Expr a -> Bool) -> Expr b -> Bool
```
d-xo, line 965: nice :)

src/EVM/Expr.hs
```diff
@@ -434,11 +435,33 @@ concPrefix (CopySlice (Lit srcOff) (Lit _) (Lit _) src (ConcreteBuf "")) = do
...
+    go _ (GVar _) = error "Internal error: cannot calculate a concrete prefix of an open expression"
 concPrefix (ConcreteBuf b) = Just (num . BS.length $ b)
 concPrefix e = error $ "Internal error: cannot compute a concrete prefix length for: " <> show e
 
 
+-- | Return the minimum possible length of a buffer. In the case of an
+-- abstract buffer, it is the largest write that is made on a concrete
+-- location. Parameterized by an environment for buffer variables.
+minLength :: Map.Map Int (Expr Buf) -> Expr Buf -> Maybe Integer
+minLength senv = go 0
```
d-xo, line 447: I think the name is a little unclear here?

```suggestion
minLength bufEnv = go 0
```

src/EVM/SMT.hs
```diff
@@ -221,9 +216,44 @@ findStorageReads = foldProp go []
...
+
+
+-- | Asserts that buffer reads beyond the size of the buffer are equal
+-- to zero. Looks for buffer reads in the a list of given predicates
+-- and the buffer and storage environments.
+assertReads :: [Prop] -> BufEnv -> StoreEnv -> [Prop]
+assertReads props benv senv = concat $ fmap assertRead allReads
+  where
+    assertRead :: (Expr EWord, Expr EWord, Expr Buf) -> [Prop]
+    assertRead (idx, Lit 32, buf) = [POr (PEq (ReadWord idx buf) (Lit 0)) (PNeg (PGEq idx (bufLength buf)))]
```
d-xo, line 244: Is it maybe worth adding an implication operation to `Prop`? I think this might be a little more readable with a `PImpl`?

src/EVM/SMT.hs
```diff
@@ -221,9 +216,44 @@ findStorageReads = foldProp go []
...
+    assertRead (idx, Lit sz, buf) = fmap (\s -> POr (PEq (ReadByte idx buf) (LitByte (num s))) (PNeg (PGEq idx (bufLength buf)))) [(0::Int)..num sz-1]
+    assertRead (_, _, _) = [] -- cannot generate assertions for accesses of symbolic size
```
d-xo, line 246: Is it sound to ignore this case? Should we be erroring out instead?

src/EVM/SMT.hs
zoep: Oh nice! Done!

src/EVM/SMT.hs
zoep: I'd say that it's sound but incomplete. If there is a copySlice with symbolic size, we will not be able to prove that some accesses are zero (however, we will not prove something that is not true, so in this sense it is not unsound). I'm fine with throwing an error here instead though.

I think this is the only case that this will show up is`copySlice` with symbolic size. This is unimplemented in SMT.hs so it will throw an error there anyway. Also, I think that such cases cause an infinite loop in symbolic execution.

d-xo: hmmmm yeah all good points, I guess I kinda feel like it's good to error out here just so that we kinda have a lil todo list for when we do start supporting fully dynamic data

d-xo (approved):
looks good

src/EVM/SMT.hs
zoep: Done!

PR #188, opened 2023-01-31
Fix small discrepancies between CLI and `checkAssert`

zoep:
## Description

Fixes some tiny discrepancies between `assert` in cli.hs and `checkAssert`/`verifyContract` in `SymExec.hs` used by unit tests:

- Use abstract base buffer when signature is provided in `assert`.
- In `checkAssert`, add upper bound constraint `txdata <= 2^64` for calldata even if no signature is provided.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
looks good!

PR #189, opened 2023-02-01
Remove some underscores from EVM records

arcz:
## Description
Lens are not used for those fields so we can get rid of some underscored without much problem.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Note: This will clash with https://github.com/ethereum/hevm/pull/202 I don't mind fixing the merge conflicts, just an FYI :) 

BTW, can we remove lenses now that we have the `.` notation? That might be more interesting than removing the `_`? Or are these `makeLenses` really needed?

d-xo:
we are still using lenses in some of the EVM records to handle nested updates and so on.

msooseth:
Ah, I see, thanks!

d-xo:
superceded by https://github.com/ethereum/hevm/pull/244

PR #190, opened 2023-02-07
Add fetchChainIdFrom

arcz:
## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -184,6 +184,11 @@ fetchSlotFrom n url addr slot = do
   sess <- Session.newAPISession
   fetchSlotWithSession n url sess addr slot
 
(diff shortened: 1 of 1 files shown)


d-xo:
what is this for?

arcz:
It's a function to fetch chain id, it was already partially done with `QueryChainId` but this function was missing.

arcz:
@d-xo it is not used by hevm itself but it is useful if you use hevm as a library, I duplicated this function in echidna:
https://github.com/crytic/echidna/blob/bdbf2bba045b7ceffc0fd319839d3737a56f1cbd/lib/Echidna/RPC.hs#L49-L56

d-xo:
got it, thanks!

PR #191, opened 2023-02-10
Buffer encoding without uninterpreted functions  

zoep:
## Description

Change buffer encoding so that: 

- A symbolic expression for buffer length is computed all the way down to base buffers (change in `bufLength` definition)
- Only one abstract variable is needed, to represent the length of abstract base buffers, therefore the uninterpreted function `bufLength` can be removed from the encoding. 

Also, a new constraint is added to the encoding that asserts that the length of a buffer cannot be larger than the maximum read location (because there is no need to). This will give an upper bound to the returned counter examples. 

This is a WIP as I also want to evaluate the performance impact of the new encoding (preliminary experiments show no slowdown and potentially speedup). I also want to evaluate the effect this change has on the size of the counterexamples.

Builds on top of #187.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -793,7 +793,7 @@ simplify e = if (mapExpr go e == e)
     go (EVM.Types.Not (EVM.Types.Not a)) = a
 
     go (Max (Lit 0) a) = a
-    go (Min (Lit 0) a) = Lit 0
+    go (Min (Lit 0) _) = Lit 0
     
     go a = a
 
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -403,10 +403,10 @@ writeWord offset val src = WriteWord offset val src
 -- If there are any writes to abstract locations, or CopySlices with an
 -- abstract size or dstOffset, an abstract expresion will be returned.
 bufLength :: Expr Buf -> Expr EWord
-bufLength = bufLengthEnv mempty
+bufLength = bufLengthEnv mempty True
 
-bufLengthEnv :: Map.Map Int (Expr Buf) -> Expr Buf -> Expr EWord
-bufLengthEnv env buf = go (Lit 0) buf
+bufLengthEnv :: Map.Map Int (Expr Buf) -> Bool -> Expr Buf -> Expr EWord
+bufLengthEnv env useEnv buf = go (Lit 0) buf
   where
     go :: Expr EWord -> Expr Buf -> Expr EWord
     go (Lit l) (ConcreteBuf b) = Lit $ max (num . BS.length $ b) l
@@ -423,10 +423,11 @@ bufLengthEnv env buf = go (Lit 0) buf
     go l (WriteWord idx _ b) = go (Max l (Add idx (Lit 32))) b
     go l (WriteByte idx _ b) = go (Max l (Add idx (Lit 1))) b
     go l (CopySlice _ dstOffset size _ dst) = go (Max l (Add dstOffset size)) dst
-    go l (GVar (BufVar a)) =
+    go l (GVar (BufVar a)) | useEnv =
       case Map.lookup a env of
         Just b -> go l b
         Nothing -> error "Internal error: cannot compute length of open expression"
+    go l (GVar (BufVar a)) = Max l (BufLength (GVar (BufVar a)))
 
 -- | If a buffer has a concrete prefix, we return it's length here
 concPrefix :: Expr Buf -> Maybe Integer
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -793,7 +793,7 @@ simplify e = if (mapExpr go e == e)
     go (EVM.Types.Not (EVM.Types.Not a)) = a
 
     go (Max (Lit 0) a) = a
-    go (Max a (Lit 0)) = a
+    go (Min (Lit 0) a) = Lit 0
     
     go a = a
 
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -959,12 +959,10 @@ eqByte (LitByte x) (LitByte y) = Lit $ if x == y then 1 else 0
 eqByte x y = EqByte x y
 
 min :: Expr EWord -> Expr EWord -> Expr EWord
-min (Lit x) (Lit y) = if x < y then Lit x else Lit y
-min x y = Min x y
+min x y = op2 Min Prelude.min x y
 
 max :: Expr EWord -> Expr EWord -> Expr EWord
-max (Lit x) (Lit y) = if x > y then Lit x else Lit y
-max x y = Min x y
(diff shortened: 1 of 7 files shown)


src/EVM/SMT.hs
```diff
@@ -378,7 +411,7 @@ prelude =  (flip SMT2) mempty $ fmap (fromLazyText . T.drop 2) . T.lines $ [i|
   )
 
   ; buffers
-  (declare-fun bufLength (Buf) Word)
+  ; (declare-fun bufLength (Buf) Word)
```
d-xo, line 414: I think we can just delete this line entirely?

src/EVM/SMT.hs
```diff
@@ -254,12 +261,36 @@ assertReads props benv senv = concat $ fmap assertRead allReads
         _ -> True
     keepRead _ = True
 
+-- | Asserts that the length of an abstract base buffer is at most the maximum location that is read
+assertMaxLen :: [Prop] -> BufEnv -> StoreEnv -> [Prop]
```
d-xo, line 265: really nice

src/EVM/Expr.hs
```diff
@@ -950,8 +960,10 @@ eqByte (LitByte x) (LitByte y) = Lit $ if x == y then 1 else 0
 eqByte x y = EqByte x y
 
 min :: Expr EWord -> Expr EWord -> Expr EWord
-min (Lit x) (Lit y) = if x < y then Lit x else Lit y
-min x y = Min x y
+min x y = op2 Min Prelude.min x y
```
d-xo, line 963: I think we can use `normArgs` here to make the matches in `simplify` easier to handle?

src/EVM/SMT.hs
zoep: Yes! Done!

src/EVM/Expr.hs
zoep: Done!

src/EVM/SMT.hs
```diff
@@ -254,12 +261,36 @@ assertReads props benv senv = concat $ fmap assertRead allReads
...
+assertMaxLen props benv senv = fmap (\(k, v) -> PLEq (BufLength (AbstractBuf k)) v) $ Map.toList bufMap
+  where
+    allReads = nubOrd $ findBufferAccess props <> findBufferAccess (Map.elems benv) <> findBufferAccess (Map.elems senv)
+
+    bufMap = foldl addBound mempty allReads
+
+    addBound m (idx, size, buf) =
+      case baseBuf buf of
+        AbstractBuf b -> Map.insertWith EVM.Expr.max b (add idx size) m
```
msooseth, line 274: Wow, this is some very tightly written code. Nice, also scary good :smiley:

msooseth (approved):
Wow, that's some tightly written code, nice!

src/EVM/SMT.hs
```diff
@@ -95,13 +95,16 @@ declareIntermediates bufs stores =
       encBs = Map.mapWithKey encodeBuf bufs
       sorted = List.sortBy compareFst $ Map.toList $ encSs <> encBs
       decls = fmap snd sorted
-  in SMT2 ([fromText "; intermediate buffers & stores"] <> decls) mempty
+  in SMT2 ([fromText "; intermediate buffers & stores"] <> decls
+          ) mempty
```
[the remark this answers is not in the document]
zoep: Oh nice catch!

test/test.hs
```diff
@@ -581,8 +581,8 @@ tests = testGroup "hevm"
               }
              }
             |]
-        (_, [Cex (_, ctr)]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s [0x32] c (Just ("fun(uint8)", [AbiUIntType 8])) [] defaultVeriOpts
-        assertBool "Access must be beyond element 2" $ (getVar ctr "arg1") > 1
+        (_, [Cex (_, _)]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s [0x32] c (Just ("fun(uint8)", [AbiUIntType 8])) [] debugVeriOpts
+        -- assertBool "Access must be beyond element 2" $ (getVar ctr "arg1") > 1
```
d-xo, line 585: why is this commented out?

test/test.hs
```diff
@@ -1637,7 +1637,7 @@ tests = testGroup "hevm"
               }
             }
             |]
-          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c Nothing [] defaultVeriOpts
+          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c Nothing [] debugVeriOpts
```
d-xo, line 1640:
```suggestion
          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c Nothing [] defaultVeriOpts
```

test/test.hs
```diff
@@ -1654,7 +1654,7 @@ tests = testGroup "hevm"
               }
             }
             |]
-          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c (Just ("f(uint256)", [AbiUIntType 256])) [] defaultVeriOpts
+          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c (Just ("f(uint256)", [AbiUIntType 256])) [] debugVeriOpts
```
d-xo, line 1657:
```suggestion
          (res, [Qed _]) <- withSolvers Z3 1 Nothing $ \s -> checkAssert s defaultPanicCodes c (Just ("f(uint256)", [AbiUIntType 256])) [] defaultVeriOpts
```

src/EVM/Expr.hs
```diff
@@ -785,6 +792,9 @@ simplify e = if (mapExpr go e == e)
     -- Double NOT is a no-op, since it's a bitwise inversion
     go (EVM.Types.Not (EVM.Types.Not a)) = a
 
+    go (Max (Lit 0) a) = a
+    go (Max a (Lit 0)) = a
```
d-xo, line 796: I think we don't need this match anymore since you used `normArgs` for `min` and `max` below?

test/test.hs
zoep: It is because of this [issue](https://github.com/ethereum/hevm/issues/194). Apparently, there are models where "arg1 > 1" and it a different instance of overflow and wraparound of addresses.

test/test.hs
zoep: Done!

src/EVM/Expr.hs
zoep: Done! Also added the corresponding case for Min

d-xo (approved):
looks great :)

PR #192, opened 2023-02-14
Bump flake.lock

arcz:
## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
looks like cvc4 is broken on linux in this nixpkgs commit? I'm actually not sure why it's in the closure since afaik we only depend on cvc5 these days...

d-xo:
hm looks like solc is broken now?

arcz:
Yeah, solc is broken for a while now :/

Issue #193, opened 2023-02-14
Simplification error when addresses wrap around in `copySlice`


zoep:
The encoding of the expression 
```let e = CopySlice (Lit 0x0) (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff) (Lit 0x2) (ConcreteBuf "\1\1\1") (ConcreteBuf "\0\0\0")```

is `(store (store ...dstBuf... 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff 0x1) 0x0 0x1)`

Essentially, it copies the first two bytes of the source buffer "\1\1\1" to the addresses `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` and `0x0` of the destination buffer '\0\0\0". Note that because of overflow, the addresses of the destination buffer wrap around.

However `simplify (ReadByte (Lit 0x0) e)` returns `0x0`, since our simplifier directly reads the byte at address `0x0` from the destination buffer as it considers that the copied slice is irrelevant from the address `0x0` (ignoring the overflow that happens).

This causes some failures in fuzzing. 

It can be fixed either by disallowing wraparound of addresses in copySlice, or by taking into account address wraparound in our simplification.

zoep:
This can be reproduced in the main branch with `cabal run test -- -p '/byte-simplification/' --quickcheck-replay=31717`.

PR #200, opened 2023-02-17
Expr: restrict readByte and readWord simplification when there is ove…

zoep:
## Description

Suggested fix for #193. 

Sometimes, in `CopySlide srcOff dstOff size src dst`, it may be that `dstOff + size` overflows and wraps around. If we are reading some byte `x` from this buffer such that `0 <= x < dstOff + size` our simplifier will read it directly from the destination buffer thinking that the read is out of range. 

This shows up in fuzz tests and causes CI to fail.

This PR prevents that by disallowing simplifications when there is an overflow. However, it restricts simplification. For example, if we do `readByte (Lit x) (CopySlide srcOff (Lit dstOff) size src dst)` with `x < dstOff` and a potentially symbolic size, we no longer simplify this to `readByte (Lit x) dst`. 

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
looks good!

PR #201, opened 2023-02-18
Windows build support

elopez:
## Description
This PR makes the hevm library buildable on Windows systems. Unfortunately `vty` does not support Windows, so the components with a `brick` UI have to be disabled for a successful build.

This can be seen working on this build of Echidna: https://github.com/crytic/echidna/pull/943

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
Interesting. Have you been able to run this? How hard do you think it would be to setup a ci pipeline here?

elopez:
We've got `echidna.exe` building and working now, which uses the hevm library 😄  You can try it out with [binaries from our CI](https://github.com/crytic/echidna/actions/runs/4209767072). I did not try to run the hevm executable itself though -- from what I can tell, it uses `EVM.TTY` which is not buildable on Windows.

We can likely add a CI pipeline here -- it would require:
1. windows runner (GH has those)
1. setting up msys2 (for gmp, building secp256k1, ff)
2. setting up GHC
3. building the dependencies
4. building hevm library

I'll give it a try and report back.

elopez:
@d-xo I added a CI job for Windows now

d-xo:
Thanks, the CI pipeline looks pretty reasonable, and the diff here is suprisingly small. I'm wondering if there is a clean way to ensure that the libff / libsecp256k1 versions we're pulling in the windows pipeline match those fetched via nix in the mac / linux pipelines.

In general I'm pretty scared of the maintainance and support burden here, but probably OK to merge on a provisional basis and see how things go. Did you run into issues when running echidna under windows? Are you willing to fix windows specific issues in hevm on an ongoing basis?

elopez:
I'll defer to @arcz for the first question -- I'm not aware of any way to subscribe to version changes in Nix. Fortunately these two dependencies don't have particularly frequent releases, so it should not be too much of a burden to check once in a while if that's needed.

As for issues, I encountered a few echidna-specific ones, mainly with text encoding when printing to the terminal, but nothing major. I'm also pleasantly surprised about how well everything worked out of the box. If you look at the echidna diff, most of the changes are tweaks in the build scripts and CI, and some changes to be able to disable the `brick` UI. I didn't have to tweak anything in the fuzzing core or hevm integration.

I'll be happy to take a look if any Windows issues crop up in hevm and help on a best-effort basis; feel free to tag me in issues or PRs 👍

d-xo:
> I'll be happy to take a look if any Windows issues crop up in hevm and help on a best-effort basis; feel free to tag me in issues or PRs

Awesome <3. Some things that we may want to consider for future prs would be to build and run the test suite (or some subset) as part of the windows ci job, and to produce distributable native windows binaries on each release.

> I'll defer to @arcz for the first question -- I'm not aware of any way to subscribe to version changes in Nix.

On my linux machine I can do the following from the hevm root, I'm not sure how easy this would be to translate to windows land, perhaps this can be achieved using nix installed via WSL?

```
$ NIXPKGS_REV=$(cat flake.lock | jq -r '.nodes.nixpkgs.locked.rev'
$ nix search "nixpkgs/$NIXPKGS_REV" "^legacyPackages.x86_64-linux.libff$" --json 2>/dev/null | jq '."legacyPackages.x86_64-linux.libff".version' -r
0.2.1
```

Also regarding WSL, I'm curious what you see as the main advantages of supporting native windows vs requiring windows users of hevm / echidna to use WSL? I'm not familiar enough with windows to understand the tradeoffs here.

elopez:
> On my linux machine I can do the following from the hevm root, I'm not sure how easy this would be to translate to windows land, perhaps this can be achieved using nix installed via WSL?
> 

We don't _need_ to do the check on Windows, so using nix on Linux is a good approach 😄 Your suggestion worked for libff but did not reveal the correct version for `secp256k1` (it has some arbitrary date set, as it's unstable on the pinned nixpkgs) so I went with checking the git rev directly through `nix eval`.

> Also regarding WSL, I'm curious what you see as the main advantages of supporting native windows vs requiring windows users of hevm / echidna to use WSL? I'm not familiar enough with windows to understand the tradeoffs here.

Personally I like WSL and don't normally install dev tools outside of it 😅 but for people that prefer having a dev environment on native Windows (e.g. Python tools, NodeJS, etc.) it's easier to keep all the tools on the same "side" of the OS. Mixing and matching tools on Windows & WSL land is often cause for trouble, and the solution tends to not be elegant (i.e. install everything twice, once in WSL and once outside). Besides, WSL is something extra you have to install and setup.

.github/scripts/install-libff.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+if [ -f "$HOME/.local/lib/libff.a" ]; then
+  echo "libff exists, exiting..."
+  exit 0
+fi
+
+if [ -d libff ]; then
+  echo "$(pwd)/libff" already exists! Using it instead of re-cloning the repo.
+else
```
d-xo, line 20: Are we expecting these lines to ever be executed in our ci env?

.github/scripts/install-libff.sh
```diff
@@ -0,0 +1,34 @@
...
+
+sed -i 's/find_library(GMP_LIBRARY gmp)/find_library(GMP_LIBRARY NAMES libgmp.a)/' CMakeLists.txt
+PREFIX="$HOME/.local"
+ARGS=("-DCMAKE_INSTALL_PREFIX=$PREFIX" "-DWITH_PROCPS=OFF" "-G" "Ninja")
+CXXFLAGS="-fPIC"
+
+mkdir -p build
+cd build
+CXXFLAGS="$CXXFLAGS" cmake "${ARGS[@]}" ..
+cmake --build . && cmake --install .
```
d-xo, line 34: total nit, but we're missing a final newline here

.github/scripts/install-libsecp256k1.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [[ "$(uname -s)" =~ ^MSYS_NT.* ]]; then
+    echo "This script is only meant to run on Windows under MSYS2"
+    exit 1
+fi
+
+if [ -f "$HOME/.local/lib/libsecp256k1.a" ]; then
+  echo "libsecp256k1 exists, exiting..."
+  exit 0
+fi
```
d-xo, line 16: are we expecting this line to ever be executed in our ci env?

d-xo (commented):
running the version check on linux is a pretty neat solution! just a few very small nits and then I think we're good to merge :)

.github/scripts/install-libff.sh
elopez: I don't think so, as the CI runs on a clean slate each time, but it's handy if you're (re)running the script locally. Do you want me to drop it so it's a bit simpler?

.github/scripts/install-libsecp256k1.sh
elopez: Also no, I'll drop these ones

.github/scripts/install-libff.sh
elopez: will fix 👍

d-xo:
Thanks!

PR #202, opened 2023-02-21
Trace Fuzzing

msooseth:
## Description
This is a set of changes to execute trace fuzzing against `go-ethereum`'s `evmtool`. The major changes are:
- new signing setup to allow signing contracts with the correct `v` value
- random contract generation. Notice that this lacks external calls (they are stripped away currently) and jumpdests are fixed up to be always correct. Also, there is always enough gas, etc.
- trace generation
- Assembler code
- upgrade to more sane base16 package
- improved JSON encoding & decoding of some structs
- new test in `test.hs` to run concrete trace generation against `evmtool` over a random contract & calldata

To check symbolic execution it also:
- Runs concrete code with symbolic interpreter, with symbolic calldata
- Replaces the symbolic calldata with concrete calldata
- Calls `Expr.simplify` on this expression
- Checks the `Return _ (Concretebuf bs) _)`'s `bs` of this expression against the output by `evmtool`.

### Major todos to improve:
1.  Use https://github.com/MariusVanDerWijden/FuzzyVM to generate contracts & calldata & etc.
3. Check symbolic execution through SMT via the above point but also generating an SMT query from it and adding the constant calldata to the SMT query, and get back the concrete output as a solution.

### Minor todos to improve:
- Add better  calldata&contract generation. For example, `sizen` is decidedly weird it start small and increases... why?
- Add better fuzzing for parameters to the fuzzer, e.g. gas limit, etc.
- Make external calls work in the fuzzer -- currently `evmtool` does something different when external contracts are called.

###  Notes on some structural changes
* added `go-ethereum` as a dependency, which is needed to have `evmtool`
* added `base16` package and removed `base16-bytestring` package. The former is way more powerful than the latter
* added a shell script to convert `jsonl` trace produced by `evmtool`. This shell script uses `awk`, `wc` `cut`, and `bash`

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


src/EVM.hs
```diff
@@ -1997,50 +1997,28 @@ cheatActions =
         \sig outOffset _ input -> case decodeStaticArgs 0 2 input of
           [sk, hash] ->
             forceConcrete2 (sk, hash) "cannot sign symbolic data" $ \(sk', hash') -> let
-              curve = getCurveByName SEC_p256k1
-              priv = PrivateKey curve (num sk')
-              digest = digestFromByteString (word256Bytes hash')
             in do
```
zoep, line 2000: It seems that there is an empty `let in` here after removing the declarations.

test/test.hs
```diff
@@ -2342,6 +2710,16 @@ instance Arbitrary (Expr Buf) where
 instance Arbitrary (Expr End) where
   arbitrary = sized genEnd
 
+data OpContract = OpContract [Op]
```
zoep, line 2713: This is also defined in Debug.hs. Is this intentional?

msooseth: Oh, yeah, that was wrong, thanks! I used it to debug. I removed it from `Debug.hs` now :)

msooseth:
I  have now updated this PR with a minior (~30) line addition to also add semi-symbolic execution check. 

This pass adds to tracing the following:
- Runs concrete code with symbolic interpreter, with symbolic calldata
- Replaces the symbolic calldata with concrete calldata
- Calls `Expr.simplify` on this expression
- Checks the `Return _ (Concretebuf bs) _)`'s `bs` of this expression against the output by `evmtool`.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Added
+- New concrete fuzzing setup based on `evmtool` in `go-ethereum` and a new tracing executor, based on old dapptools trace executor. This also adds a simplistic random contract generation, and improves the signing infrastructure
```
d-xo, line 9: May
```suggestion
- A new differential fuzzing test harness that compares the concrete semantics, as well as parts of the symbolic semantics against the geth evm implementation
```

convert_trace_to_json.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
d-xo, line 1: Can we either do this in haskell or tidy the script away under `test` somewhere so it's not cluttering up the root directory.

src/EVM/Assembler.hs
```diff
@@ -0,0 +1,104 @@
+{-# LANGUAGE DataKinds #-}
```
d-xo, line 1: Not a big deal, but it would be nice to have a [module description](https://haskell-haddock.readthedocs.io/en/latest/markup.html#the-module-description) here. Just the module and name fileds is fine.

src/EVM/Sign.hs
```diff
@@ -0,0 +1,72 @@
+module EVM.Sign where
```
d-xo, line 1: A module description would be great here too

src/EVM/Transaction.hs
```diff
@@ -5,33 +5,39 @@ import Prelude hiding (Word)
...
-} deriving Show
+} deriving (Show, Generic)
+instance JSON.ToJSON AccessListEntry
 
 data TxType = LegacyTransaction
             | AccessListTransaction
             | EIP1559Transaction
-  deriving (Show, Eq)
+  deriving (Show, Eq, Generic)
+instance JSON.ToJSON TxType where
```
d-xo, line 37: just a lil nit, but would imo be more readable with a newline to separate these

src/EVM/Transaction.hs
```diff
@@ -47,25 +53,63 @@ data Transaction = Transaction {
     txAccessList :: [AccessListEntry],
     txMaxPriorityFeeGas :: Maybe W256,
     txMaxFeePerGas :: Maybe W256
-} deriving Show
+} deriving (Show, Generic)
+instance JSON.ToJSON Transaction where
```
d-xo, line 57: same here re: newline

src/EVM/Transaction.hs
```diff
@@ -5,33 +5,39 @@ import Prelude hiding (Word)
...
 
 data TxType = LegacyTransaction
             | AccessListTransaction
             | EIP1559Transaction
-  deriving (Show, Eq)
+  deriving (Show, Eq, Generic)
+instance JSON.ToJSON TxType where
+  toJSON t = case t of
+               EIP1559Transaction -> "0x2"
+               _ -> error "unimplemented"
```
d-xo, line 40: can we fill out the other types here

src/EVM/Transaction.hs
```diff
@@ -47,25 +53,63 @@ data Transaction = Transaction {
...
+                         , ("r",                 (JSON.toJSON $ show $ t.txR))
+                         , ("s",                 (JSON.toJSON $ show $ t.txS))
+                         , ("to",                (JSON.toJSON $ t.txToAddr))
+                         , ("nonce",             (JSON.toJSON $ show $ t.txNonce))
+                         , ("value",             (JSON.toJSON $ show $ t.txValue))
+                         , ("type",              (JSON.toJSON $ t.txType))
+                         , ("accessList",        (JSON.toJSON $ t.txAccessList))
+                         , ("maxPriorityFeePerGas", (JSON.toJSON $ show $ fromJust $ t.txMaxPriorityFeeGas))
+                         , ("maxFeePerGas",      (JSON.toJSON $ show $ fromJust $ t.txMaxFeePerGas))
+                         , ("chainId",           (JSON.toJSON ("0x1" :: String))) -- NOTE: should be part of struct?
```
d-xo, line 71: what does this comment mean?

src/EVM/Transaction.hs
```diff
@@ -47,25 +53,63 @@ data Transaction = Transaction {
...
 sender :: Int -> Transaction -> Maybe Addr
 sender chainId tx = ecrec v' tx.txR  tx.txS hash
   where hash = keccak' (signingData chainId tx)
         v    = tx.txV
         v'   = if v == 27 || v == 28 then v
                else 27 + v
 
+-- We don't know which of 27 or 28 is correct
+-- So we try 28 and try to recover & check if it's correct
+-- if it isn't correct, we need to use the other one
```
d-xo, line 106: This comment is outdated now that this logic has been moved into `Sign.hs`

src/EVM/Types.hs
```diff
@@ -464,14 +466,22 @@ instance Show ByteStringS where
       fromBinary =
         Text.decodeUtf8 . toStrict . toLazyByteString . byteStringHex
 
-instance JSON.ToJSON ByteStringS where
-  toJSON = JSON.String . Text.pack . show
+instance JSON.FromJSON ByteString where
```
d-xo, line 469: I'm a little nervous that we will have issues related to [orphan instances](https://wiki.haskell.org/Orphan_instance) if one of our deps defines the same instance. Would it be very annoying to use the `ByteStringS` newtype above for json serialization of transactions?

src/EVM/Assembler.hs
msooseth: Thanks! Yeah, that's a great idea. I just added it in a commit, will push and resolve ;)

src/EVM/Sign.hs
msooseth: Good point, added it as well, will push and resolve :)

src/EVM/Transaction.hs
msooseth: Yeah, thanks, good point! I fixed it in all places now :) Will push and resolve!

src/EVM/Transaction.hs
msooseth: Yes, good point, I just did :)

src/EVM/Transaction.hs
msooseth: The field `chainID` is not part of the Transaction datatype. It's the only element needed for this datatype to be 1-to-1 usable in `evmtool` as a transaction. I wrote this comment to remind myself that I should discuss with someone whether this should be made part of the datatype or if it's OK as-is, added to the `ToJSON` but not actually be part of the datatype.

What do you think? I think in many (or only some?) places in HEVM, chainID is assumed to be `1`. So I thought maybe it's OK to simply sneak this in there.

msooseth: Hmmm... actually, many places where `Transaction` is being used, `chainId` also needs to be provided. So maybe we could make it part of the datatype? what do you think?

src/EVM/Transaction.hs
msooseth: Thanks, I removed it. I think it's pretty self-explanatory now, so I think it doesn't need a replacement?

test/test.hs
```diff
@@ -70,6 +92,336 @@ main = defaultMain tests
...
+  EVMToolTrace
+    { evmPc :: Int
+    , evmOp :: Int
+    , evmGas :: W256
+    , evmMemSize :: Integer
+    , evmDepth :: Int
+    , evmRefund :: Int
+    , evmOpName :: String
+    , evmStack :: [W256]
+    , evmError :: Maybe String
```
d-xo, line 105: I don't think we need these prefixes

test/test.hs
```diff
@@ -70,6 +92,336 @@ main = defaultMain tests
...
+    } deriving (Generic, Show)
+
+instance JSON.FromJSON EVMRejected where
+  parseJSON = JSON.withObject "EVMRejected" $ \v -> EVMRejected
+    <$> v .: "index"
+    <*> v .: "error"
+
+data EVMToolAlloc =
+  EVMToolAlloc
+  { balance :: W256 -- this is wrongly written by default, hence the non-default ToJSON
```
d-xo, line 246:
```suggestion
  { balance :: W256
```

test/test.hs
```diff
@@ -70,6 +92,336 @@ main = defaultMain tests
...
+                         , ("nonce"  , (JSON.toJSON $ b.nonce))
+                         ]
+
+emptyEVMToolAlloc :: EVMToolAlloc
+emptyEVMToolAlloc = EVMToolAlloc { balance = 0
+                                 , code = mempty
+                                 , nonce = 0
+                                 }
+
+getEVMToolRet :: OpContract -> ByteString -> Int -> IO (Maybe EVMToolResult, (Either (EVM.Error, [VMTrace]) (Expr 'End, [VMTrace], VMTraceResult)))
```
d-xo, line 263: this function seems to be doing more than it's name suggests, and the return type is pretty hard to parse. perhaps this could be tidied a little?

test/test.hs
```diff
@@ -1233,17 +1650,17 @@ tests = testGroup "hevm"
...
             |]
-          let pre vm = (Lit 0) .== view (state . callvalue) vm
+          let pre vm = (Lit 0) .== Control.Lens.view (state . callvalue) vm
               post prestate poststate =
                 let (x,y) = case getStaticAbiArgs 2 prestate of
                         [x',y'] -> (x',y')
                         _ -> error "expected 2 args"
-                    this = Expr.litAddr $ view (state . codeContract) prestate
-                    prestore =  view (env . storage) prestate
+                    this = Expr.litAddr $ Control.Lens.view (state . codeContract) prestate
```
d-xo, line 1658: Can we maybe make this a little more compact by fiddling with the import defintions?

msooseth:
TODO: make this into a separate file from `test.sh`

src/EVM/Transaction.hs
msooseth: Yep, much cleaner, actually! I managed to remove a bunch of parameters being passed around

src/EVM/Types.hs
msooseth: This has now been fixed :) Thanks for pointing this out! :)

test/test.hs
```diff
@@ -70,6 +91,285 @@ main = defaultMain tests
...
+    <$> v .: "pc"
+    <*> v .: "op"
+    <*> v .: "gas"
+    <*> v .: "memSize"
+    <*> v .: "depth"
+    <*> v .: "refund"
+    <*> v .: "opName"
+    <*> v .: "stack"
+
+genBlockHash:: Int -> Expr 'EWord
```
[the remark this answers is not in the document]
msooseth: Yep, good point, fixed. The `Gen` indeed should reserved for module generating fuzz tests

src/EVM/Assembler.hs
```diff
@@ -0,0 +1,114 @@
+{-|
+Module      : Assembler for EVM opcodes used in the HEVM symbolic checker
+Description :
+Copyright   : (c) d-xo, msoos, 2023
+License     : GPL-3
+Maintainers : see https://github.com/ethereum/hevm
+Stability   : experimental
+Portability : Architecture-independent
```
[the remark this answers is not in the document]
msooseth: Fixed!

src/EVM/Sign.hs
```diff
@@ -0,0 +1,82 @@
+{-|
+Module      : Helper functions to sign a transaction and derive address from
+Description :        for the EVM given a secret key
+Copyright   : (c) d-xo, msoos, 2023
+License     : GPL-3
+Maintainers : see https://github.com/ethereum/hevm
+Stability   : experimental
+Portability : Requires a sec256k1 library, which is architecture-dependent
```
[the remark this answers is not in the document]
msooseth: Fixed!

test/test.hs
```diff
@@ -70,6 +92,319 @@ main = defaultMain tests
...
+data EVMToolEnv = EVMToolEnv
+  { _coinbase    :: Addr
+  , _timestamp   :: Expr EWord
+  , _number      :: W256
+  , _prevRandao  :: W256
+  , _gasLimit    :: Data.Word.Word64
+  , _baseFee     :: W256
+  , _maxCodeSize :: W256
+  , _schedule    :: FeeSchedule.FeeSchedule Data.Word.Word64
+  , _blockHashes :: Map.Map Int W256
```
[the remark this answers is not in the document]
msooseth: Actually, no! I removed them now, yay! :) I think when I was writing this code, we haven't yet enabled some GHC flags, and it was not possible to have them without `_`. I had to merge with `main` to get this functionality. I tried using it before, but it lead to extremely complicated errors from GHC, so I gave up after multiple tries. The trick was to merge with `main` and then the functionality was enabled :)

test/test.hs
```diff
@@ -70,6 +92,319 @@ main = defaultMain tests
...
+    , recRoot :: String
+    , recStatus :: String
+    , recCumulativeGasUsed :: String
+    , recLogsBloom :: String
+    , recLogs :: Maybe String
+    , recTransactionHash :: String
+    , recContractAddress :: String
+    , recGasUsed :: String
+    , recBlockHash :: String
+    , recTransactionIndex :: String
```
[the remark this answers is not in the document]
msooseth: Unfortunately, `type` here is a reserved keyword, so I can't all make it "clean". I can make all of them clean, except for `recType`. What should we do?

test/test.hs
msooseth: Fixed  :)

test/test.hs
msooseth: Thanks, yeah, it was a huge mess. I now cut it into two functions. Should be a lot cleaner :)

src/EVM/Transaction.hs
msooseth: @d-xo Let's discuss about making it a `Maybe`?

test/test.hs
msooseth: Let me know what you think of the new setup!

test/test.hs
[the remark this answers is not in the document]
msooseth: Todo `_type` then when new GHC comes out, we can fix.

test/test.hs
msooseth: Todo `_error` then when new GHC comes out, we can fix.

d-xo (approved):
looks great! feel free to merge once ci is green :)

test/test.hs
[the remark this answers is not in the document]
msooseth: OK, made it into `_type` as discussed :)

test/test.hs
msooseth: Fixed back to what it was originally :)

src/EVM/Transaction.hs
msooseth: So it's staying in as-is as discussed, and we'll make it a Maybe in case it needs to be one. I'm resolving this because we agreed on this already!

PR #204, opened 2023-02-23
Add support for function pointers

samalws:
## Description
This PR adds support for function pointers in the ABI. Function pointers are represented with a 24-byte bytestring, and act very similarly to `AbiBytes 24`.

## Checklist

- [X] tested locally
- [X] added automated tests
- [ ] updated the docs (nothing to add)
- [X] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - The `hevm` library can now be built on Windows systems.
+- Support for function pointers in ABI
 
 ## [0.50.3] - 2023-02-17
 
--- a/src/EVM/ABI.hs
+++ b/src/EVM/ABI.hs
@@ -101,6 +101,7 @@ data AbiValue
   | AbiArrayDynamic AbiType (Vector AbiValue)
   | AbiArray        Int AbiType (Vector AbiValue)
   | AbiTuple        (Vector AbiValue)
+  | AbiFunction     BS.ByteString
   deriving (Read, Eq, Ord, Generic)
 
 -- | Pretty-print some 'AbiValue'.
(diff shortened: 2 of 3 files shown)


src/EVM/ABI.hs
```diff
@@ -516,6 +533,7 @@ parseAbiValue (AbiArrayType n typ) =
   AbiArray n typ <$> do a <- listP (parseAbiValue typ)
                         return $ Vector.fromList a
 parseAbiValue (AbiTupleType _) = error "tuple types not supported"
+parseAbiValue AbiFunctionType = error "function types not supported"
```
d-xo, line 536: Is there a reason for this? I puls prefer to have it implemented unless it's extremely inconvenient

d-xo (commented):
Thanks! Looks great generally, just one comment :)

src/EVM/ABI.hs
samalws: As far as I know, the only ways to make a function pointer constant in solidity are putting the name of a function (we couldn't turn that into its raw bytes without more context, so we couldn't construct a `AbiFunction` value) or making a bytes24 with the value you want and then using `assembly { a := b }` to skip type checking and assign the function pointer to be the bytes24 value (kind of a hack; if you want, through, we could make `parseAbiValue AbiFunctionType` parse bytes24)

d-xo: yes I think that makes sense, we use this code when constructing partially symbolic calldata buffers for symbolic execution (we parse a user provided string into some abi value and write that as a concrete region over a symbolic buffer), and I think that it would be nice to be able to do this for function pointers as well.

samalws: Okay, done

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved simplification for arithmetic expressions
 - Construction of storage counterexamples based on the model returned by the SMT solver.
 - Static binaries for macos
+- Support for function pointers in ABI
```
d-xo, line 22: This should be under an `## unreleased` heading not the 0.50.3 release

d-xo:
Thanks!

PR #208, opened 2023-02-24
cli: equivalence: pretty print cexs

d-xo:
## Description

Pretty prints the discovered counterexamples when equivalence checking. There is still room for improvement here, ideally we would also display the actual differing results, not just the inputs that produce a difference.

Example output:

```
Found 36 total pairs of endstates
Asking the SMT solver for 19 pairs
Reuse of previous queries was Useful in 0 cases
Not equivalent. The following inputs result in differing behaviours:

-----

Calldata:
  0x4dad5d1e8d40fffffe0000000000000011c19e55f9045ff8809cc098d644a0100006003885000000020000000000000011585dbcffffbf80

Transaction Context:
  CallValue: 0x0



-----

Calldata:
  0x4dad5d1e24740024800113afd83c000000000000000000000011a68df5da4707fe5fbe7e92160024800bb0980b85000000000000000020

Transaction Context:
  CallValue: 0x0



-----

Calldata:
  0x4dad5d1ed200000000497b5c01000000000000000ffffffffffffffffffc000000000080d1ffc0004000000000800000200000000000000000000000000400000000007f

Transaction Context:
  CallValue: 0x0
  ```

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/hevm-cli/hevm-cli.hs
+++ b/hevm-cli/hevm-cli.hs
@@ -315,7 +315,7 @@ equivalence cmd = do
       False -> do
         putStrLn "No discrepancies found"
         when (any isTimeout res) $ do
-          putStrLn "But timeout(s) occurred:"
+          putStrLn "But timeout(s) occurred"
           exitFailure
       True -> do
         let cexs = mapMaybe getCex res
--- a/hevm-cli/hevm-cli.hs
+++ b/hevm-cli/hevm-cli.hs
@@ -15,7 +15,6 @@ import qualified EVM.Stepper
 import EVM.SymExec
 import EVM.Debug
 import qualified EVM.Expr as Expr
-import EVM.SMT
 import EVM.Solvers
 import qualified EVM.TTY as TTY
 import EVM.Solidity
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: hevm-cli/hevm-cli.hs [hunk 1]
```diff
@@ -316,10 +315,14 @@ equivalence cmd = do
       False -> do
         putStrLn "No discrepancies found"
         when (any isTimeout res) $ do
-          putStrLn "But timeout(s) occurred"
+          putStrLn "But timeout(s) occurred:"
```
msooseth, line 318: I think you may have forgotten to print the timeouts here? Probably a `<> (filter isTimeout res)` is missing here?

hevm-cli/hevm-cli.hs
```diff
@@ -394,21 +397,23 @@ assert cmd = do
...
               unknowns
-                | null (getTimeouts cexs) = []
+                | null timeouts = []
                 | otherwise =
                    [ ""
                    , "Could not determine reachability of the following end states:"
                    , ""
-                   ] <> fmap (formatExpr) (getTimeouts cexs)
+                   ] <> fmap (formatExpr) timeouts
           T.putStrLn $ T.unlines (counterexamples <> unknowns)
```
msooseth, line 417: Such nice code!

msooseth (commented):
It's great but I didn't understand one part, with the `:` change... maybe something is missing there or maybe it's printed as part of the `assert` change, but then it'll be printed before rather than after the `:`?

hevm-cli/hevm-cli.hs
d-xo, line 318:
```suggestion
          putStrLn "But timeout(s) occurred"
```

msooseth (approved):
lgtm

Issue #207, opened 2023-02-24
`WriteWord` simplification can be wrong when offset is large


msooseth:
The PR pipeline of  https://github.com/ethereum/hevm/pull/198 broke (see logs [here](https://pipelines.actions.githubusercontent.com/serviceHosts/847bf41c-3bf4-44e1-aa43-2f4738618db6/_apis/pipelines/1/runs/1286/signedlogcontent/3?urlExpires=2023-02-24T16%3A06%3A11.4229503Z&urlSigningMethod=HMACV1&urlSignature=u7%2Btuew3yRsTUtM0dt2CXBqNuqDt2l5%2FxaxzFDpdZCc%3D )) because this randomly generation expression, unrated to the PR itself:

`EVM.Expr.simplify WriteWord (SDiv (Sub (LT (Lit 0x1a) (Lit 0x2b)) (Lit 0x59)) (Lit 0x57)) (Lit 0xba0f7483eaf0000033e5f50af97000007b670a8fe8e000017b866ff8ee9) (WriteWord (Lit 0x10) (Lit 0x194f1c1b0a5600000ccdec88b707000004743426102c000005e06b915fdc) (AbstractBuf "esc_NFIIGOFlwfrBCZWnIWOEvFBT"))`

Overflows Expr.simplify, in particular, the offset (SDiv (Sub (LT (Lit 0x1a) (Lit 0x2b)) (Lit 0x59)) (Lit 0x57)) happens to be:

```
EVM.Expr.simplify (SDiv (Sub (LT (Lit 0x1a) (Lit 0x2b)) (Lit 0x59)) (Lit 0x57))
Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```

i.e. very large. This is related to the CopySlice simplifier overflow, where a similar issue happened and has been fixed by @zoep .

asymmetric:
@msooseth not sure if you're aware, but [this url](https://pipelines.actions.githubusercontent.com/serviceHosts/847bf41c-3bf4-44e1-aa43-2f4738618db6/_apis/pipelines/1/runs/1286/signedlogcontent/3?urlExpires=2023-02-24T16%3A06%3A11.4229503Z&urlSigningMethod=HMACV1&urlSignature=u7%2Btuew3yRsTUtM0dt2CXBqNuqDt2l5%2FxaxzFDpdZCc%3D) doesn't resolve anymore.

msooseth:
[...]

PR #209, opened 2023-02-27
Bounds checking without overflows in the simplifier

zoep:
## Description
To simplify a read that happens after some write(s) on a buffers, we check wether the interval `[ir, ir + szr)` of the read is included, overlaps, or is disjoint from `[iw, iw+ szw)`. However, for large numbers `ir + szr` and `iw+szw` might overflow and wrap around, since the arithmetic is modulo 2^256. This makes bounds checking tricky, since doing it in the straightforward way is incorrect. For example saying that the intervals are disjoint if `ir +szr <= iw || iw + szw <= ir` is incorrect. This was a bug in the simplifier that caused failures in fuzing (#207 and #193). Previous fix for #193, disallowed simplification if there was an overflow (#200) but only for some cases. 

This PR rewrites the bounds checking so no overflow can happen when we check the bounds. This works even if there is wraparound in addresses, so we do not need to check for overflow.

It works as follows:

We want to check if the interval A 
```
0    ir              ir+szr      
|----|****************|--------
 ```
 is included, overlaps, or is disjoint from interval B
```
0    iw             iw+szw      
|----|****************|--------

```
with `ir+szr` and `iw+szw` potentially overflowing. 

First, shift both intervals left by `iw`. Now, interval B is 

```
0           szw           2^256
|************|-------------|
```

and interval A 

```
0    i          i+szr      
|----|************|--------
```

where `i = ir - iw`. 

The new intervals overlap if and only if the initial intervals overlap.

Now
a) if `i < szw` and `szr <= szw - i`, then interval A is included in B. The second condition can be omitted if `szr = 1` (i.e, we are reading a byte). 
b) if `i >=  szw` and `i <= 2^256 - szr`, the two intervals are disjoint. If the second condition guarantees that interval A has not wrapped around, overlapping with B. It can be omitted if `szr = 1` (i.e, we are reading a byte).
c) otherwise, the two intervals partially overlap. 
 
 Note that no overflow or underflow can happen in any of the expressions used in the bounds checking conditions. These checks will work even if there is wraparound in the bounds of the initial intervals.

## Checklist

- [X] tested locally
- [X] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -215,19 +215,18 @@ readByte i@(Lit x) (WriteByte (Lit idx) val src)
     then val
     else readByte i src
 readByte i@(Lit x) (WriteWord (Lit idx) val src)
-  = if idx <= x && x <= idx + 31
+  = if x - idx < 32
     then case val of
            (Lit _) -> indexWord (Lit $ x - idx) val
            _ -> IndexWord (Lit $ x - idx) val
     else readByte i src
 readByte i@(Lit x) (CopySlice (Lit srcOffset) (Lit dstOffset) (Lit size) src dst)
-  = if dstOffset <= x && x < (dstOffset + size)
+  = if x - dstOffset < size
     then readByte (Lit $ x - (dstOffset - srcOffset)) src
     else readByte i dst
 readByte i@(Lit x) buf@(CopySlice _ (Lit dstOffset) (Lit size) _ dst)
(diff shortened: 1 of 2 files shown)


msooseth (approved):
This is a mind-bender but I think I understood the principle and played through the weird ones in my head.

d-xo (approved):
looks great <3

PR #224, opened 2023-03-15
SMT Guided Counterexample Shrinking

d-xo:
## Description

- removes assertions on buffer lengths derived from the max read index
- implements a sound shrinking routines for buffer models

Previous to this PR we inserted constraints for each buffer that stated that their length could not be larger than the maximum index read from that buffer. This approach was unsound as branch conditions could contain arbitrary restrictions on the length of a buffer without reading from it (via BufLength).

Here we remove these constraints and use the max read index as a starting point for a counterexample shrinking routine. If our initial query returns sat, we check if any of the returned buffer models are very large (we use 1kb as the cutoff), and if they are we run the initial query again, but restrict the length to the smaller of 1kb and the max read index. If the query remains sat, we return this model, if not we double the maximum size and try again.

This should hopefully give us small counterexamples for calldata in a sound manner.

closes https://github.com/ethereum/hevm/issues/211 and https://github.com/ethereum/hevm/issues/112

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -39,13 +39,17 @@ import EVM.Types
 import EVM.Traversals
 import EVM.CSE
 import EVM.Keccak
-import EVM.Expr hiding (copySlice, writeWord, op1, op2, op3, drop)
+import EVM.Expr (writeByte, bufLengthEnv, containsNode, bufLength, minLength)
+import qualified EVM.Expr as Expr
+
 
 -- ** Encoding ** ----------------------------------------------------------------------------------
+
+
 -- variable names in SMT that we want to get values for
 data CexVars = CexVars
   { calldataV :: [Text]
-  , buffersV :: [Text]
+  , buffersV :: Map Text (Expr EWord) -- buffers and guesses at their maximum size
   , storeReads :: [(Expr EWord, Expr EWord)] -- a list of relevant store reads
   , blockContextV :: [Text]
   , txContextV :: [Text]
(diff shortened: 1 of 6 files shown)


test/test.hs
```diff
@@ -544,6 +544,19 @@ tests = testGroup "hevm"
...
+         Just c <- solcRuntime "C"
+             [i|
+             contract C {
+               function fun(uint256 a) public pure {
+                 assert(false);
+               }
+             }
+             |]
+         res <- withSolvers Z3 1 Nothing $ \s -> checkAssert s [0x1] c Nothing [] debugVeriOpts
+         print res
```
d-xo, line 557: this isn't actually testing anything right now 🤦‍♀️

src/EVM/SMT.hs
```diff
@@ -975,17 +1003,17 @@ getStore getVal sreads = do
...
+queryValue :: (Text -> IO Text) -> Expr EWord -> IO W256
+queryValue _ (Lit w) = pure w
+queryValue getVal w = do
+  let expr = toLazyText $ exprToSMT w
+  raw <- getVal expr
+  case parseCommentFreeFileMsg getValueRes (T.toStrict raw) of
+    Right (ResSpecific (valParsed :| [])) ->
+      case valParsed of
+        (_, TermSpecConstant sc) -> pure $ parseW256 sc
+        _ -> error "Internal Error: cannot parse model for storage index"
```
zoep, line 1015: We could use a more generic error here, since now the function is not only used for storage indices

zoep (approved):
Looks good! Just some small comments

PR #232, opened 2023-03-17
Change interpret to take vm arg instead of StateT

arcz:
## Description
This simplifies the code a bit and reduces peak memory on ethereum-tests from ~10G to ~9G.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


arcz:
@d-xo This is ready. While waiting for tests, I also did some cleanup to the modules I was touching.

d-xo:
Looks like a failure from our differential fuzz testing harness vs geth

d-xo (approved):
Looks good, not an issue for this pr, but in future it would make review easier if we can split these kind of formatting changes out into separate prs (or at least separate commits in the same pr)...

PR #233, opened 2023-03-20
SymExec: tidy flattenExpr

d-xo:
## Description

Tidies up `SymExec` a little by removing the list of props in the return type of `flattenExpr` and instead storing all branch conditions in the assertions field of each leaf node.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -31,7 +31,7 @@ import Data.List (foldl', sortBy)
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
 import qualified Control.Monad.State.Class as State
-import Data.Bifunctor (first, second)
+import Data.Bifunctor (second)
 import Data.Text (Text)
 import Data.Map (Map)
 import qualified Data.Map as Map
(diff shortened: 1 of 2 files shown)


src/EVM/SymExec.hs
```diff
@@ -646,19 +646,19 @@ equivalenceCheck solvers bytecodeA bytecodeB opts calldata' = do
         -- if we can statically determine that the end states differ, then we
         -- ask the solver to find us inputs that satisfy both sets of branch
         -- conditions
-        PBool True  -> Just . Set.fromList $ aProps <> bProps
+        PBool True  -> Just . Set.fromList $ extractProps aEnd <> extractProps bEnd
```
zoep, line 649: So before we would never really take into account the props inside expressions for equivalence checking? Does this make a difference semantically?

zoep (approved):
Real cool cleanup!

src/EVM/SymExec.hs
d-xo: oh true, it was probably a soundness bug. didn't even notice when I was making the change...

PR #236, opened 2023-03-21
Foundry support

d-xo:
## Description

Adds support for running ds-test tests from foundry projects:

- renames `hevm dapp-test` to `hevm test`
- adds a `--project-type` flag to `hevm test` that defaults to `Foundry` (but also accepts `DappTools`)
- removes the `--json-file` argument (doesn't really make sense in the context of a foundry project)
- removes support for reading solc json files produced via `--combined-json`
- sets up [cachix](https://www.cachix.org/) and pushes our build results there (not really needed as part of this pr in the end, but left it in since it's a small change and nice to have regardless).
- fixes the up / down arrows in the test picker window for `hevm test --debug`
- reworks some of the internal data structures to support differences in the way that source map file id's are assigned in foundry projects (due to parallel compilation) see [207c6c6d7aeb79bc2a9ab816c70135ad83729168](https://github.com/ethereum/hevm/pull/236/commits/207c6c6d7aeb79bc2a9ab816c70135ad83729168).

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


zoep (approved):
Looks great!

PR #237, opened 2023-03-23
Change BadCheatCode error to take just Word32

arcz:
## Description
It's always Just so there is no need for Maybe.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
nice :)

PR #244, opened 2023-03-30
Tidy records

d-xo:
## Description

This removes all remaining prefixes (both underscores and record specifc ones) from all records.

In order to remove the underscores I switched from the classic lens library to [optics](https://hackage.haskell.org/package/optics-0.4.2/docs/Optics.html), which has a template haskell based system for generating overloaded labels (see [here](https://hackage.haskell.org/package/optics-core-0.4.1/docs/Optics-Label.html)). This is preferable to the system provided by `generic-lens`, which was significantly slower than the template haskell lenses from the classic lens library.

optics does introduce some additional line noise (it uses `%` instead of `.` for lens composition, and field names need to be prefixed with `#` when using overloaded labels), but I think it's worth it to get rid of prefixes, and I also see some benefit to using a different symbol for lens composition since `.` is now used for record access as well. Optics also has nicer error messages than the old lens library, thanks to it's more opaque type interface.

I also hooked up the blockchain test suite to our benchmarking harness to make sure that we don't introduce any performance regressions when switching away from lenses, on my laptop I got ~30s to execute the state tests benchmarks both before and after switching over to optics.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo:
Also worth noting is that after making this change, I ran into quite a few warnings related to ambiguous record updates. It seems as though a future release of ghc will drop support for type driven record update disambiguation (see [here](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0366-no-ambiguous-field-access.rst)), I'm not sure when that will be, and there also seem to be new proposals that would allow disambiguation with similar syntax (e.g. [here](https://github.com/ghc-proposals/ghc-proposals/pull/537), so I've ignored the warnings for now. In the worst case we will have to rewrite a few record updates to be a little more verbose so it seems fairly harmless...

src/EVM/UnitTest.hs
```diff
@@ -827,31 +830,31 @@ indentLines n s =
...
+      , indentLines 2 (formatTestLogs dapp.eventMap vm.logs)
       , "\n"
       ]
     else ""
 
 -- TODO
 failOutput :: VM -> UnitTestOptions -> Text -> Text
 failOutput vm UnitTestOptions { .. } testName =
-  let ?context = DappContext { info = dapp, env = vm._env._contracts}
+  let ?context = DappContext { info = dapp, env = vm.env.contracts}
```
zoep, line 849: maybe add whitespace before closing brace

zoep (approved):
Looks good! just some minor comments

PR #245, opened 2023-03-31
EVM: prank deducts value from the correct address

d-xo:
## Description

Fixes https://github.com/ethereum/hevm/issues/240. We were not respecting prank when transfering value during a call

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


zoep (approved):
looks good!

PR #246, opened 2023-04-11
Fix off-by-one bug in EVM.Debug.srcMapCodePos

samalws-tob:
## Description
`EVM.Debug.srcMapCodePos` subtracts one from the character offset when taking characters in order to count the number of newlines. This results in one too few characters being taken. For example, if the first character in the solidity file is a newline and the offset is 1, we would expect the function to return that we are on line 2 but it would actually return that we are on line 1. This PR fixes the off-by-one bug.

In particular, this fixes a problem in Echidna where incorrect lines were being highlighted in the coverage report.
Coverage report before PR:
Screenshot 2023-04-11 at 4 26 18 PM
Coverage report after PR:
Screenshot 2023-04-11 at 4 27 20 PM

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - The `prank` cheatcode now transfers value from the correct address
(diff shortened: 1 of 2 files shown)


d-xo:
thanks!

PR #252, opened 2023-04-19
Partial execution traces

d-xo:
## Description

First of all, apologies for the very messy diff here, this pr started as an attempt to fix https://github.com/ethereum/hevm/issues/137, which (due to various cyclic module dependencies) required moving most of the types in EVM.hs into Types.hs. Once this was done, a few interesting refactors presented themselves to me and I got a little carried away without paying attention to the quality of the diff that I was producing...

Aside from the moved type definitions, this PR has a few main components:

1. Rework and unification of error types:

In our previous implementation, there were two error types, one in EVM.hs and one in Types.hs (that was used in Expr). There was quite some duplication and overlap between these types, but the reason to keep them separate was that the type from EVM.hs contained some constructors that were used for effect handling and control flow. In this PR, a new core error type (`EvmError`) has been introduced that is used in both `Expr` and `VM`, and the additional constructors have been split out into two types:

- `HandleEffect`: all side effect related constructors (e.g. PleaseAskSmt, PleaseFetchContract)
- `Unfinished`: partial executions (e.g. due to hitting max iterations)

The second type also allows us to represent partial execution traces in `Expr`, and as a result we can lift some restrictions and allow analysis even if we hit an unexpected symbolic argument. We can also display much nicer error output once we hit max iterations.

2. Rework handling of max-iterations

Our previous implementation of bounded exploration had some flaws:

- max-iterations was not respected for loops with concrete branching conditions
- any JUMPI visited more than once was considered to be a loop

This PR lifts the first restriction and also introduces a new loop detection heuristic (based on a comparison between stack elements in the current and past stacks at any visit to a given JUMPI point). This heuristic is able to fairly reliably distinguish between actual loop heads, and JUMPI instructions that are contained within an internal function. This new heuristic is the default, but can be configured via the cli to allow analysis of code where it fails.

3. Rework handling of ask-smt-iterations

In addition the default `askSmtIters` has been reduced to 1. This is unfortunate, but anything else is unsound, for example, consider the following case:

```solidity
contract C {
  function fun(uint j) external payable returns (uint) {
    require(j <= 3);
    uint count = 0;
    for (uint i = 0; i < j; i++) count++;
    return count;
  }
}
```

If we have `askSmtIters` set to 5, then we will not start asking the smt solver about the satisfiability of jump conditions at the loop head until we are at iteratin 5. At this point the branch conditions are always unsatisfiable (due to the `require` statement above), and so we will loop forever (since in this case a `sat` is required to break out of the loop).

However, askSmtIters is now only checked for branch points that are marked as loops by the loop detection heuristic, so we should be able to skip some queries that were previously dispatched for branch points inside commonly used internal functions (e.g. overflow checks).

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


src/EVM/SymExec.hs
```diff
@@ -251,68 +264,105 @@ interpret fetcher maxIter askSmtIters vm =
...
 
 maxIterationsReached :: VM -> Maybe Integer -> Maybe Bool
 maxIterationsReached _ Nothing = Nothing
 maxIterationsReached vm (Just maxIter) =
   let codelocation = getCodeLocation vm
-      iters = view (at codelocation % non 0) vm.iterations
+      (iters, _) = view (at codelocation % non (0, [])) vm.iterations
   in if num maxIter <= iters
      then Map.lookup (codelocation, iters - 1) vm.cache.path
      else Nothing
```
zoep, line 336: Maybe trowing an error here would be more appropriate since we always expect to find the code location in the map and `Nothing` means that max iterations has been reached

zoep (approved):
Looks good! Just few small comments

src/EVM/SymExec.hs
d-xo: agree that it would be a bit nicer semantically, but I think it would make some other logic a little more complex, so I would probably prefer to just leave it as is for now.

The issue is that `iterations` is only set for a given codelocation in the callback executed after the `PleaseAskSMT` query is handled, so the first time we visit a JUMPI the iterations map can be empty. We could rework the logic that updates the iterations a bit, but honestly I don't think it would really be cleaner than what we have now...

PR #253, opened 2023-04-20
Prove traces

d-xo:
## Description

This is based on top of https://github.com/ethereum/hevm/pull/252, as it requires the reorganised types there.

This adds pretty traces in counter examples from `prove_` tests to match the output from a counterexample generated from a fuzz test. We do this by adding the existing traces contained within the VM to each leaf type in `Expr End`. 

As an example the output from the `prove_transfer` test in `test/contracts/fail/dsProveFail.sol` now looks like:

```
> [FAIL] prove_transfer(uint256,address,uint256)

Failure: prove_transfer(uint256,address,uint256)

  Counterexample:

    result:   Revert: 0x4e487b710000000000000000000000000000000000000000000000000000000000000011
    calldata: prove_transfer(0,0x0000000000000000000000000000000000000000,1)

    test.sol:SolidityTest
     ├╴constructor
     ├╴setUp()
     │  └╴create <unknown contract>@0xCe71065D4017F316EC606Fe4422e11eB2c47c246 (test.sol:8)
     │     └╴← 0x1925 bytes of code
     └╴prove_transfer(uint256,address,uint256)
        ├╴call 0xCe71065D4017F316EC606Fe4422e11eB2c47c246::mint<symbolic> (test.sol:60)
        │  ├╴Transfer(<symbolic>) (lib/erc20.sol:187)
        │  └╴← 0x
        ├╴call 0xCe71065D4017F316EC606Fe4422e11eB2c47c246::balanceOf<symbolic> (test.sol:62)
        │  └╴← (<symbolic>)
        └╴call 0xCe71065D4017F316EC606Fe4422e11eB2c47c246::transfer<symbolic> (test.sol:63)
           └╴error Revert Panic(17) <source not found>
```

previously we would only show the result and the calldata (without the pretty execution trace)

fixes https://github.com/ethereum/hevm/issues/137

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Format.hs
+++ b/src/EVM/Format.hs
@@ -195,7 +195,6 @@ showTraceTree dapp vm =
   in pack $ concatMap showTree traces
 
 showTraceTree' :: DappInfo -> Expr End -> Text
-showTraceTree' _ (ITE {}) = error "Internal Error: ITE does not contain a trace"
 showTraceTree' dapp leaf =
   let forest = traceForest' leaf
       traces = fmap (fmap (unpack . showTrace dapp (traceContext leaf))) forest
--- a/src/EVM/Format.hs
+++ b/src/EVM/Format.hs
@@ -263,19 +263,6 @@ showTrace dapp vm trace =
             Nothing ->
               logn
 
-    QueryTrace q ->
-      case q of
-        PleaseFetchContract addr _ ->
-          "fetch contract " <> pack (show addr) <> pos
-        PleaseFetchSlot addr slot _ ->
-          "fetch storage slot " <> pack (show slot) <> " from " <> pack (show addr) <> pos
-        PleaseAskSMT {} ->
-          "ask smt" <> pos
-        --PleaseMakeUnique {} ->
-          --"make unique value" <> pos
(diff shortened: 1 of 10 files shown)


zoep (approved):
Looks good! Jus one small comment

src/EVM/Format.hs
```diff
@@ -189,15 +191,22 @@ formatSBinary _ = error "formatSBinary: implement me"
 showTraceTree :: DappInfo -> VM -> Text
 showTraceTree dapp vm =
   let forest = traceForest vm
-      traces = fmap (fmap (unpack . showTrace dapp vm)) forest
+      traces = fmap (fmap (unpack . showTrace dapp (vm.env.contracts))) forest
+  in pack $ concatMap showTree traces
+
+showTraceTree' :: DappInfo -> Expr End -> Text
+showTraceTree' _ (ITE {}) = error "Internal Error: ITE does not contain a trace"
```
d-xo, line 198:
```suggestion
```

PR #256, opened 2023-05-06
Add Data to AbiType

andrevidela:
## Description

For meta-programming reasons I've needed `AbiType` to conform to `Data`. Maybe it would make sense to add it to other data declarations? I'm not sure which ones so I've left them alone, let me know if we should update other instance declarations as well.

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


d-xo (approved):
Apologies for the delay. This looks good. I'm fine to just add these instances ad hoc as needed.

PR #259, opened 2023-05-18
Add basic forking cheatcodes

arcz:
## Description
Basic implementation of forking: https://book.getfoundry.sh/forge/fork-testing

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -120,16 +120,7 @@ makeVm o =
     }
   , logs = []
   , traces = Zipper.fromForest []
-  , block = Block
-    { coinbase = o.coinbase
-    , timestamp = o.timestamp
-    , number = o.number
-    , prevRandao = o.prevRandao
-    , maxCodeSize = o.maxCodeSize
-    , gaslimit = o.blockGaslimit
-    , baseFee = o.baseFee
-    , schedule = o.schedule
-    }
+  , block = block
   , state = FrameState
     { pc = 0
     , stack = mempty
@@ -145,7 +136,19 @@ makeVm o =
     , returndata = mempty
     , static = False
     }
-  , env = Env
+  , env = env
+  , cache = cache
+  , burned = 0
+  , constraints = snd o.calldata
+  , keccakEqs = mempty
+  , iterations = mempty
+  , allowFFI = o.allowFFI
+  , overrideCaller = Nothing
+  , forks = Seq.singleton (ForkState env block cache "")
+  , currentFork = 0
+  }
+  where
+  env = Env
     { sha3Crack = mempty
     , chainId = o.chainId
     , storage = o.initialStorage
@@ -153,14 +156,17 @@ makeVm o =
     , contracts = Map.fromList
       [(o.address, o.contract )]
     }
-  , cache = Cache mempty mempty mempty
-  , burned = 0
-  , constraints = snd o.calldata
-  , keccakEqs = mempty
-  , iterations = mempty
-  , allowFFI = o.allowFFI
-  , overrideCaller = Nothing
-  }
+  block = Block
+    { coinbase = o.coinbase
+    , timestamp = o.timestamp
+    , number = o.number
+    , prevRandao = o.prevRandao
+    , maxCodeSize = o.maxCodeSize
+    , gaslimit = o.blockGaslimit
+    , baseFee = o.baseFee
+    , schedule = o.schedule
+    }
+  cache = Cache mempty mempty mempty
 
 -- | Initialize empty contract with given code
 initialContract :: ContractCode -> Contract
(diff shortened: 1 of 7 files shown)


test/contracts/pass/cheatCodesFork.sol
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/// @dev This contract's state should depend on which fork we are on
+contract TestState {
+    uint256 public state;
+    function setState(uint256 _state) external {
+        state = _state;
+    }
+}
+/// @dev This contract's state should be persistent across forks, because it's the contract Echidna deploys
```
d-xo, line 25:
```suggestion
/// @dev This contract's state should be persistent across forks, because it's the contract hevm deploys
```

test/contracts/pass/cheatCodesFork.sol
```diff
@@ -0,0 +1,70 @@
...
+    uint256 persistentState;
+
+    constructor() {
+        stateContract = address(new TestState());
+        forkId1 = hevm.createFork("foo"); // If the default fork's id is 0, then this would be 1
+        forkId2 = hevm.createFork("bar"); // and this would be 2
+        persistentState = 0;
+    }
+
+    function test_ForkedState() external {
```
d-xo, line 40: according to the foundry docs, the state of `msg.sender` should also be persisted betweeen forks. Could we also test that behaviour here (I guess by sending some eth to msg.sender and checking that it is persisted between forks?).

src/EVM.hs
```diff
@@ -1542,8 +1548,57 @@ cheatActions =
...
+                              { env = vm.env, block = vm.block, cache = vm.cache }
+                            ) vm.currentFork  vm.forks
+                          , currentFork = forkId'
+                          }
+                    _ ->
+                      -- no-op: current storage and/or saved storage is symbolic
+                      -- we don't support this at the moment
+                      pure ()
+                Nothing ->
+                  pure () -- no-op: fork not found
```
d-xo, line 1600: do we wanna noop or error here? what does foundry do?

src/EVM.hs
```diff
@@ -1542,8 +1548,57 @@ cheatActions =
...
+            forceConcrete forkId "forkId must be concrete" $ \(fromIntegral -> forkId') -> do
+              saved <- Seq.lookup forkId' <$> gets (.forks)
+              case saved of
+                Just forkState -> do
+                  vm <- get
+                  case (vm.env.storage, forkState.env.storage) of
+                    (ConcreteStore store, ConcreteStore savedStore) -> do
+                      let
+                        -- the current contract is persited across forks
+                        self = vm.state.contract
```
d-xo, line 1578: I don't know exactly how foundry behaves here, but according to the docs only the state of the the test contract itself is persisted between forks, and I think we're going to persist the wrong contract if we call this from a subcontract that's deployed from the top level test contract?

src/EVM.hs
```diff
@@ -1542,8 +1548,57 @@ cheatActions =
...
+                          , block = forkState.block
+                          , forks = Seq.adjust' (\state -> (state :: ForkState)
+                              { env = vm.env, block = vm.block, cache = vm.cache }
+                            ) vm.currentFork  vm.forks
+                          , currentFork = forkId'
+                          }
+                    _ ->
+                      -- no-op: current storage and/or saved storage is symbolic
+                      -- we don't support this at the moment
+                      pure ()
```
d-xo, line 1598: What exactly is the difficultty in supporting symbolic storage here? I think if we can't do it easily it's probably better to error out?

d-xo (commented):
CI is only failing because of some unrelated windows build error...

samalws-tob:
@arcz @d-xo  I did a bit of work on this PR, result is [here](https://github.com/samalws-tob/hevm/tree/forking). Changes:
- Brought it up to date to current the main branch
- Renamed the test function (previously it wasn't actually running the test, and was always passing regardless of what was in the function)
- Made it error instead of no-op when the fork isn't found in `selectFork`
- Made `msg.sender` get persisted when switching forks
- Added a test for whether `msg.sender.balance` is persisted

Right now the test fails, because `msg.sender` being required to be concrete screws things up. A test that only says `{ assert(msg.sender.balance == msg.sender.balance); }` would also fail at the moment. **Any ideas for how to get around this?**

d-xo:
Thanks for the work on this pr @samalws-tob, looks super nice.

> Right now the test fails, because `msg.sender` being required to be concrete screws things up. A test that only says `{ assert(msg.sender.balance == msg.sender.balance); }` would also fail at the moment. **Any ideas for how to get around this?**

I'm not sure I fully understand the question. `msg.sender` is currently fully abstract, but we do have some restrictions around accessing balances of fully abstract addresses (this is due to potential unsoundness due to aliasing between symbolic addresses), which I guess is the issue that you were running into? The long term fix here would probably be to restructure balances to be sound even if we have aliased symbolic addresses (probably by having one global symbolic mapping storing balances, instead of a concrete haskell mapping as we do at the moment), this is a larger job, and out of scope for this pr. I think in the short term you can maybe rework the tests so that you only look up balances of contracts that have actually been deployed (e.g. would using `address(this).balance` instead of `msg.sender.balance` work?).

samalws-tob:
Using `address(this).balance` instead of `msg.sender.balance` wouldn't work, since we want to test that the caller's state is preserved in addition to the test contract's state. One possible workaround solution would be having two contracts, `TestContractDeployer` and `TestContract`, where `TestContract` contains the current test code while `TestContractDeployer` simply deploys a `TestContract` and calls its test function; this way, from the point of view of the test code (which lives in `TestContract`), `msg.sender` will always be concrete, since it is the address of the `TestContractDeployer`. Does this seem reasonable? I can make a demo if you want.

PR #274, opened 2023-06-18
Range assertion for environment variables

zoep:
## Description

This PR adds assertions for the range of environment variables that are less than word size (256bits). This was necessary for equivalence proofs with Act code. 

Also, it adds some code to print SMT queries generated by the equivalence checker when debug flag is on.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [X] updated the changelog


d-xo (approved):
looks good!

PR #275, opened 2023-06-18
Preprocess SMT2 sexpressions

zoep:
## Description

This PR adds a preprocessing step for SMT2 scripts that strips comments and puts each sexpression in its own line.  Having sepxrs that span across multiple lines triggers a bug in CVC5. 

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/src/EVM/Solvers.hs
+++ b/src/EVM/Solvers.hs
@@ -306,10 +306,13 @@ readSExpr h = go 0 0 []
          then pure $ line : prev
          else go (ls + ls') (rs + rs') (line : prev)
 
-
--- From a list of lines, take each separate SExpression and put it in its own list
+-- From a list of lines, take each separate SExpression and put it in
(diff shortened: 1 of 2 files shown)


src/EVM/Solvers.hs
```diff
@@ -231,7 +228,7 @@ solverArgs solver timeout = case solver of
     [ "-in" ]
   CVC5 ->
     [ "--lang=smt"
-    , "--no-interactive"
+    -- , "--no-interactive"
```
d-xo, line 231: can we remove this instead of commenting it?

d-xo (approved):
looks good! just a tiny nit.

src/EVM/Solvers.hs
zoep: done!

PR #287, opened 2023-06-28
Move slow tests to benchmarks

msooseth:
## Description

This removes some slow-running tests and we are adding them to https://github.com/eth-sc-comp/benchmarks instead. I also adjusted some of the property-based tests to run only 50 instead of the default 100.

Before: 1850.08 s
After:628.73 s

Related to: https://github.com/ethereum/hevm/issues/282

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


d-xo (approved):
looks good!

PR #288, opened 2023-06-28
Better error printing

msooseth:
## Description

Errors were not consistently marked with "Internal error" -- and sometimes it was spelled differently as well. This tries to unify all error messages. Some of the errors in `hevm-cli.hs` were changed to have a consistent `Error:` printed in front, when it's likely not an "internal" error.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Solidity.hs
+++ b/src/EVM/Solidity.hs
@@ -120,12 +120,12 @@ instance Read SlotType where
   readsPrec _ t@('m':'a':'p':'p':'i':'n':'g':'(':s) =
     let (lhs,rhs) = case T.splitOn " => " (pack s) of
           (l:r) -> (l,r)
-          _ -> error $ internalError "could not parse storage item: " <> t
+          _ -> internalError $ "could not parse storage item: " <> t
         first = fromJust $ parseTypeName mempty lhs
         target = fromJust $ parseTypeName mempty (T.replace ")" "" (last rhs))
         rest = fmap (fromJust . (parseTypeName mempty . (T.replace "mapping(" ""))) (take (length rhs - 1) rhs)
     in [(StorageMapping (first NonEmpty.:| rest) target, "")]
-  readsPrec _ s = [(StorageValue $ fromMaybe (error $ internalError "could not parse storage item: " <> s) (parseTypeName mempty (pack s)),"")]
+  readsPrec _ s = [(StorageValue $ fromMaybe (internalError $ "could not parse storage item: " <> s) (parseTypeName mempty (pack s)),"")]
 
 data SolcContract = SolcContract
   { runtimeCodehash  :: W256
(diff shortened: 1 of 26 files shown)


src/EVM/Types.hs
```diff
@@ -1233,6 +1233,8 @@ abiKeccak =
 
 -- Utils -------------------------------------------------------------------------------------------
 
+internalError:: [Char] -> [Char]
+internalError a = "internalError " ++ a
```
zoep, line 1237: We could also make this function throw the error
``` error $ "Internal error:" <> a```
to avoid writing `error $ internalError` everywhere, since this is the only way it is used.

Comment (zoep) [inline src/EVM/Types.hs:1237]: Maybe could stylise `internalError` as `Internal error: ` for better readability since it is an output message.

Comment (msooseth) [reply src/EVM/Types.hs:1237]: Oh that's a good point :D

Comment (msooseth) [reply src/EVM/Types.hs:1237]: Good point!

File: src/EVM/Solidity.hs [hunk 2]
```diff
@@ -605,8 +605,8 @@ toCode :: Text -> ByteString
 toCode t = case BS16.decodeBase16 (encodeUtf8 t) of
   Right d -> d
   Left e -> if containsLinkerHole t
-            then error "unlinked libraries detected in bytecode"
-            else error $ T.unpack e
+            then error $ internalError "unlinked libraries detected in bytecode"
+            else error $ internalError $ T.unpack e
```
zoep, line 609: Is this always an internal error?

msooseth: Actually, no, you are right :S Thanks! Sorry.  I am changing it to simply "Error:"

src/EVM/Types.hs
msooseth: Fixed! Thank you!

d-xo:
I'm wondering what this is going to do to our stack traces when we fire an error? I think it will always point to the definition of `internalError` now?

msooseth:
> I'm wondering what this is going to do to our stack traces when we fire an error? I think it will always point to the definition of `internalError` now?

Actually, good point, didn't think of that. Let me test!

msooseth:
Hmmm, we can do this.

Change the function to:

```
internalError:: HasCallStack => [Char] -> a
internalError m = error $ "Internal error: " ++ m ++ " -- " ++ (prettyCallStack callStack)
```

Then our error becomes (changed a function to error out to test):
```
  contract-quickcheck-run
    random-contract-concrete-call:                                FAIL (0.07s)
      *** Failed! (after 1 test):
      Exception:
        Internal error: whatever -- CallStack (from HasCallStack):
          internalError, called at src/EVM.hs:748:26 in hevm-0.51.1-inplace:EVM
          error, called at src/EVM/Types.hs:1243:19 in hevm-0.51.1-inplace:EVM.Types
          internalError, called at src/EVM.hs:748:26 in hevm-0.51.1-inplace:EVM
      OpContract [...]
      Use --quickcheck-replay=17565 to reproduce.
      Use -p '/random-contract-concrete-call/' to rerun this test only.
```

So this prints a bit of a call stack, which could actually be helpful anyway, and it resolves the problem of only printing error from `Types.hs` (line `1243`). It now prints the call stack too, which in this case correctly points to `EVM.hs`'s line `748`.

d-xo (approved):
nice, looks great :)

zoep (approved):
Looks great!

PR #293, opened 2023-06-30
Bring back combined JSON loading

arcz:
## Description
Combined JSON is still used by some projects (like Echidna) and the recent removal of support made it hard to migrate to newer hevm versions. While the ultimate goal is to get rid of it, it is good to have it still around.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/EVM/Test/Utils.hs
+++ b/test/EVM/Test/Utils.hs
@@ -75,6 +75,7 @@ compile DappTools root src = do
   createDirectory (root <> "/out")
   T.writeFile (root <> "/out/dapp.sol.json") json
   readBuildOutput root DappTools
+compile CombinedJSON _root _src = error "unsupported"
 compile Foundry root src = do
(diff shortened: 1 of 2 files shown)


test/EVM/Test/Utils.hs
```diff
@@ -75,6 +75,7 @@ compile DappTools root src = do
   createDirectory (root <> "/out")
   T.writeFile (root <> "/out/dapp.sol.json") json
   readBuildOutput root DappTools
+compile CombinedJSON _root _src = error "unsupported"
```
arcz, line 78: @d-xo I wasn't sure what to do here

d-xo: this is fine imo

d-xo (approved):
looks good!

PR #295, opened 2023-07-04
Adding debug flags and tools

msooseth:
## Description
Allows to give `-f debug` flag to `cabal` so it builds with extensive debug information and eventlog enabled. We now also include `eventlog2html` as part of our cabal. This should aid in debugging in the future.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -61,6 +61,13 @@ common shared
     ghc-options: -Werror
   if flag(devel)
     ghc-options: -j
+  if flag(debug)
+    ghc-options:
+      -j
+      -g3
+      -fdistinct-constructor-tables
+      -finfo-table-map
+      -eventlog
   ghc-options:
     -Wall
     -Wno-unticked-promoted-constructors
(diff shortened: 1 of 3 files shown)


hevm.cabal
```diff
@@ -241,7 +253,8 @@ executable hevm
     vty,
     stm,
     spawn,
-    optics-core
+    optics-core,
+    eventlog2html
```
d-xo, line 257: Does this actually need to be in the cabal file? Can it go in the nix shell instead? I would prefer that it's not listed as a direct dep of the exe unless actually required...

msooseth: Yeah, good point. Moved to nix :)

msooseth:
Annoyingly, some random bug got exposed via fuzzing :) I'm tracking it down.

d-xo (approved):
lgtm

arcz (approved):

PR #296, opened 2023-07-06
Removing unused sha3Crack

msooseth:
## Description
Removes the old `sha3Crack` which is collected but never used. It has been replaced with the actually used `keccakEqs`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

PR #313, opened 2023-07-12
Revert shacrack remove

msooseth:
## Description
Oooops, my understanding of this code was wrong, and it was actually useful. SHA256 is not the same as Keccak(!). So we need a reverse lookup for both. This reverts PR #296 including the changelog.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
This revert is supposed to fix this bug:

```
a = SHA256 (WriteWord (Lit 0x1) (Lit 0x0) (ConcreteBuf "1"))
b = Expr.simplify a
checkEquiv a b
```

PR #318, opened 2023-07-18
Mutable memory

arcz:
## Description
Closes https://github.com/ethereum/hevm/issues/292.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -23,6 +23,7 @@ import EVM.Types
 import EVM.Sign qualified
 import EVM.Concrete qualified as Concrete
 
+import Control.Monad.ST (ST)
 import Control.Monad.State.Strict hiding (state)
 import Data.Bits (FiniteBits, countLeadingZeros, finiteBitSize)
 import Data.ByteArray qualified as BA
@@ -48,6 +49,8 @@ import Data.Typeable
 import Data.Vector qualified as V
 import Data.Vector.Storable qualified as SV
 import Data.Vector.Storable.Mutable qualified as SV
+import Data.Vector.Unboxed qualified as VUnboxed
+import Data.Vector.Unboxed.Mutable qualified as VUnboxed.Mutable
 import Data.Word (Word8, Word32, Word64)
 import Witch (into, tryFrom, unsafeInto)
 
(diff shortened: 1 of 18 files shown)


msooseth:
BTW, the previous failure, `:main --quickcheck-replay=116473 -p "buffer-simplification"` was reproducible only on buggy Z3. I am thinking about going back to old Z3...

arcz:
yeah, the tests now fail quite often and I believe this is all caused by the new z3

src/EVM/Types.hs
```diff
@@ -680,6 +683,14 @@ data FrameState = FrameState
   }
   deriving (Show, Generic)
 
+data Memory s
+  = ConcreteMemory !(STVector s Word8)
+  | SymbolicMemory !(Expr Buf)
+
+instance Show (Memory s) where
+  show (ConcreteMemory _) = "<can't show mutable memory>"
```
d-xo, line 691: why not?

arcz: because we need to be in ST or IO monad to do that

src/EVM.hs
```diff
@@ -543,7 +548,10 @@ exec1 = do
               burn g_verylow $
                 accessMemoryWord x $ do
                   next
-                  assign (#state % #stack) (readWord (Lit x) mem : xs)
+                  xd <- readMemory (Lit x) (Lit 32)
+                  let w = Expr.readWordFromBytes (Lit 0) xd
+                  assign (#state % #stack) (w : xs)
+                  -- assign (#state % #stack) (readWord (Lit x) mem : xs)
```
d-xo, line 554:
```suggestion
```

src/EVM.hs
```diff
@@ -1932,33 +1967,76 @@ accessMemoryRange f l continue =
...
+            expandMemory (fromIntegral (memOffset' + size'))
+            gets (.state.memory) >>= \case
+              ConcreteMemory mem' -> do
+                let src =
+                      if srcOffset' >= fromIntegral (BS.length b) then
+                        BS.replicate (fromIntegral size') 0
+                      else
+                        BS.take (fromIntegral size') $
+                        padRight (fromIntegral size') $
+                        BS.drop (fromIntegral srcOffset') b
```
d-xo, line 1991: can we use `tryFrom` / `into` here?

src/EVM.hs
```diff
@@ -1932,33 +1967,76 @@ accessMemoryRange f l continue =
...
+            let pastEnd :: Int = (fromIntegral offset + fromIntegral size) - fromIntegral memSize
+            let fromMemSize = if pastEnd > 0 then fromIntegral size - pastEnd else fromIntegral size
 
-readMemory :: Expr EWord -> Expr EWord -> VM -> Expr Buf
-readMemory offset size vm = copySlice offset (Lit 0) size vm.state.memory mempty
+            a <- VUnboxed.freeze $ VUnboxed.Mutable.slice (fromIntegral offset) fromMemSize mem
+            let dataFromMem = BS.pack $ VUnboxed.toList a
+            pure $ ConcreteBuf $ dataFromMem <> BS.replicate (fromIntegral pastEnd) 0
+        _ ->
+          error "readMemory transition"
```
d-xo, line 2032: I think we can freeze memory and call `copySlice` on it here?

d-xo (commented):
Aside from the unfortunate proliferation of type parameters this is surprisingly clean. I had expected integrating St to be more involved tbh. Really nice work <3. 

I wonder if it's possible to add some type aliases or newtypes or something to tidy up some of the clutter?

d-xo:
What's left to do here? I have a bunch of larger changes I wanna make and I'd like to get this in first to avoid more merge work for you @arcz

d-xo (approved):
I think this looks great. No comments. IMO we can merge as soon as CI passes here.

PR #319, opened 2023-07-18
flake.nix: disable foundry if on darwin

siraben:
## Description
This is so that `nix develop` can work on darwin. Note that currently some upstream problems with ormolu still need to be resolved, but that is not related.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -32,6 +32,7 @@
           z3-latest
           cvc5
           git
+        ] ++ lib.optional (!pkgs.stdenv.isDarwin) [
           foundry.defaultPackage.${system}
         ];
 


d-xo:
Thanks! We should also update the `Dapp-Tests` test group in `test.hs` to use `DappTools` instead of `Foundry` if we're on macos (check `runSolidityTest`  in `test/EVM/Test/Utils.hs`)

d-xo:
closing in favour of: https://github.com/ethereum/hevm/pull/347

PR #320, opened 2023-07-18
Generalize abstractVM so that it can be used for contract creation and add keccak computation pass

zoep:
## Description

This PR adds two functionalities useful for Act:

1) Generalises the `abstractVm` function so that it can be used for contract creation. 
2) Adds a pass that simplifies arguments of Keccak calls, checking it a call can be computed to a concrete call. If so it adds an equation of the form `Peq (Keccak bs) concrete_hash` to the asserted propositions. This is useful for Act because the Act translation will not compute keccak by default so it is missing some information compared to the EVM code.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Keccak.hs
+++ b/src/EVM/Keccak.hs
@@ -4,7 +4,7 @@
     Module: EVM.Keccak
     Description: Expr passes to determine Keccak assumptions
 -}
-module EVM.Keccak (keccakAssumptions) where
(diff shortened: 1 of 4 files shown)


d-xo (approved):
looks good :)

PR #321, opened 2023-07-18
Symbolic contract address

d-xo:
## Description

Adds support for symbolic contract addresses to hevm. It's still a little rough, and I wanna do a bit of clean up and add tests, but for now the existing tests are finally passing so I'm opening this as a draft. Sorry for the huge diff, this required reworking how we model storage, which consequently meant changing basically every file :/

The main change is that we now have per contract storage again, where storage is keyed on a new address type (`Expr EAddr`), which can be either concrete or abstract. The plan is to constrain symbolic addresses in smt in such a way that we can rely on the fact that if two symbolic addresses are disctinct syntactically then they will also be distinct semantically. This means we can avoid needing to query the solver when doing contract creation or calling into a contract at a symbolic address. The downside is that we will miss some interesting cases where e.g. the caller and origin are aliased. These constraints still needed to be added to the SMT encoding.

Contract balances have also been made symbolic as part of this PR. Which means we can support calls to symbolic addresses with symbolic values. Looking back this probably should have been done separately, but I didn't really notice until it was too late...

I removed the `EmptyStore` constructor for `Expr Storage` and changed the types so that it now represents a single (W256 -> W256) mapping. `AbstractStore` now holds an `Expr EAddr` so that we know which address to use when doing SMT encoding. The SMT encoding has (I hope) been optimised somewhat as we now no longer have one large nested array, but instead have one array per contract. Cex construction now has an additional phase `expandCex` where we add some context (any concrete storage in the pre state) that has been simplified away by the time we get to our final `Expr`. 

While debugging I also ended up changing a little bit how we interface with the SMT solvers, and we now check for errors after sending each line. I think this should be a little more robust than our previous approach.

Instead of returning a global store in the `Success` node, we now return a map from `Expr EAddr` to `Expr EContract`, where `Expr EContract` is a simplified version of the full `Contract` type we use in the `VM` (with extraneous metadata fields removed). This change required changing the equivalence checking code so that it now compares storage, nonce, code and balance for each contract in the state (as opposed to storage only in the previous version). This constitutes a soundness fix. Test cases that excersice this new behaviour still need to be added. 

I also changed the way we model nonces. They are now `Maybe W64`, where `Nothing` can be read as "symbolic". I chose this representation, since I don't think we will be able to get useful models for nonces back from the SMT solver since they are only used as inputs to address generation, which involves hashing.

I also had to make some changes to the cli options  (we can no longer pass in a starting state), this should be possible to add in again at some point, I just wasn't sure what the best syntax would be for the cli, and the pr was already getting out of control.

Still to do:

- [ ] general tidyup
- [x] switch to text labels for `SymAddr` (so we can have meaningful names in the output)
- [x] test cases for symbolic prank
- [x] general symbolic address test cases
- [x] test cases for the new equivalence checker

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


msooseth (commented):
Sorry, this is the only thing I could find :S

msooseth:
Note: test failures are likely due to `baseStore_symaddr_XXXX` being wrongly formatted due to strange address names being emitted by `formatEAddr`, perhaps not escaping some special characters of SMTLIB2. Such as `SymAddr "\49857~AAcA\v$\GS5\RS\176050{1NdT\\\164848n`"))),Lit 0x7)` and `while sending the following line: (declare-const baseStore_symaddr_"entrypoint" Storage)`  latter likely due to the `"`, former perhaps also due to `"` and maybe also `$` and a number of `\`.

d-xo:
> Note: test failures are likely due to `baseStore_symaddr_XXXX` being wrongly formatted due to strange address names being emitted by `formatEAddr`, perhaps not escaping some special characters of SMTLIB2. Such as `SymAddr "\49857~AAcA\v$\GS5\RS\176050{1NdT\\\164848n"))),Lit 0x7)` and `[...] while sending the following line: (declare-const baseStore_symaddr_"entrypoint" Storage)` latter likely due to the `"`, former perhaps also due to `"` and maybe also `$` and a number of `\`.

 I think it was due to the `Arbitrary` instance for `Expr EAddr` in the end

d-xo:
I've thought a bit more about aliasing rules for symbolic addresses and I think we need to do the following to be (mostly) precise and sound:

- an address passed in as calldata can alias to anything
- tx.origin cannot alias to any deployed contract (it must be an eoa)
- coinbase can alias to anything
- caller can alias to anything (this is a bit of an overaproximation, since some contracts are not able to call into themselves, but determining which ones those are seems tricky)
- deployed contracts cannot alias to each other

In general, we need to be careful to never put an address that can alias with a deployed contract as a key in the `contracts` mapping. This probably means we need to segregate the state somehow, not entirely sure what's best. Since this PR is already huge, my plan is to merge this with an overaproximiated aliasing ruleset (i.e. anything can alias to anything), and close any obvious soundness holes with this approach (probably just disallowing symbolic addresses in a bunch of places), and then implement these aliasing rules in a separate PR.

src/EVM/Types.hs
```diff
@@ -615,14 +643,27 @@ data VM = VM
...
-  , allowFFI       :: Bool
-  , overrideCaller :: Maybe Addr
+  , config         :: RuntimeConfig
   }
   deriving (Show, Generic)
 
 -- | Alias for the type of e.g. @exec1@.
 type EVM a = State VM a
 
+-- | The VM base state (i.e. should new contracts be created with abstract balance / storge?)
```
zoep, line 653:
```suggestion
-- | The VM base state (i.e. should new contracts be created with abstract balance / storage?)
```

src/EVM/Expr.hs
```diff
@@ -651,8 +641,8 @@ simplify e = if (mapExpr go e == e)
     go (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x0) _ dst) = dst
```
zoep, line 642: Maybe we could also add a simplification of `(WAddr (LitAddr val))` to `(Lit $ into val)`?

d-xo:
After discussing a bit with @zoep we think that we are sound even with full address aliasing if the only way to add a contract to the storage mapping at runtime is via a call to create. This ensures that the addresses (both symbolic and concrete) in the storage mapping can never be aliased, and lets us assume that syntactic equality for symbolic addresses implies semantic equality (e.g. `SymAddr "a"` and `SymAddr "b"` cannot alias if both are present in the storage map).

We will still have some false positives due to aliasing, but that seems fine for now, and can be dealth with in a future pr.

d-xo:
IMO this is ready to go. I'm gonna take some time again on monday to go over it once more and make any small tidyups that I think are needed, but I think the core functionality is working and ready to merge.

src/EVM/Expr.hs
d-xo: done.

hevm-cli/hevm-cli.hs
```diff
@@ -65,7 +63,7 @@ data Command w
       , origin        :: w ::: Maybe Addr       <?> "Tx: origin"
       , coinbase      :: w ::: Maybe Addr       <?> "Block: coinbase"
       , value         :: w ::: Maybe W256       <?> "Tx: Eth amount"
-      , nonce         :: w ::: Maybe W256       <?> "Nonce of origin"
+      , nonce         :: w ::: Maybe Word64     <?> "Nonce of origin"
```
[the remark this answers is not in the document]
d-xo: I think I'm gonna leave it as for now to avoid blowing up this diff any further.

d-xo:
Not sure why this is failing ci :( . Tests are passing locally. My best guess is we're getting killed oom on the ethereum-tests, which would imply that I introduced a perf regression somewhere...

d-xo:
I got ~10mb max resident memory on this branch vs ~8mb on `main` when running: `cabal run -O2 ethereum-tests`. Really have no idea why ci is getting killed on ubuntu..... :(

d-xo:
Running `cabal run --enable-profiling ethereum-tests -- +RTS -s -RTS` on `main` and `symbolic-contract-address` also gives almost identical numbers:

### `symbolic-contract-address`

```
1,095,639,762,176 bytes allocated in the heap
  79,871,445,288 bytes copied during GC
   4,842,802,160 bytes maximum residency (77 sample(s))
     197,633,040 bytes maximum slop
           11818 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     242373 colls,     0 par   23.097s  23.157s     0.0001s    0.0050s
  Gen  1        77 colls,     0 par   12.701s  12.702s     0.1650s    1.0597s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time  342.380s  (342.183s elapsed)
  GC      time   35.798s  ( 35.859s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time  378.178s  (378.042s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    3,200,071,910 bytes per MUT second

  Productivity  90.5% of total user, 90.5% of total elapsed
```

###  `main`

```
1,092,839,144,160 bytes allocated in the heap
  69,478,387,856 bytes copied during GC
   4,645,248,976 bytes maximum residency (77 sample(s))
     216,420,400 bytes maximum slop
           10378 MiB total memory in use (1142 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     241512 colls,     0 par   18.782s  18.837s     0.0001s    0.0062s
  Gen  1        77 colls,     0 par   11.509s  11.510s     0.1495s    0.7225s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time  371.270s  (370.773s elapsed)
  GC      time   30.291s  ( 30.347s elapsed)
  RP      time    0.000s  (  0.000s elapsed)
  PROF    time    0.000s  (  0.000s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time  401.561s  (401.120s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    2,943,517,790 bytes per MUT second

  Productivity  92.5% of total user, 92.4% of total elapsed
```

src/EVM/SymExec.hs
```diff
@@ -726,7 +718,48 @@ equivalenceCheck' solvers branchesA branchesB opts = do
...
+      (_, Partial {}) -> PBool False
+      (ITE _ _ _, _) -> internalError "Expressions must be flattened"
+      (_, ITE _ _ _) -> internalError "Expressions must be flattened"
+      (a, b) -> if a == b
+                then PBool False
+                else PBool True
+
+    statesDiffer :: Map (Expr EAddr) (Expr EContract) -> Map (Expr EAddr) (Expr EContract) -> Prop
+    statesDiffer aState bState
+      = if Set.fromList (Map.keys aState) /= Set.fromList (Map.keys bState)
```
zoep, line 744: Just a note that when we do equivalence checking for Act we will want to check for equivalence up to a renaming of addresses. We can integrate this into this equivalence checker or write an other one for Act.

src/EVM/SymExec.hs
```diff
@@ -911,10 +953,24 @@ subBufs model b = Map.foldlWithKey subBuf b model
                       else a
           e -> e
 
-subStore :: Map W256 (Map W256 W256) -> Expr a -> Expr a
-subStore model b = mapExpr go b
+subStores :: Map (Expr EAddr) (Map W256 W256) -> Expr a -> Expr a
```
msooseth, line 956: Maybe this function could have a bit of a comment about what it's doing. Its function signature does not indicate it to be something trivial :S

msooseth (approved):
Nice!

Also, love the tests

PR #322, opened 2023-07-18
revert to default z3 (latest is buggy)

d-xo:
## Description

turns out latest has a bug that's giving us a lot of false positives and messing up ci...

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


arcz (approved):

arcz:
I wonder if this is the reason why the older version is the default 🤔

PR #323, opened 2023-07-19
Optimize CI

d-xo:
## Description

- switches from cachix to magic-nix-cache (local cache should be a bit faster than cachix cache)
- runs cabal check and rpc-tests in parallel with the main `nix build` (duplicates a bit of work but we finish faster in the end)
- runs rpc-tests on linux only (running on macos was taking 20mins)
- uses the github cache for hackage deps on windows
- runs ethereum-tests as part of `nix build` (avoids shell setup time)
- disbles library profiling in ci (avoids double builds)
- disabled haddock generation in ci (not needed to build & test executables)
- runs ci on push to main and pull request only (avoids duplicated builds on prs)

Taken together this seems to get us down to ~10mins for ci execution. Ideally we would also cache all our nix deps in the github actions cache (instead of pulling from cache.nixos.org) but there doesn't seem to be a really good way to do that atm (magic-nix-cache only caches stuff that gets built during the actions run).

Note that this change also means that we no longer run ci on every push to every branch. If you want CI to run for a branch, you need to open a draft. This is a bit unfortunate, but it seems to be a limitation of the github actions workflow triggers. I decided that cutting ci time was more important than having it run on every commit...

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ jobs:
       - uses: haskell/actions/setup@v2
         id: cabal
         with:
-          ghc-version: '9.4.5'
+          ghc-version: '9.2.8'
       - name: run cabal check
         run: cabal check --verbose=3
       - name: build and install dependencies
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,8 @@ name: "Build"
 on:
   pull_request:
   push:
+    branches:
+      main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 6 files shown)


d-xo:
hm, it seems like we're not getting cache hits on macos and windows....

msooseth (approved):
lgtm. I am not extra-good at this yml build stuff, but I have a few pipelines working in GH. None of them this elaborate, though. The nix and the cache especially is non-trivial but it looks OK actually.

PR #324, opened 2023-07-20
flake.nix: install cabal-install from head

d-xo:
## Description

This is pretty ugly (and will probably make updating to a newer nixpkgs annoying), but it does give us access to the multiple components features that were added to cabal in https://github.com/haskell/cabal/pull/8726. Which is imo a large enough workflow upgrade that it's worth this temporary hack until we get an official release. 

See https://well-typed.com/blog/2023/03/cabal-multi-unit/ for docs.

To spawn a repl session for the tests where `:r` will reload changes made in `lib:hevm`, and run a single test:

```
cabal repl --enable-multi-repl test lib:hevm lib:test-utils
ghci> import Main
ghci> runSubSet "<TESTNAME>"
```

One unfortunate downside is that `Paths.getDataFileName` seems to be broken in a multi-component repl, so the tests that trace vs geth and the `Dapp-Tests` test group are broken. They still work in the normal single component repl.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "cabal-head": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1689711523,
+        "narHash": "sha256-lmECGcDNg+rgGAYTaK6v3xMiblrA+W0LOBDrWvaOr4Y=",
+        "owner": "haskell",
+        "repo": "cabal",
+        "rev": "baa767a90dd8c0d3bafd82b48ff8e83b779f238a",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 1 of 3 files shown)


msooseth (approved):
lgtm. I see what you mean about updating, seeing the nix file changes, but I think we can go back to released cabal once the multi-component support is released. So I think it's a good workaround, and it's definitely rather slow as-is.

d-xo:
> Do you think we can add `multi-repl: True` in the `hevm.cabal` file? I think it'd make sense to make it default, no?

oh good idea, didn't realise we could do that...

PR #325, opened 2023-07-20
SymExec: fix and test evalProp

d-xo:
## Description

noticed that `evalProp` was not really doing what we wanted it to, and mostly failed to actually evaluate props down to their simplest form. 

This pr fixes that and adds actual tests for it.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
NICE. I love it!

PR #335, opened 2023-07-28
UnitTest: reverts do not fail symbolic tests

d-xo:
## Description

This change makes it so that symbolic tests will no longer report any reachable revert as a failure, and instead treat only the following as failures:

1. A failed user defined assert (e.g `assert(0 == 1)`)
2. A failed ds-test assertion violation (e.g. `assertEq`)

This makes tests execute faster, jmakes them easier to write, and aligns with the way halmos and kevm work.

I also tidied a little bit the solidity test files and the code that executes them.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/contracts/pass/dsProvePass.sol
+++ b/test/contracts/pass/dsProvePass.sol
@@ -48,29 +48,22 @@ contract SolidityTest is DSTest {
         assertEq(b, c.a());
     }
 
-    function proveFail_revertSmoke() public {
+    function prove_no_fail_require() public {
         require(false);
+    }
+
+    function proveFail_userAssertSmoke() public {
+        assert(false);
     }
 
     function proveFail_assertSmoke() public {
         assertTrue(false);
     }
 
-    // Takes too long, disabling here, moving to benchmarks
-    // function prove_transfer(uint supply, address usr, uint amt) public {
-    //     if (amt > supply) return; // no underflow
-    //
-    //     token.mint(address(this), supply);
-    //
-    //     uint prebal = token.balanceOf(usr);
-    //     token.transfer(usr, amt);
-    //     uint postbal = token.balanceOf(usr);
-    //
-    //     uint expected = usr == address(this)
-    //                     ? 0    // self transfer is a noop
-    //                     : amt; // otherwise `amt` has been transfered to `usr`
-    //     assertEq(expected, postbal - prebal);
-    // }
+    // requires don't fail passing tests, but are not treated as successes in proveFail tests.
+    function proveFail_require() public {
+        require(false);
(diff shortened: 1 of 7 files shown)


msooseth:
Failed:

```
2023-07-28T14:02:10.4766299Z hevm>     Prove-Tests-Fail:                                             Running 2 tests for src/unit-tests.t.sol:SolidityTest
[...]
2023-07-28T14:02:27.2898298Z hevm> FAIL (17.52s)
2023-07-28T14:02:27.2900472Z hevm>       test/test.hs:663:
2023-07-28T14:02:27.2902398Z hevm>       test result
2023-07-28T14:02:27.2904380Z hevm>       expected: False
2023-07-28T14:02:27.2906159Z hevm>        but got: True
2023-07-28T14:02:27.2908349Z hevm>       Use -p '/Prove-Tests-Fail/' to rerun this test only.
```

test/contracts/pass/dsProvePass.sol
```diff
@@ -48,29 +48,22 @@ contract SolidityTest is DSTest {
...
-    //     token.transfer(usr, amt);
-    //     uint postbal = token.balanceOf(usr);
-    //
-    //     uint expected = usr == address(this)
-    //                     ? 0    // self transfer is a noop
-    //                     : amt; // otherwise `amt` has been transfered to `usr`
-    //     assertEq(expected, postbal - prebal);
-    // }
+    // requires don't fail passing tests, but are not treated as successes in proveFail tests.
+    function proveFail_require() public {
```
msooseth, line 64: I am confused, sorry. This is supposed to not fail, because it's a `require`. However, this says `proveFail`, so it's expecting a failure. So this test fails. However, this is in the file `dsProvePass.sol`, so this file is supposed to pass, and it can only pass if this failure is proved, but this failure will not happen, due to `require` so this will fail the test. Yet it passes. I'm confused, sorry. Can you explain what's going on here?

d-xo: this test is just wrong. It should be in `dsProveFail`.

The semantics should IMO be:

`prove_`: only reachable user defined assertion violations or ds-test assertion violations should trigger a failure
`proveFail_`: all branches should either revert or trigger a ds-test assertion violation

d-xo: wait no, this test should actually be here. lol. yeah this is bit of a mindfuck, think we can probably just remove the `proveFail` stuff in the long term.

msooseth (approved):
lgtm

PR #336, opened 2023-07-28
SymExec: default symbolic values when displaying calldata

d-xo:
## Description

Sometimes we get back a model that does not mention all variables that are present in the input calldata buffer. This can happen if those variables are not relevant to the branch that we are producing a model for.

In this case we can simply set the values that remain symbolic afters subsituting in our cex to some default value. We will still have issues here if we have lost some constraints over those variables during simplification, but it's better than what we're doing now (i.e. just printing "Any" if we can't get a fully concrete model for calldata out).

A fully general (and hopefully correct) approach could look like: https://github.com/ethereum/hevm/issues/334

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


zoep (approved):
Looks good!

PR #329, opened 2023-07-26
SymExec: run whole expression simplifier on branch conditions

d-xo:
## Description

Runs the whole expression simplifier on branch conditions. Haven't really benchmarks, but should hopefully help us simplify down some stuff that the inline rules match (e.g. `Eq a a == Lit 1`).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

PR #338, opened 2023-07-31
A slightly improved version of #329

msooseth:
## Description
Effectively, an improved version of #329 with some extra simplifications. I also moved evalProp, because I believe it belongs in Expr.hs and because of another simplification step I was working on required it to be there. I think it's better there? Anyway. I also added a small improvement to mask removal when the mask is full-1.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


d-xo (approved):
lgtm

PR #339, opened 2023-07-31
Add `check` signature prefix to Dapp checking prefixes

msooseth:
## Description
This is needed for some tests is sc-comp. Added test to see if this `check` is now working. It seems to be.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


d-xo (approved):
lgtm

PR #340, opened 2023-07-31
The Purge

d-xo:
## Description

Removes a bunch of code and features in the interests of:

- Reducing our maintaince surface
- Making treewide refactors easier
- Reducing our support burden

Removed features / modules:

- The debugger (TTY.hs)
- All concrete unit test runners (including fuzzers / invariant testing) (UnitTest.hs)
- Cache / State serialization (Facts.hs)
- Traversal into the traces on the `End` nodes in Traversals.hs (complex and not relevant for generated SMT)
- Dev.hs / Debug.hs (not useful)
- Patricia.hs (not used anywhere)
- StorageLayout.hs (only used in TTY)
- Hexdump.hs (exact same thing is available via lib on hackage)
- all the EIP modifiers in FeeSchedule.hs (we don't care about historic gas pricing)

Most of these should be relatively uncontroversial, with the exception of the TTY.hs and Traversals.hs changes.

- TTY.hs is a very big piece of code that nobody on the team understands well. It is not really useful for internal debugging (tracing vs geth is more helpful), and has been slowly bitrotting over the last years.
- Traversing into the traces in the end nodes is complicated to write, and not relevant for what we want to use these traversals for. It could also result in extraneous terms being introduced into our smt queries (if the trace mentions some var that isn't relevant for the path condtions). IMO the real long term fix here is to pull the `End` nodes out of `Expr` so we don't need to consider them in this file, but this removes some maintaince weight in the short term. I would consider adding this back if others have opinions.

Still need to document changes / breakage before this can be merged. Would also wanna wait for symbolic contract addresses to avoid dealing with another round of treewide conflicts on that pr.

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog


msooseth (approved):
Wow, this is really cool. I will miss the debugging a bit, but honestly, it's best to focus on the things we are good at.

zoep (approved):
Nice cleanup!

src/EVM.hs
```diff
@@ -1643,9 +1641,8 @@ cheatActions =
                 _ -> vmError (BadCheatCode sig)
               _ -> vmError (BadCheatCode sig)
           else
-            let msg = encodeUtf8 "ffi disabled: run again with --ffi if you want to allow tests to call external scripts"
-            in vmError . Revert . ConcreteBuf $
-              abiMethod "Error(string)" (AbiTuple . V.fromList $ [AbiString msg]),
+            let msg = "ffi disabled: run again with --ffi if you want to allow tests to call external scripts"
+            in partial $ UnexpectedSymbolicArg vm.state.pc msg [],
```
d-xo, line 1645: This was not showing up for failing `prove` tests since we stopped treating `revert` as a failure there, switching it to `partial` surfaces the error again.

src/EVM.hs
```diff
@@ -1671,9 +1668,8 @@ cheatActions =
             Just a'@(LitAddr _) -> fetchAccount a' $ \_ ->
               accessStorage a' slot $ \res -> do
                 assign (#state % #returndata % word256At (Lit 0)) res
-                forceConcrete res "FIXME" $ \res' -> do
-                  let buf = ConcreteBuf $ word256Bytes res'
-                  copyBytesToMemory buf (Lit 32) (Lit 0) outOffset
+                let buf = writeWord (Lit 0) res (ConcreteBuf "")
```
d-xo, line 1671: We weren't handling symbolic values here

test/test.hs
```diff
@@ -659,20 +692,9 @@ tests = testGroup "hevm"
...
-        let testFile = "test/contracts/fail/invariantFail.sol"
-        runSolidityTest testFile "invariantFirst" >>= assertEqual "test result" False
-        runSolidityTest testFile "invariantCount" >>= assertEqual "test result" False
     , testCase "Cheat-Codes-Pass" $ do
         let testFile = "test/contracts/pass/cheatCodes.sol"
         runSolidityTest testFile ".*" >>= assertEqual "test result" True
-    , testCase "Cheat-Codes-Fail" $ do
-        let testFile = "test/contracts/fail/cheatCodes.sol"
-        runSolidityTestCustom testFile "testBadFFI" Nothing Nothing False Nothing Foundry >>= assertEqual "test result" False
-        runSolidityTestCustom testFile "test_prank_underflow" Nothing Nothing False Nothing Foundry >>= assertEqual "test result" False
```
d-xo, line 675: These were moved into `tests.hs` since they both did not fail as `prove` tests (since neither reverts nor partials trigger failures in `prove` tests).

d-xo:
Noticed a few addresses that I had forgotten to make symbolic in the previous pr, and I just changed them here inline while merging. Switching all the solidity tests to `prove` runners also exposed a few small bugs (e.g. not handling `ExtCodeSize` for symbolic addresses properly).

I'm gonna merge this once ci passes.

PR #342, opened 2023-07-31
SMT: inline emptybuf into call sites

d-xo:
## Description

This means that buf expressions that are built on top of an empty buf (i.e. all of memory / returndata) now have their own segregated array. This should hopefully be faster and avoid the "multiple write chains" error that we've been experiencing with cvc5.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
lgtm

Issue #333, opened 2023-07-28
Unable to parse `METHConstants.json`


d-xo:
Steps:

1. Clone: https://github.com/Philogy/meth-weth/tree/main
2. `forge build`
3. `hevm test`

result:

Error: unable to parse: /home/me/src/github.com/philogy/meth-weth/out/METHConstants.sol/METHConstants.json

PR #344, opened 2023-08-02
Handle nonexistent source map during JSON parsing

msooseth:
## Description
When the bytestring is empty, solidity doesn't create a "sourceMap"  JSON attribute. This leads us to a failure to parse the JSON, manifested in #333 . This fixes that. I opted for checking if the `sourceMap` is present. We can also check if `runtimeCode` is empty. What do you think is best? If the empty check is best, we can just revert the last commit. I think the checking for sourceMap is cleaner, but may hide some intent.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


d-xo (approved):
lgtm

PR #345, opened 2023-08-02
support bitwuzla

d-xo:
## Description

Adds support to hevm for bitwuzla. It has a new local search based bv solver, and it seems **fast**. At least after playing around with it for a short while it seems to be significantly better than z3 on non linear bv arithmetic. 

This required the following changes:

- explicit args to `push`/`pop`
- inline all type defs (not 100 % if this was actually needed)
- replace all `define-const` with `let` bindings (this is potentially better anyway)
- pipe solver stderr to stdout

Bitwuzla does not seem to have a timeout setting, not sure if there's a way to work around that...

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "bitwuzla-pkgs": {
+      "locked": {
+        "lastModified": 1690980179,
+        "narHash": "sha256-Tvnm9I3O06WGWd0C87jYcKmjUle02+BHyBf5Hm3QXnY=",
+        "owner": "d-xo",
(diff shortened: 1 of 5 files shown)


msooseth:
I rebased this on top of `main`, I hope that's good!

BTW, we should likely have at least
```
* replace all define-const with let bindings (this is potentially better anyway)
* pipe solver stderr to stdout
```

and allow bitwutzla as an option in CLI. I think it'd make sense?

d-xo:
Closed in favour of #438

PR #346, opened 2023-08-03
Ux tweaks

d-xo:
## Description

A bunch of small fixes and ux tweaks. This should hopefully be ok to review commit by commit.

- add `number` arg to test `cli` endpoint (was previously always failing cos it was expecting this)
- add default values for symbolic addresses
- prove tests now fail if there are no reachable assertion violations but all branches revert
- prove tests fall back to displaying calldata cexs as raw bytes if the model provided by the solver has a different length than the length of the arguments in the signature

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Oh. I chased this down with #357 :D We actually both put `number` at the same position in the record, I feel happy!

msooseth (approved):
Nice! I am thinking we may want to remove the `proveFail`, but that's another discussion to be had.

PR #352, opened 2023-08-04
abstract gas

d-xo:
## Description

Models gas as a `Maybe Word64`, where `Nothing` represents some totally arbitrary gas value. This allows us to remove most of the remaining calls to `forceConcrete` in `EVM.hs` and lets us start executing programs that read/write to symbolic memory indicies. We're still missing a few encoding tricks to deal with fully abstract dynamic data (i.e. copySlice with a symbolic size), but this is an important building block to make that work.

### Implementation

Gas is stored in a newtype wrapper over `Maybe Word64` called `Gas`. This has a `Num` instance defined where `Nothing` propogates. I didn't define `Ord` or `Integral` instances since they seemed unsafe: `toInteger` would allow the production of some concrete gas value from a totally symbolic one, and the semantics of comparison seemed very unclear (is `Nothing` < `0`?). Still this was enough that the changes required to the gas logic were relatively small (and in fact after my first pass fixing all type errors, I had only 38 failing cases in the blockchain tests).

If the `gas` field in the `vm` is `Nothing`, calls to the `GAS` opcode will return a term `SymGas x`, where `x` is a unique identifier for each new `SymGas` value. These should be constrained to be unequal in SMT. We could consider introducing a more precise scheme where we assign each new frame a unique id, and then `SymGas` terms would take a frame id, as well as a counter allowing us to order these terms over time (i.e. `SymGas <FrameId> <PerFrameIdx>`). This would allow us to add constraints in SMT that state that available gas decreases monotonically across call frames. I'm not sure if this level of precision is really useful, or what the impact on solver perf would be, so decided to keep things simpler and more abstract for now.

While refactoring here I also tidied up the unlawful `TryFrom` instances for for `W256 -> Word64` and replaced them with calls to a new `TruncateFrom`class. This was helpful when refactoring to avoid introducing any unexpected truncations. 

### Performance

This use of `Maybe` here introduces an additional pointer indirection when writing to gas, and we can no longer use the `UNPACK` pragma. Performance when running `ethereum-tests` has consequently taken a significant hit (`45.137 s ± 131 ms` here vs `34.235 s ± 510 ms` on main). One can imagine that executing with gas set to `Nothing` may be even faster than `main`, but I don't have any numbers to back that up yet.

I did experiment with some alternative approaches that would let us keep gas as an unpacked `Word64`. The most fruitful involved splitting all gas accounting into a separate `burnGas :: EVM s ()` function that was conditionally invoked based on a `modelGas :: Bool` field in `VM`, but untangling the gas logic for `create` / `delegatecall` was somewhat tricky, and I worried that separating the logic in that way would make maintainance and refactoring harder.

I plan on experimenting with alternatives and benchmarking a bit more before commiting to this approach.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Wow this is actually a cool idea!

Regarding this: `These should be constrained to be unequal in SMT. We could consider introducing a more precise scheme where we assign each new frame a unique id, and then SymGas terms would take a frame id, as well as a counter allowing us to order these terms over time (i.e. SymGas <FrameId> <PerFrameIdx>). This would allow us to add constraints in SMT that state that available gas decreases monotonically across call frames. I'm not sure if this level of precision is really useful, or what the impact on solver perf would be, so decided to keep things simpler and more abstract for now.` -- I'd not bother for the moment. We can do that later, perhaps as a flag like `--abst-gas-constr` or something. But firstly, it'd be great to have this working.

The performance hit on concrete execution I wouldn't worry about much. Maybe I'm wrong but that shouldn't impact the primary use-case much I think? I'm nowadays thinking that the really hard parts of HEVM are: (1) making sure interpretation is complete and correct (2) giving answers within a reasonable timeframe, say < 1hour. ~30% more spent in interpreting bytecode shouldn't really make a difference on (2), since the interpretation is likely to take < 10s I think? But abstract gas can make a huge positive dent on completeness. So overall, it seems like a big win to me.

src/EVM.hs
```diff
@@ -1456,16 +1477,19 @@ notStatic continue = do
...
+  case (amt, available) of
+    (G (Just n), G (Just a)) ->
+      if n <= a
+        then do
+          #state % #gas %= (subtract (G (Just n)))
+          #burned %= (+ (G (Just n)))
+          continue
+        else
+          vmError (OutOfGas a n)
+    _ -> noop
```
msooseth, line 1492:
```suggestion
    _ -> continue
```

It took me almost a day to figure this one out, but I learned so much  :) It's an infinite loop without it.

If you run:
```
:main -p "access-out-of-bounds-array"
```

Then eventually you'll end up in an `OpSstore`, which calls:

```
                      burn (storage_cost + cold_storage_cost) $ do
                        doStore
                        [...]
```

Where `doStore` is:
```
              let doStore = do
                    next
                    assign (#state % #stack) xs
                    modifying (#env % #contracts % ix self % #storage) (writeStorage x new)
```

i.e. it calls `next`, which advances the PC. Without this advancement of the PC, it will loop forever on `OpSstore`. Since `noop` doesn't call `continue`, which in this case is `doStore` (and some other things), which would call `next`, we never advance the PC, and stay inside `OpSstore` forever.

msooseth (changes requested):
So cool! I found one small bug, and added a way to fix it :)

msooseth:
Yaaaaay! With my very-very minor fix, it actually passes the tests! So nice :star_struck: 



BTW, this is kinda funny:

```
Checking file: /nix/store/sayhv9lb1nl58rrwsgx928bcng20hqd1-source/test/libyul/yulOptimizerTests/unusedAssignEliminator/for_multi_break.yul

  - Max Iterations Reached in contract: SymAddr "entrypoint" pc: 19

Found 720801 total pairs of endstates
```

That's.... a bit too many endstates :D Like, 0.7 million. It does finish it off, though! I'll see if there is some static analysis we could do to have less endstates.

msooseth:
This PR has been merged in another form, it was super-cool, thanks again! I'm closing this as this is now outdated. Thanks for that great PR!

Mate

PR #354, opened 2023-08-04
EVM: Add `deal` and `assume` cheatcodes

d-xo:
## Description

Adds deal and assume.

Right now we can't `deal` to a symbolic address that isn't a key in the contracts mapping. We need this to preserve soundness for symbolic addresses under aliasing. In the future we could potentially add some extra state to track balances for these kind of addresses (probably some smt array Addr -> W256 will be sufficient).

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/contracts/pass/cheatCodes.sol
+++ b/test/contracts/pass/cheatCodes.sol
@@ -128,7 +128,7 @@ contract CheatCodes is DSTest {
     }
 
     // this is not supported yet due to restrictions around symbolic address aliasing...
-    function prove_deal_unknown_address(address e, uint val) public {
+    function proveFail_deal_unknown_address(address e, uint val) public {
         hevm.deal(e, val);
         assert(e.balance == val);
     }
--- a/test/contracts/pass/cheatCodes.sol
+++ b/test/contracts/pass/cheatCodes.sol
@@ -4,6 +4,8 @@ import "ds-test/test.sol";
 
 interface Hevm {
     function warp(uint256) external;
+    function deal(address,uint256) external;
+    function assume(bool) external;
     function roll(uint256) external;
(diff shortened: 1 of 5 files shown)


msooseth (approved):
Wow, this is really cool! I actually understood the function passing and everything. Kinda fun! I am approving and I'll be merging too. Now I also understand the point of `proveFail` and I now see why it's important to have for testing.

test/contracts/pass/cheatCodes.sol
```diff
@@ -123,7 +127,56 @@ contract CheatCodes is DSTest {
         assertEq(prankster.prankme(), address(this));
     }
 
-    function test_prank_val() public {
+    // this is not supported yet due to restrictions around symbolic address aliasing...
+    function prove_deal_unknown_address(address e, uint val) public {
+        hevm.deal(e, val);
+        assert(e.balance == val);
+    }
```
msooseth, line 134:
```suggestion
    // TODO this is not supported yet due to restrictions around symbolic address aliasing...
    /*function prove_deal_unknown_address(address e, uint val) public {
        hevm.deal(e, val);
        assert(e.balance == val);
    }*/
```

msooseth: This is failing in CI, so perhaps we can just take it out, since it's a known issue, and add it as a TODO?

d-xo: prefixed it with `proveFail` for now

msooseth: Hmmm that can be a bit misleading, but it's OK!

msooseth (approved):
lgtm

PR #356, opened 2023-08-05
Solidity: better error messages for --sig

d-xo:
## Description

Better error messages when parsing solidity output in `functionAbi`. This entire procedure is a hack honestly, it would be nicer to have some pure haskell abi -> function sig parser...

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
So much better! Thank you! Yes, error messages could be better indeed :)

PR #358, opened 2023-08-10
Constant folding during path checking

msooseth:
## Description
This is a PR I have been sitting on, and had some partial experience regarding its performance, but I finally got some better proof :) While running PR #352 , I noticed:

```
Checking file: /nix/store/sayhv9lb1nl58rrwsgx928bcng20hqd1-source/test/libyul/yulOptimizerTests/unusedAssignEliminator/for_multi_break.yul

  - Max Iterations Reached in contract: SymAddr "entrypoint" pc: 19

Found 720801 total pairs of endstates
Reuse of previous queries was Useful in 364751 cases
"OK. Took 154.138271977s seconds"
```

Which is sad. On `main`, it behaves exactly the same. Sad. I thought... maybe my PR can fix it! And indeed it can:

```
Checking file: /nix/store/sayhv9lb1nl58rrwsgx928bcng20hqd1-source/test/libyul/yulOptimizerTests/unusedAssignEliminator/for_multi_break.yul

  - Max Iterations Reached in contract: symbolic(entrypoint) pc: 19

Found 37249 total pairs of endstates
Asking the SMT solver for 16900 pairs
Reuse of previous queries was Useful in 15286 cases
"OK. Took 7.409072883s seconds"
OK (7.42s)
```

That's some nice improvement right there! So now I'm more confident in the performance of this PR :) This PR basically adds a certain level of constant folding to branch exploration. It's a slightly enhanced version of #329 .

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -844,7 +844,15 @@ simplify e = if (mapExpr go e == e)
 
 
 simplifyProps :: [Prop] -> [Prop]
+<<<<<<< Updated upstream
 simplifyProps = remRedundantProps . map evalProp . flattenProps
+=======
+simplifyProps ps = if canBeSat then simplified else [PBool False]
+  where
+    simplified = remRedundantProps . map evalProp . flattenProps $ ps
+    canBeSat = constFoldProp simplified
+
+>>>>>>> Stashed changes
 
 -- | Evaluate the provided proposition down to its most concrete result
 evalProp :: Prop -> Prop
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -846,6 +846,8 @@ simplify e = if (mapExpr go e == e)
 simplifyProps :: [Prop] -> [Prop]
 simplifyProps = remRedundantProps . map evalProp . flattenProps
 
+
+
 -- | Evaluate the provided proposition down to its most concrete result
 evalProp :: Prop -> Prop
 evalProp prop =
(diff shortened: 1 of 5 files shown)


src/EVM/Expr.hs
```diff
@@ -1057,3 +1058,52 @@ evalProp prop =
...
+  }
+  deriving (Show)
+
+-- | Folds constants
+constFoldProp :: [Prop] -> ConstState
+constFoldProp ps = execState (mapM constProp ps) (ConstState mempty True)
+  where
+    constProp :: Prop -> State ConstState Prop
+    constProp prop = go (evalProp prop)
+    go :: Prop -> State ConstState Prop
```
d-xo, line 1075: Why do we need the return value here? Don't we throw it away with the call to `execState` later?

src/EVM/Expr.hs
```diff
@@ -1057,3 +1058,52 @@ evalProp prop =
...
+
+data ConstState = ConstState
+  { values :: Map.Map (Expr EWord) W256
+  , canBeSat :: Bool
+  }
+  deriving (Show)
+
+-- | Folds constants
+constFoldProp :: [Prop] -> ConstState
+constFoldProp ps = execState (mapM constProp ps) (ConstState mempty True)
```
d-xo, line 1071:
```suggestion
constFoldProp ps = execState (mapM (go . evalProp) ps) (ConstState mempty True)
```

src/EVM/Expr.hs
```diff
@@ -1057,3 +1058,52 @@ evalProp prop =
...
+
+
+data ConstState = ConstState
+  { values :: Map.Map (Expr EWord) W256
+  , canBeSat :: Bool
+  }
+  deriving (Show)
+
+-- | Folds constants
+constFoldProp :: [Prop] -> ConstState
```
d-xo, line 1070: I think we could just return `canBeSat` here?

src/EVM/SymExec.hs
```diff
@@ -308,9 +311,12 @@ interpret fetcher maxIter askSmtIters heuristic vm =
                     performQuery
                   -- otherwise just try both branches and don't ask the solver
```
d-xo, line 312: can we update the comment here?

src/EVM/Expr.hs
msooseth: Oh, yes, of course! Good idea :) Fixed in 88381efaaa93df187691d2cb404d8c8598ec5722

src/EVM/SymExec.hs
msooseth: Ooops, thanks, good point! Fixed in 2d29afa4542dedf66ac3afe9863580695b89cfad

src/EVM/Expr.hs
msooseth: Yeah, good point. This has now been fixed!

d-xo: we can still simplify further here and change the type of `go` to `Prop -> State ConstState ()`

msooseth: Ah yess! I finally understood what you meant, thanks! OK, it's done! I'm resolving now. Can you please check and then we can merge!

msooseth:
OK, so this should now be ready to merge I think? I need one more review for the fix to the `constFoldProp`'s return value being removed and we'll be good to go!! Yaaaaay

src/EVM/Expr.hs
```diff
@@ -845,6 +846,8 @@ simplify e = if (mapExpr go e == e)
 simplifyProps :: [Prop] -> [Prop]
 simplifyProps = remRedundantProps . map evalProp . flattenProps
 
+
+
```
d-xo, line 850:
```suggestion
```

d-xo (approved):

msooseth:
I realized we need a slightly more powerful prop simplifier, so I added a subfuction, `simpInnerExpr`.

I also realized we introduced a subtle "bug" that made some of our tests not very useful: We added simplification to `assertProps`, but we were using it to check simplification :D So now I added a function, `assertPropsNoSimp` that is called from `assertProps`, after simplification. This way, we can call it to check simplification correctness.

I also added a bunch of concrete tests, which highlighted the need for `simpInnerExpr`. This means this PR need a new review.... I'm sorry. I hope these new changes are not too big and make sense.

PR #359, opened 2023-08-10
Better eq simp

msooseth:
## Description
This does a bit better Eq IR simplification.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -786,8 +786,21 @@ simplify e = if (mapExpr go e == e)
     go (EVM.Types.Not (EVM.Types.Not a)) = a
 
     -- Some trivial min / max eliminations
-    go (Max (Lit 0) a) = a
-    go (Min (Lit 0) _) = Lit 0
+    go (Max a b) = case (a, b) of
+                    (Lit 0, _) -> b
+                    _ -> EVM.Expr.max a b
+    go (Min a b) = case (a, b) of
+                     (Lit 0, _) -> Lit 0
+                     _ -> EVM.Expr.min a b
+
+    -- Some trivial mul eliminations
+    go (Mul a b) = case (a, b) of
+                     (Lit 0, _) -> Lit 0
+                     (Lit 1, _) -> b
+                     _ -> mul a b
+    -- Some trivial div eliminations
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -685,9 +685,14 @@ simplify e = if (mapExpr go e == e)
       | a == b = Lit 1
       | otherwise = Lit 0
     go (Eq (Lit 0) (Sub a b)) = Eq a b
-    go o@(Eq a b)
-      | a == b = Lit 1
-      | otherwise = o
+    go (Eq a b@(Lit _)) = Eq b a
```
d-xo, line 688: This isn't covered by the `normArgs` stuff in Expr.hs?

src/EVM/Expr.hs
```diff
@@ -685,9 +685,14 @@ simplify e = if (mapExpr go e == e)
...
+    go (Eq a b) =
+      case (aSimp, bSimp) of
```
d-xo, line 690: Is this needed? Don't we already call `simplify` everywhere until fixpoint as part of `mapExpr`?

msooseth: Oh of course, what was I thinking. Sorry.

msooseth:
Honestly, this is not  a big gain. I think it's maybe worth it for the sanity of the produced equations. With this and the constant folding PR together, and `--quickcheck-replay 100` we get:

```
All 176 tests passed (296.21s)
*** Exception: ExitSuccess
```

Otherwise, we get:

```
All 176 tests passed (302.11s)
*** Exception: ExitSuccess
```

Basically, no difference (2% :laughing: ). So it doesn't do much. But it does make things more aligned, I think. Anyway, not important. I'll work on other stuff. I just spotted this and thought to fix it.

msooseth:
Ah, one sec, there is more:

```
ghci> Expr.simplify (Max (Lit 0) (Lit 5))
Lit 0x5
ghci> Expr.simplify (Max (Lit 5) (Lit 0))
Max (Lit 0x5) (Lit 0x0)
```

Let me make this more thorough.

msooseth:
OK, it should be better now... again, not sure this is the most important thing to do.

d-xo (approved):

PR #360, opened 2023-08-10
[Review needed] Adding LitAddr, but maybe this is the wrong fix?

msooseth:
## Description
This adds SMT for `LitAddr`, which is missing currently. Not sure this is the right "fix", though... @d-xo  when you are back, perhaps you could take a peek? Alternatively, maybe @arcz if you could take a look? I also added an test, currently with `ignoreTest` that triggers some of this behaviour -- however, it still doesn't correctly assert-fail, potentially due to

```

    msg: "call target has unknown code"
    program counter: 129
    arguments: 
      0x0000000000000000000000000000000000000000
```

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -210,6 +210,7 @@ referencedWAddrs term = foldTerm go mempty term
   where
     go = \case
       WAddr(a@(SymAddr _)) -> Set.singleton (formatEAddr a)
+      a@(LitAddr _) -> Set.singleton (formatEAddr a)
       _ -> mempty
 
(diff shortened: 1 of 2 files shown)


d-xo:
The reason we trigger a warning here is that sending eth via `call` executes code in the recipient. In this case `msg.sender` has unknown code, and we don't support execution of unknown code yet. 

I don't think that adding `LitAddr` to the SMT encoding is going to help here, but maybe I'm missing some context?

msooseth:
Yeah, probably I'm just lost :) Closing.

PR #362, opened 2023-08-11
Abstraction refinement of ReadWord

msooseth:
## Description
This adds an extra abstraction of the memory at the IR level. We have discussed this and I believe @d-xo had a diff that was either exactly this or similar to this. It's a single line difference on top of #330, namely:
```
e@(ReadWord {})-> abstrExpr e
```

This is actually very powerful and improves the speed a lot: ~30% faster runtime on unit tests... except for:

```
testCase "calldata beyond calldatasize is 0 (symbolic calldata)" $ do
testCase "calldata beyond calldatasize is 0 (concrete dalldata prefix)" $ do
testCase "calldata symbolic access" $ do
```

Where it struggles AND eventually it gives the wrong result. Very strange. I will work on this next. I want to understand why it doesn't work. There must be some weird reason and I have a feeling it may be a deeper issue. In general, I think this kind of abstraction-refinement at the IR level is likely very powerful, perhaps much more powerful than at the SMT level.

This is a work-in-progress.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -144,7 +144,7 @@ data Command w
       , smttimeout    :: w ::: Maybe Natural            <?> "Timeout given to SMT solver in seconds (default: 300)"
       , maxIterations :: w ::: Maybe Integer            <?> "Number of times we may revisit a particular branching point"
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
-      , abstractArithmetic    :: w ::: Bool                      <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
+      , abstractArithmetic    :: w ::: Bool             <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , abstractMemory    :: w ::: Bool                      <?> "Use abstraction-refinement for Memory. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , askSmtIterations :: w ::: Integer               <!> "1" <?> "Number of times we may revisit a particular branching point before we consult the smt solver to check reachability (default: 1)"
       }
@@ -216,7 +216,7 @@ equivalence cmd = do
                           , maxIter = cmd.maxIterations
                           , askSmtIters = cmd.askSmtIterations
                           , loopHeuristic = cmd.loopDetectionHeuristic
-                          , abstRefineArith = cmd.abstractArith
+                          , abstRefineArith = cmd.abstractArithmetic
                           , abstRefineMem = cmd.abstractMemory
                           , rpcInfo = Nothing
                           }
@@ -303,7 +303,7 @@ assert cmd = do
       maxIter = cmd.maxIterations,
       askSmtIters = cmd.askSmtIterations,
       loopHeuristic = cmd.loopDetectionHeuristic,
-      abstRefineArith = cmd.abstractArith,
+      abstRefineArith = cmd.abstractArithmetic,
       abstRefineMem = cmd.abstractMemory,
       rpcInfo = rpcinfo
     }
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -144,6 +144,7 @@ data Command w
       , smttimeout    :: w ::: Maybe Natural            <?> "Timeout given to SMT solver in seconds (default: 300)"
       , maxIterations :: w ::: Maybe Integer            <?> "Number of times we may revisit a particular branching point"
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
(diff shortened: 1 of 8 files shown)


cli/cli.hs
```diff
@@ -144,6 +144,7 @@ data Command w
       , smttimeout    :: w ::: Maybe Natural            <?> "Timeout given to SMT solver in seconds (default: 300)"
       , maxIterations :: w ::: Maybe Integer            <?> "Number of times we may revisit a particular branching point"
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
+      , abstRefine    :: w ::: Bool                      <?> "Use abstraction refinement"
```
d-xo, line 147: I think the name could be clearer here. I'm also wondering if we want to be more granular in the options that we expose? Perhaps at least a separate option for memory and the arithmetic ops?

d-xo: Do we wanna default to True or False here btw?

src/EVM/Fetch.hs
```diff
@@ -238,13 +238,13 @@ type Fetcher s = Query s -> IO (EVM s ())
 -- will be pruned anyway.
 checkBranch :: SolverGroup -> Prop -> Prop -> IO BranchCondition
 checkBranch solvers branchcondition pathconditions = do
-  checkSat solvers (assertProps [(branchcondition .&& pathconditions)]) >>= \case
+  checkSat solvers (assertProps [(branchcondition .&& pathconditions)] False) >>= \case
```
d-xo, line 241: Why always `False` here?

src/EVM/SMT.hs
```diff
@@ -100,6 +101,11 @@ data SMTCex = SMTCex
   }
   deriving (Eq, Show)
 
+
+-- | Used for abstraction-refinement of the SMT formula
```
d-xo, line 105:
```suggestion
-- | Used for abstraction-refinement of the SMT formula. Contains assertions that make our query fully precise. These will be added to the assertion stack if we get `sat` with the abstracted query.
```

src/EVM/Fetch.hs
msooseth: I didn't want to do abstaction-refinement in the middle of interpretation. I was afraid that the solver times out here due to abst-refine and then it's super-hard to debug. Would adding a TODO here be OK? Basically, I am a bit cautious.

Also, currently there is no easy way to pass the abt-refine configuration (yes/no/partial) here. So it may be quite confusing if I did abst-refinement there, even though the user asked for no abst-refinement. I didn't want to make this PR more complicated by finding a way to pass the configuration here. Maybe I'm wrong. What do you think?

d-xo: I guess we would have to add the option to the `PleaseAskSMT` query type? This might be a little awkward. I think I would expect the command line options here to be respected?

cli/cli.hs
```diff
@@ -144,6 +144,8 @@ data Command w
       , smttimeout    :: w ::: Maybe Natural            <?> "Timeout given to SMT solver in seconds (default: 300)"
       , maxIterations :: w ::: Maybe Integer            <?> "Number of times we may revisit a particular branching point"
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
+      , abstractArith    :: w ::: Bool                      <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
```
d-xo, line 147:
```suggestion
      , abstractArithmetic    :: w ::: Bool                      <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
```

d-xo, line 147: we should also add these options to `symbolic` and `equivalence` I think?

cli/cli.hs
d-xo: I would be  fine with switching to cvc5 as the default, but agree that it can be done in a separate pr.

src/EVM/SMT.hs
```diff
@@ -138,65 +144,115 @@ declareIntermediates bufs stores =
...
+                  let
+                    next = s.count
+                    bs' = Map.insert e next s.words
+                  put $ s{words=bs', count=next+1}
+                  pure $ Var (TS.pack ("abst_" ++ show next))
+
+smt2Line :: Builder -> SMT2
+smt2Line txt = SMT2 [txt] mempty mempty
+
+assertProps :: Bool -> Bool -> [Prop] -> SMT2
```
d-xo, line 196: maybe we can add a `CheckSatOpts` or smth here? I don't really love having two `Bool` params...

src/EVM/Fetch.hs
msooseth: As discussed, we'll do this as part of a more global resturcturing. Otherwise, even if done in a precise way, it's about 160 LoC change, which is kinda sad.

msooseth: For the moment, I added a default, that can be changed and will change it throughout the codebase. However, that's not configurable from the command line. We'll fix this later.

The 160-line diff can be seen here: 7402256000b3d9234a91ad4ea7d8493860d01f09

cli/cli.hs
msooseth: Ah, good point! Yes, just added it!

src/EVM/SMT.hs
msooseth: Good point! I now have a type, `AbstRefineConfig` so it's a lot easier to pass around and to understand!

PR #363, opened 2023-08-24
Simplify Props

msooseth:
## Description
We don't simplify `Prop`-s -- we simplify the expressions they utilize, but not the `Prop`-s themselves. So we quite often end up with stuff like:

```
      (PAnd
        (PGEq
          (Var "arg1")
          0x0
        )
        (PLEq
          (Var "arg1")
          0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
        )
      )
```

in the `Expr 'End`. Notice that the above is always TRUE. 

With this PR, we  get rid of these, and more. The basic idea is:

* Take out the `Prop` list from `Success`/`Failure`/`Partial`
* Flatten it by making `[PAnd x y]` ->` [x,y]`
* Do trivial simplifications on the `PLT`,`PLEq`, and transform all `GT` to `LT` and `GEq` to `LEq`
* Remove all `PBool True`-s

This gets rid of some of the redundant `(assert true)` in our SMT2, and makes it smaller in general.  We get something like this even for a trivial SMT2:

```
 ; logic
@@ -233,13 +233,9 @@
 (assert (and (bvuge (max (_ bv36 256) txdata_length) (_ bv36 256)) (bvult (max (_ bv36 256) txdata_length) (_ bv18446744073709551616 256))))
 (assert (and (bvuge arg1 (_ bv0 256)) (bvule arg1 (_ bv115792089237316195423570985008687907853269984665640564039457584007913129639935 256))))
 (assert (= arg1 (_ bv10 256)))
-(assert true)
-(assert true)
-(assert true)
-(assert true)
 (assert (not (= (ite (= txvalue (_ bv0 256)) (_ bv1 256) (_ bv0 256)) (_ bv0 256))))
-(assert (and (bvuge (max (_ bv36 256) txdata_length) (_ bv36 256)) (bvult (max (_ bv36 256) txdata_length) (_ bv18446744073709551616 256))))
-(assert (and (bvuge arg1 (_ bv0 256)) (bvule arg1 (_ bv115792089237316195423570985008687907853269984665640564039457584007913129639935 256))))
+(assert (bvule (_ bv36 256) (max (_ bv36 256) txdata_length)))
+(assert (bvult (max (_ bv36 256) txdata_length) (_ bv18446744073709551616 256)))
 (assert (= (bvsub arg1 (_ bv10 256)) (_ bv0 256)))
 (assert (not (= (ite (= txvalue (_ bv0 256)) (_ bv1 256) (_ bv0 256)) (_ bv0 256))))
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -13,7 +13,7 @@ import Data.ByteString qualified as BS
 import Data.DoubleWord (Int256, Word256(Word256), Word128(Word128))
 import Data.List
 import Data.Map.Strict qualified as Map
-import Data.Maybe (mapMaybe)
+import Data.Maybe (mapMaybe, isJust)
 import Data.Semigroup (Any, Any(..), getAny)
 import Data.Vector qualified as V
 import Data.Vector.Storable qualified as VS
@@ -630,34 +630,38 @@ getAddr (GVar _) = error "cannot determine addr of a GVar"
 -- ** Whole Expression Simplification ** -----------------------------------------------------------
 
 -- Calls prop simplification for the appropriate types of Expr
-simplifyPropExp :: ([Prop] -> [Prop]) -> Expr b -> Expr b
-simplifyPropExp f expr = case expr of
+applyToProps :: ([Prop] -> [Prop]) -> Expr b -> Expr b
+applyToProps f expr = case expr of
   Failure a b c -> Failure (f a) b c
   Partial a b c -> Partial (f a) b c
   Success a b c d -> Success (f a) b c d
-  ITE (Lit 0) _ c -> simplifyPropExp f c
-  ITE (Lit _) b _ -> simplifyPropExp f b
-  ITE a b c-> ITE a (simplifyPropExp f b) (simplifyPropExp f c)
+  ITE (Lit 0) _ c -> applyToProps f c
+  ITE (Lit _) b _ -> applyToProps f b
+  ITE a b c-> ITE a (applyToProps f b) (applyToProps f c)
   a -> a
 
--- Removes PBool True from list
-removeTrueProps :: [Prop] -> [Prop]
-removeTrueProps p = filter (\x -> x /= PBool True) . nubOrd $ p
+remRedundantProps :: [Prop] -> [Prop]
+remRedundantProps p = collapseFalse . filter (\x -> x /= PBool True) . nubOrd $ p
+  where
+    collapseFalse :: [Prop] -> [Prop]
+    collapseFalse ps = if isJust $ find (== PBool False) ps then [PBool False] else ps
 
--- Makes [PAnd a b] into [a,b]
-flattenProps :: [Prop] -> [Prop]
-flattenProps (a:ax) = case a of
-                       PAnd x1 x2 -> x1:x2:flattenProps ax
-                       x -> x:flattenProps ax
-flattenProps [] = []
 
-simplifyProp :: Expr a -> Expr a
-simplifyProp p = simplifyPropExp (removeTrueProps . map (mapProp' go) . flattenProps) p
+simplifyProp :: Prop -> Prop
+simplifyProp p = mapProp' go p
   where
     go :: Prop -> Prop
+    -- Handle "OR". Notice: "AND" is flattened and does not exist
+    go (POr (PBool True) _) = PBool True
+    go (POr _ (PBool True)) = PBool True
+
+    -- PImp
+    go (PImpl (PBool True) b) = go b
+    go (PImpl (PBool False) _) = PBool True
+
     -- rewrite everything as LEq or LT
     go (PGEq a b) = go $ PLEq b a
-    go (PGT a b) = go $ PLT a b
+    go (PGT a b) = go $ PLT b a
 
     -- trivial LT/LEq comparisions
     go (PLT  (Var _) (Lit 0)) = PBool False
@@ -671,6 +675,16 @@ simplifyProp p = simplifyPropExp (removeTrueProps . map (mapProp' go) . flattenP
 
     -- fallback
     go a = a
+
+simplifyPropExpr :: Expr a -> Expr a
+simplifyPropExpr e = applyToProps (remRedundantProps . map simplifyProp . flattenProps) e
+  where
+    -- Makes [PAnd a b] into [a,b]
+    flattenProps :: [Prop] -> [Prop]
+    flattenProps (a:ax) = case a of
+                           PAnd x1 x2 -> x1:x2:flattenProps ax
+                           x -> x:flattenProps ax
+    flattenProps [] = []
 
 
 -- | Simple recursive match based AST simplification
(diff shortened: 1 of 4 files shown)


src/EVM/Expr.hs
```diff
@@ -628,6 +633,63 @@ getAddr (GVar _) = error "cannot determine addr of a GVar"
...
+  a -> a
+
+remRedundantProps :: [Prop] -> [Prop]
+remRedundantProps p = collapseFalse . filter (\x -> x /= PBool True) . nubOrd $ p
+  where
+    collapseFalse :: [Prop] -> [Prop]
+    collapseFalse ps = if isJust $ find (== PBool False) ps then [PBool False] else ps
+
+
+simplifyProp :: Prop -> Prop
```
d-xo, line 654: whats the difference between this and `evalProp`?

msooseth: Oh!! NICE catch! Actually, this does less AND more at the same time :D OK, let me merge the two and use evalProp here :)

msooseth: NICE! So much better, thank you! I am sometimes super-blind. You are so meticulous, it's so much better this way.

So b44c31a3d43f36046949d0c7f012bebad932e27e remedies this -- now using `evalProp`, and also improved `evalProp`. I have something to learn from the way `evalProp` was written, some constant simplifications were so much better done!

src/EVM/Expr.hs
```diff
@@ -628,6 +633,32 @@ getAddr (GVar _) = error "cannot determine addr of a GVar"
 
 -- ** Whole Expression Simplification ** -----------------------------------------------------------
 
+-- Calls prop simplification for the appropriate types of Expr
+applyToProps :: ([Prop] -> [Prop]) -> Expr b -> Expr b
```
d-xo, line 637: do we actually need this? or can we just use `foldl'` and `fmap` combined with `mapProp`?

src/EVM/Expr.hs
```diff
@@ -628,6 +633,32 @@ getAddr (GVar _) = error "cannot determine addr of a GVar"
...
+  ITE a b c-> ITE a (applyToProps f b) (applyToProps f c)
+  a -> a
+
+remRedundantProps :: [Prop] -> [Prop]
+remRedundantProps p = collapseFalse . filter (\x -> x /= PBool True) . nubOrd $ p
+  where
+    collapseFalse :: [Prop] -> [Prop]
+    collapseFalse ps = if isJust $ find (== PBool False) ps then [PBool False] else ps
+
+simplifyPropExpr :: Expr a -> Expr a
```
d-xo, line 653:
```suggestion
simplifyProps :: Expr End -> Expr End
```

I think we can restrict the type here?

src/EVM/Expr.hs
```diff
@@ -1028,30 +1059,39 @@ evalProp prop =
...
     go (PAnd (PBool False) _) = PBool False
     go (PAnd _ (PBool False)) = PBool False
-
-    go (POr (PBool l) (PBool r)) = PBool (l || r)
     go (POr (PBool True) _) = PBool True
     go (POr _ (PBool True)) = PBool True
+    go (POr (PBool l) (PBool r)) = PBool (l || r)
 
+    -- Imply
     go (PImpl (PBool l) (PBool r)) = PBool ((Prelude.not l) || r)
```
d-xo, line 1087: I know it's not part of this PR, but I'm curious what benefit we get from this? Is it easier for the solver? Or does it make constant folding more powerful?

src/EVM/Expr.hs
d-xo: actually I think this can just be inlined into the existing expression simplifier by calling `remRedundantProps . map evalProp . flattenProps` on `Success` / `Failure` / `Partial`? Then we can get rid of `applyToProps` and `simplifyPropExpr`?

src/EVM/Expr.hs
msooseth: I'm resolving because this has been fixed.

src/EVM/Expr.hs
msooseth: Hmmm... `mapProp` doesn't apply this simplification:

```
  ITE (Lit 0) _ c -> applyToProps f c
  ITE (Lit _) b _ -> applyToProps f b
  ITE a b c-> ITE a (applyToProps f b) (applyToProps f c)
```

which does partial constant folding of `ITE`. The function `simplify` does apply this simplification, but that can't be called here, because it deals with a different type, right? Or is there something I'm missing? 

As for the rest, I don't even know how to do `fold'` and `fmap` such that it only applies to `ITE` and not the other things. That's what we'd need, right? Sorry, what you wrote is a bit mystical to me :D

src/EVM/Expr.hs
msooseth: Pushed as 92638ffb702592d77d4886be7ac8bf6f1982f222:

```
     -- Imply
-    go (PImpl (PBool l) (PBool r)) = PBool ((Prelude.not l) || r)
+    go (PImpl _ (PBool True)) = PBool True
     go (PImpl (PBool True) b) = b
     go (PImpl (PBool False) _) = PBool True
```

Basically, True is implied by everything. This means that for all combinations of `(PBool a)` and `(PBool b)`, the original rewrite is now useless. So are now more generic! Cool idea :)

msooseth: Resolving.

src/EVM/Expr.hs
msooseth: I made this a bit cleaner in 306f7087118856193460d765122977666a5f5fa5 Let me know what you think. However, it's not inlined, because I don't understand how to do that. Let's discuss in person :)

src/EVM/Expr.hs
```diff
@@ -628,6 +633,31 @@ getAddr (GVar _) = error "cannot determine addr of a GVar"
...
+      ITE (Lit 0) _ c -> applyToProps f c
+      ITE (Lit _) b _ -> applyToProps f b
+      ITE a b c-> ITE a (applyToProps f b) (applyToProps f c)
+      a -> a
+    -- removes redundant (constant True/False) props
+    remRedundantProps :: [Prop] -> [Prop]
+    remRedundantProps p = collapseFalse . filter (\x -> x /= PBool True) . nubOrd $ p
+      where
+        collapseFalse :: [Prop] -> [Prop]
+        collapseFalse ps = if isJust $ find (== PBool False) ps then [PBool False] else ps
```
d-xo, line 660:
```suggestion
      collapseFalse :: [Prop] -> [Prop]
      collapseFalse ps = if isJust $ find (== PBool False) ps then [PBool False] else ps
```

msooseth: Ah, good point, fixed in 74425f39c3742e4396edae5aef67361fb3ece276

d-xo (approved):
maybe just one or two small concrete testcases and then this lgtm

PR #366, opened 2023-08-28
A fairly regular `CopySlice` simplification

msooseth:
## Description
This simplifies away some of the `ConcreteBuf` when we `CopySlice` from a `WriteWord`+`ConcreteBuf` combo. If the `ConcreteBuf` is larger than what we will copy, the last parts of the `ConcreteBuf` is useless.

Without:
```
            (CopySlice
              srcOffset: 0x0
              dstOffset: 0x0
              size:      0x40
              src:
                (WriteWord
                  idx:
                    0x0
                  val:
                    (WAddr
                      (SymAddr "arg1")
                    )
                )
                (ConcreteBuf
                  Length: 96 (0x60) bytes
                  0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0040:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0050:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 80   ................
                )
            )
            (ConcreteBuf "")
```

With:

```
            (CopySlice
              srcOffset: 0x0
              dstOffset: 0x0
              size:      0x40
              src:
                (WriteWord
                  idx:
                    0x0
                  val:
                    (WAddr
                      (SymAddr "arg1")
                    )
                )
                (ConcreteBuf
                  Length: 64 (0x40) bytes
                  0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                  0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                )
            )
```

Notice that there was no need for all that 96B `ConcreteBuf` -- the `CopySlice` will cut it out anyway. So we only keep 64B, which is what `CopySlice` will copy.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -663,8 +663,15 @@ simplify e = if (mapExpr go e == e)
     go (WriteWord a b c) = writeWord a b c
 
     go (WriteByte a b c) = writeByte a b c
-    go (CopySlice srcOff@(Lit 0) dstOff size@(Lit sz) (WriteWord (Lit 0) value (ConcreteBuf buf)) dst) = (CopySlice srcOff dstOff size (WriteWord (Lit 0) value (ConcreteBuf simplifiedBuf)) dst)
-        where simplifiedBuf = BS.take (unsafeInto sz) buf
+    go orig@(CopySlice srcOff@(Lit n) dstOff size@(Lit sz)
+        -- It doesn't matter what wOffs we write to, because only the first
+        -- n+sz of ConcreteBuf will be used by CopySlice
+        (WriteWord wOff value (ConcreteBuf buf)) dst)
+          -- Let's not deal with overflow
+          | n+sz >= n && n+sz >= sz = (CopySlice srcOff dstOff size
+              (WriteWord wOff value (ConcreteBuf simplifiedBuf)) dst)
+          | otherwise = orig
+            where simplifiedBuf = BS.take (unsafeInto (n+sz)) buf
     go (CopySlice a b c d f) = copySlice a b c d f
 
     go (IndexWord a b) = indexWord a b
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -663,6 +663,8 @@ simplify e = if (mapExpr go e == e)
     go (WriteWord a b c) = writeWord a b c
 
     go (WriteByte a b c) = writeByte a b c
+
+    -- truncate some concrete source buffers to the portion relevant for the CopySlice if we're copying a fully concrete region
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -663,6 +663,8 @@ simplify e = if (mapExpr go e == e)
     go (WriteWord a b c) = writeWord a b c
 
     go (WriteByte a b c) = writeByte a b c
+    go (CopySlice srcOff@(Lit 0) dstOff size@(Lit sz) (WriteWord (Lit 0) value (ConcreteBuf buf)) dst) = (CopySlice srcOff dstOff size (WriteWord (Lit 0) value (ConcreteBuf simplifiedBuf)) dst)
```
d-xo, line 666: how hard would it be to extend this to work for all literals src offsets?

d-xo (approved):
could be a little more general maybe, but otherwise lgtm

src/EVM/Expr.hs
msooseth: Yeah, it'd be possible :D I am always terrified of overflows. I'll try to figure the overflows it out. It can be quite painful :D I'll get back to this tomorrow.

msooseth: I have now updated this to be more generic. It is more generic in every way. I put a guard to make sure it does not overflow.

msooseth:
I have now made it more general, let me know what you think! :)

src/EVM/Expr.hs
```diff
@@ -663,6 +663,15 @@ simplify e = if (mapExpr go e == e)
     go (WriteWord a b c) = writeWord a b c
 
     go (WriteByte a b c) = writeByte a b c
+    go orig@(CopySlice srcOff@(Lit n) dstOff size@(Lit sz)
```
d-xo, line 666:
```suggestion

    -- truncate some concrete source buffers to the portion relevant for the CopySlice if we're copying a concretely sized region
    go orig@(CopySlice srcOff@(Lit n) dstOff size@(Lit sz)
```

d-xo (approved):
lgtm!

PR #367, opened 2023-08-30
SStore/SLoad rewrites

msooseth:
## Description
This rewrites SStore/SLoad to be way more efficient. The changes are very significant, the `Expr` becomes even humanly readable. It does this by matching on typical `Keccak` use in SStore/SLoad, and reverse lookup of array indices.

This only deals with ONE dynamic array, because of the lonely and fixed `0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563`. It's possible to deal with more than one, by hashing 0...N. Hashing of a 32B `0x0` is the value above. We can do better later. Note that while it deals with only one dynamic array, it understands the first 256 elements of that array.

For an example program that has multiple maps and an array:

```solidity
        contract MyContract {
          uint[] a;
          mapping(uint => uint) items1;
          mapping(uint => uint) items2;
          function transfer(uint acct, uint val1, uint val2) public {
            uint beforeVal1 = items1[acct];
            uint beforeVal2 = items2[acct];
            unchecked {
              items1[acct] = val1+1;
              items2[acct] = val2+2;
              a[0] = val1 + val2 + 1;
              a[1] = val1 + val2 + 2;
              assert(items1[acct]+items2[acct]+a[0]+a[1] > beforeVal1 + beforeVal2);
            }
          }
        }
```

We get:

```diff
-                    (SLoad
-                      slot:
-                        (Keccak
-                          (CopySlice
-                            srcOffset: 0x0
-                            dstOffset: 0x0
-                            size:      0x40
-                            src:
-                              (WriteWord
-                                idx:
-                                  0x0
-                                val:
-                                  (Var "arg1")
-                              )
-                              (ConcreteBuf
-                                Length: 64 (0x40) bytes
-                                0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 01   ................
-                              )
-                          )
-                          (ConcreteBuf "")
-                        )
-                      storage:
-                        (SStore
-                          slot:
-                            (Keccak
-                              (ConcreteBuf
-                                Length: 32 (0x20) bytes
-                                0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                              )
-                            )
-                          val:
-                            (Add
-                              0x1
-                              (Add
-                                (Var "arg2")
-                                (Var "arg3")
-                              )
-                            )
-                        )
-                        (SStore
-                          slot:
-                            (Add
-                              (Keccak
-                                (ConcreteBuf
-                                  Length: 32 (0x20) bytes
-                                  0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                  0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                )
-                              )
-                              0x1
-                            )
-                          val:
-                            (Add
-                              0x2
-                              (Add
-                                (Var "arg2")
-                                (Var "arg3")
-                              )
-                            )
-                        )
-                        (SStore
-                          slot:
-                            (Keccak
-                              (CopySlice
-                                srcOffset: 0x0
-                                dstOffset: 0x0
-                                size:      0x40
-                                src:
-                                  (WriteWord
-                                    idx:
-                                      0x0
-                                    val:
-                                      (Var "arg1")
-                                  )
-                                  (ConcreteBuf
-                                    Length: 64 (0x40) bytes
-                                    0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 02   ................
-                                  )
-                              )
-                              (ConcreteBuf "")
-                            )
-                          val:
-                            (Add
-                              0x2
-                              (Var "arg3")
-                            )
-                        )
-                        (SStore
-                          slot:
-                            (Keccak
-                              (CopySlice
-                                srcOffset: 0x0
-                                dstOffset: 0x0
-                                size:      0x40
-                                src:
-                                  (WriteWord
-                                    idx:
-                                      0x0
-                                    val:
-                                      (Var "arg1")
-                                  )
-                                  (ConcreteBuf
-                                    Length: 64 (0x40) bytes
-                                    0000:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0010:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
-                                    0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 01   ................
-                                  )
-                              )
-                              (ConcreteBuf "")
-                            )
-                          val:
-                            (Add
-                              0x1
-                              (Var "arg2")
-                            )
-                        )
-                        (AbstractStore (SymAddr "entrypoint"))
+                    (Add
+                      0x1
+                      (Var "arg2")
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


src/EVM/Expr.hs
```diff
@@ -575,26 +576,58 @@ stripWrites off size = \case
...
+readStorage slot s@(AbstractStore _) = Just $ SLoad slot s
+readStorage (Lit l) (ConcreteStore s) = Lit <$> Map.lookup l s
+readStorage slot store@(ConcreteStore _) = Just $ SLoad slot store
+readStorage slot s@(SStore prevSlot val prev) = case (prevSlot, slot) of
+  -- if address and slot match then we return the val in this write
+  _ | prevSlot == slot -> Just val
+  -- if the slots don't match (see previous guard) and are lits, we can skip this write
+  (Lit _, Lit _) -> readStorage slot prev
+  -- special case of map + map (different slot) -> skip write
+  (Keccak (CopySlice (Lit 0) (Lit 0) (Lit 0x40) (WriteWord (Lit 0) _ (ConcreteBuf a)) (ConcreteBuf "")), Keccak (CopySlice (Lit 0) (Lit 0) (Lit 0x40) (WriteWord (Lit 0) _ (ConcreteBuf b)) (ConcreteBuf ""))) | dontMatchSlot a b -> readStorage slot prev
```
d-xo, line 588: We might benefit from a few pattern synonyms here for readability?

msooseth:
d76114fa14e4f69985946be2b1ea8302da6aad83 makes it generic to all 0..255 arrays.

d-xo (approved):
maybe we should start generating array and mapping slots as part of our fuzz generators? but apart from that this looks amazing, super nice!

src/EVM.hs
```diff
@@ -1321,7 +1321,7 @@ accessStorage addr slot continue = do
                 contract <- preuse (#cache % #fetched % ix addr')
                 case contract of
                   Nothing -> internalError "contract marked external not found in cache"
-                  Just fetched -> case readStorage (Lit slot') fetched.storage of
+                  Just fetched -> case readStorage (Expr.structureArraySlots (Lit slot')) fetched.storage of
```
d-xo, line 1324: I guess these are here to make sure we only call it once? I'm not sure how I feel about having to remember to do this every time we call `readStorage`? Maybe we can just do it once at the start of `readStorage` and then pull all the actual logic out into some `go` in a where block that doesn't call `structureArraySlots`?

msooseth: Good point. I fixed this now, a lot cleaner!

msooseth:
> maybe we should start generating array and mapping slots as part of our fuzz generators? but apart from that this looks amazing, super nice!

Ah, that's a good point. OK, let me do that tomorrow, that really ought to be done before this PR is merged. I think it's shaping up nicely otherwise. I added a bunch of test cases now.

test/test.hs
```diff
@@ -240,6 +411,13 @@ tests = testGroup "hevm"
     , testProperty "simpProp-equivalence-sym" $ \(LitProp p) -> ioProperty $ do
         let simplified = pand (Expr.simplifyProps [p])
         checkEquivProp simplified p
+    -- This would need to be a fuzz test I think. The SMT encoding of Keccak is not precise
+    -- enough for this to succeed
+    , ignoreTest $ testProperty "storage-slot-simp-property" $ \(StorageExp s) -> ioProperty $ do
```
d-xo, line 416: we need this I think: https://github.com/ethereum/hevm/issues/380  

The counterexamples I'm seeing here give two outputs that are very close together. I'm fine to merge as is and add this in a separate pr.

d-xo (approved):
so nice! <3 <3

PR #376, opened 2023-09-11
More precise smt address encoding

d-xo:
## Description

This adds some more assertions during encoding around the values of symbolic addresses. Still needed is a pass that asserts that all symbolic addresses are pairwise distinct with all known concrete addresses.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
Wow, this is actually really good!

msooseth:
I have fixed this up and it's working now, yay! The new PR is at: #641 

Closing so we can review and merge that PR instead :)

PR #377, opened 2023-09-11
Fix in init code arguments

zoep:
## Description

Correctly pass the arguments when creating new contracts 

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -192,7 +192,7 @@ abstractVM
   -> ST s (VM s)
 abstractVM cd contractCode maybepre create = do
   let value = TxValue
-  let code = if create then InitCode contractCode mempty else RuntimeCode (ConcreteRuntimeCode contractCode)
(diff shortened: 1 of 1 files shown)


src/EVM/SymExec.hs
```diff
@@ -192,7 +192,7 @@ abstractVM
   -> ST s (VM s)
 abstractVM cd contractCode maybepre create = do
   let value = TxValue
-  let code = if create then InitCode contractCode mempty else RuntimeCode (ConcreteRuntimeCode contractCode)
+  let code = if create then InitCode contractCode (fst cd) else RuntimeCode (ConcreteRuntimeCode contractCode)
```
d-xo, line 195: hm, I'm kind of in two minds here, this is obviously a very clean change, but it's also a bit of an abuse of semantics since calldata isn't really a thing that makes sense for constructors. 

If we go this route we should also maybe make sure to avoid writing calldata to the vm state if `create` is true?

d-xo (approved):
thought about it a bit more and decided that I'm fine with this change

PR #379, opened 2023-09-18
EVM: respect prank address in callChecks

### LINKED ISSUE #343: prank call check wrong sender balance
same as https://github.com/ethereum/hevm/pull/245 
check wrong sender balance after prank
reproduce
echidna 2.2.1 hevm 0.51.3
```
let this.balance=0
let msg.sender.balance=100
let to = 0x10000
hevm.prank(msg.sender)
address(to).call{value:1}("")  //faild with BalanceTooLow 0x1,0x0
```

d-xo:
## Description

Previously we would incorrectly fail here if sending value after calling `vm.prank` since `callChecks` did not take the value of `overrideCaller` into account when checking if the caller had sufficient balance.

There is some duplication of the logic in `transfer` here, but I'm not sure how to handle it better, since I do not want `transfer` to be incomplete, and we can't rely on `transfer` for callcode and delegatecall (which don't actually move any eth).

This fixes #343.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1013,14 +1013,9 @@ callChecks this xGas xContext xTo xValue xInOffset xInSize xOutOffset xOutSize x
       availableGas <- use (#state % #gas)
       let recipientExists = accountExists xContext vm
       (cost, gas') <- costOfCall fees recipientExists xValue availableGas xGas xTo
-      burn (cost - gas') $
-        branch (Expr.gt xValue this.balance) $ \case
-          True -> do
-            assign (#state % #stack) (Lit 0 : xs)
-            assign (#state % #returndata) mempty
-            pushTrace $ ErrorTrace (BalanceTooLow xValue this.balance)
-            next
-          False ->
+      let from = fromMaybe vm.state.contract vm.config.overrideCaller
+      fromBal <- preuse $ #env % #contracts % ix from % #balance
+      let checkCallDepth =
             if length vm.frames >= 1024
             then do
               assign (#state % #stack) (Lit 0 : xs)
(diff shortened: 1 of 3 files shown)


msooseth:
I'm re-kicking it, seems to have died during fuzzing, either running out of time or out of stack.

src/EVM.hs
```diff
@@ -1013,21 +1013,50 @@ callChecks this xGas xContext xTo xValue xInOffset xInSize xOutOffset xOutSize x
...
+          burn (cost - gas') $
+            branch (Expr.gt xValue fb) $ \case
+              True -> do
+                assign (#state % #stack) (Lit 0 : xs)
+                assign (#state % #returndata) mempty
+                pushTrace $ ErrorTrace (BalanceTooLow xValue this.balance)
+                next
+              False -> checkCallDepth
+
+        -- from is not in the state, we insert if if safe to do so and run the checks again
```
msooseth, line 1041:
```suggestion
        -- from is not in the state, we insert it if safe to do so and run the checks again
```

msooseth (approved):
lgtm

d-xo:
looks like we're hitting some failures in `writeWord` on this branch, which is very confusing since I didn't touch that code at all in this pr. I noticed them locally, but couldn't reproduce them consistently enough to debug properly and so assumed they weren't related to my changes and decided to push anyway. I'm now suspicious that they are somehow related since they seem to repro pretty consistenly in ci, and we never see them on `main`....

msooseth:
I'm re-kicking it. That bug you hit is a randomly occurring, known one, I think. I think it'd be worthwhile to get this merged :) Let's see if it works after re-kick. If it does, I'll merge :)

PR #382, opened 2023-09-20
Keccak improvements: minimum distance, all keccaks extracted including concrete ones

### LINKED ISSUE #380: Assert Large Gaps for Keccak
We should more precisly encode keccak to ensure that different inputs do not give outputs that are close together. This is needed to avoid spurious counterexamples for arrays.

msooseth:
## Description
This adds minimum distance requirement to Keccak and enables the ignored automated test (which now passes).

I also updated the minimum value of Keccak from `Lit 50` to `Lit 2**32`. This is a valid assumption, `Lit 50` is way too conservative, and seems like a random value. We can make it `Lit 256` but `2**32` is fine.

Closes #380 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Keccak.hs
+++ b/src/EVM/Keccak.hs
@@ -14,31 +14,32 @@ import EVM.Traversals
 import EVM.Types
 import EVM.Expr
 
-newtype BuilderState = BuilderState
-  { keccaks :: Set (Expr EWord) }
+
+newtype KeccakStore = KeccakStore
+  { keccakEqs :: Set (Expr EWord) }
   deriving (Show)
 
-initState :: BuilderState
-initState = BuilderState { keccaks = Set.empty }
+initState :: KeccakStore
+initState = KeccakStore { keccakEqs = Set.empty }
 
-go :: forall a. Expr a -> State BuilderState (Expr a)
-go = \case
+keccakFinder :: forall a. Expr a -> State KeccakStore (Expr a)
+keccakFinder = \case
   e@(Keccak _) -> do
     s <- get
-    put $ s{keccaks=Set.insert e s.keccaks}
+    put $ s{keccakEqs=Set.insert e s.keccakEqs}
     pure e
   e -> pure e
 
-findKeccakExpr :: forall a. Expr a -> State BuilderState (Expr a)
-findKeccakExpr e = mapExprM go e
+findKeccakExpr :: forall a. Expr a -> State KeccakStore (Expr a)
+findKeccakExpr e = mapExprM keccakFinder e
 
-findKeccakProp :: Prop -> State BuilderState Prop
-findKeccakProp p = mapPropM go p
+findKeccakProp :: Prop -> State KeccakStore Prop
+findKeccakProp p = mapPropM keccakFinder p
 
-findKeccakPropsExprs :: [Prop] -> [Expr Buf]  -> [Expr Storage]-> State BuilderState ()
+findKeccakPropsExprs :: [Prop] -> [Expr Buf]  -> [Expr Storage]-> State KeccakStore ()
 findKeccakPropsExprs ps bufs stores = do
-  mapM_ findKeccakProp ps;
-  mapM_ findKeccakExpr bufs;
+  mapM_ findKeccakProp ps
+  mapM_ findKeccakExpr bufs
   mapM_ findKeccakExpr stores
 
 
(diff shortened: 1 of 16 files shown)


test/test.hs
```diff
@@ -411,13 +411,10 @@ tests = testGroup "hevm"
...
         checkEquivProp simplified p
-    -- This would need to be a fuzz test I think. The SMT encoding of Keccak is not precise
-    -- enough for this to succeed
-    , ignoreTest $ testProperty "storage-slot-simp-property" $ \(StorageExp s) -> ioProperty $ do
-        T.writeFile "unsimplified.expr" $ formatExpr s
-        let simplified = Expr.simplify s
-        T.writeFile "simplified.expr" $ formatExpr simplified
-        checkEquiv simplified s
+    , testProperty "storage-slot-simp-property" $ \(StorageExp s) -> ioProperty $ do
+        let s2 = Expr.structureArraySlots s
```
msooseth, line 415: Note: this is needed, otherwise, the tests fail due to inability for the system to reverse some Keccak values. While it could be fixed without this hack, it would involve writing a minimal simplifier, I believe. That would likely be more error-prone than trusting `structureArraySlots`, which I think is something we kinda have to trust anyway, and is quite straightforward.

msooseth:
Let's not merge this yet -- I wanna add to this the collection of Keccaks and the fix to add reversing for all keccak calls.

src/EVM/Keccak.hs
```diff
@@ -67,13 +72,20 @@ injProp _ = internalError "expected keccak expression"
...
+keccakAssumptions ps bufs stores = injectivity <> minValue <> minDiffOfPairs
   where
     (_, st) = runState (findKeccakPropsExprs ps bufs stores) initState
 
     injectivity = fmap injProp $ combine (Set.toList st.keccaks)
     minValue = fmap minProp (Set.toList st.keccaks)
-
+    minDiffOfPairs = map minDistance $ filter (uncurry (/=)) [(a,b) | a<-(Set.elems st.keccaks), b<-(Set.elems st.keccaks)]
+     where
+      minDistance :: ((Expr EWord), (Expr EWord)) -> Prop
```
d-xo, line 83:
```suggestion
      minDistance :: (Expr EWord, Expr EWord) -> Prop
```

test/test.hs
```diff
@@ -303,7 +303,7 @@ tests = testGroup "hevm"
   -- applying some simplification rules, and then using the smt encoding to
   -- check that the simplified version is semantically equivalent to the
   -- unsimplified one
-  , adjustOption (\(Test.Tasty.QuickCheck.QuickCheckTests n) -> Test.Tasty.QuickCheck.QuickCheckTests (min n 50)) $ testGroup "SimplifierTests"
+  , adjustOption (\(Test.Tasty.QuickCheck.QuickCheckTests n) -> Test.Tasty.QuickCheck.QuickCheckTests (min n 50000)) $ testGroup "SimplifierTests"
```
d-xo, line 306: did you mean to commit this?

test/test.hs
d-xo: I'm fine with this, but maybe just add a comment explaining what's going on?

src/EVM/Keccak.hs
```diff
@@ -50,8 +52,11 @@ combine lst = combine' lst []
       let xcomb = [ (x, y) | y <- xs] in
       combine' xs (xcomb:acc)
 
+maxW32 :: W256
+maxW32 = into (maxBound :: Word32)
+
 minProp :: Expr EWord -> Prop
-minProp k@(Keccak _) = PGT k (Lit 50)
+minProp k@(Keccak _) = PGT k (Lit maxW32)
```
d-xo, line 59: I think we only really need this to avoid potential collisions with value type storage slots? I agree that 50 might be a little small, but I would also probably prefer to keep our assumptions here as weak as we need to get usable results? What do you think about setting it to 256 instead (this seems to be safely enough for all real world solidity contracts?), or do you think that 2^32 is useful for other reasons?

msooseth: 256 is a good idea, that was my initial thought, too!

test/test.hs
msooseth: Good point, thanks, added an explanation :)

test/test.hs
msooseth: Ooops, no. Good catch. Sorry!

src/EVM/Keccak.hs
msooseth: Ok, it's 256 all around now :)

msooseth:
Ok, let's see if this can be merged independently of the new work I'm doing. I'll rebase so we can cut it into 2 :)

PR #388, opened 2023-09-27
EVM: fix concrete prefix detection during contract creation

d-xo:
## Description

This replaces `Expr.concPrefix` with a much simpler routine that:

1. converts the input buffer into a vector of bytes
2. scans through this vector until it finds the first non Lit byte
3. uses this to split the input buf into a concrete and symbolic part

While 1. has non trivial complexity, we were already doing it later anyway (over the concrete part only), so this hopefully doesn't impact perf too much.

There are still a few issues to work out:

- [ ] the routine as described does not support constructor args that have a fully abstract size (i.e. bytes[], string)
- [ ] it breaks some of the existing tests

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -428,34 +428,6 @@ bufLengthEnv env useEnv buf = go (Lit 0) buf
         Nothing -> internalError "cannot compute length of open expression"
     go l (GVar (BufVar a)) = EVM.Expr.max l (BufLength (GVar (BufVar a)))
 
--- | If a buffer has a concrete prefix, we return it's length here
-concPrefix :: Expr Buf -> Maybe W256
-concPrefix (WriteWord (Lit srcOff) _ (ConcreteBuf ""))
-  | srcOff == 0 = Nothing
-  | otherwise = Just srcOff
-concPrefix (CopySlice (Lit srcOff) (Lit _) (Lit _) src (ConcreteBuf "")) = do
-  sz <- go 0 src
-  pure . into $ (unsafeInto sz) - srcOff
-  where
-    go :: W256 -> Expr Buf -> Maybe Integer
-    -- base cases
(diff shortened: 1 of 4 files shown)


d-xo:
I think we will probably need to use `Vector.create` to produce the concrete prefix in reasonable time using this new approach, while also supporting dynamic constructor args (i.e. stop iterating over the input buffer as soon as we hit a non lit byte).

I guess fixing the currently failing tests will require improving `Expr.readByte`.

We should also probably expand our test suite coverage around symbolic constructor args in general since the old `concPrefix` was buggy in a lot of different ways that went undetected by the current test suite.

src/EVM/Expr.hs
```diff
@@ -478,6 +454,54 @@ minLength bufEnv = go 0
       b <- Map.lookup a bufEnv
       go l b
 
+-- returns the largest prefix that is guaranteed to be concrete (if one exists)
+concretePrefix :: Expr Buf -> Maybe (Vector (Expr Byte))
```
d-xo, line 458: this should just return `Maybe (Vector Byte)` since they should all be concrete. Even better would be a `ByteString`, but the fast creation routines for `ByteString` all seem to expect a concrete size in advance (which we don't always know).

src/EVM/Expr.hs
```diff
@@ -478,6 +454,54 @@ minLength bufEnv = go 0
...
+concretePrefix b = let v = mkvec in case V.length v of
+  0 -> Nothing
```
zoep, line 460: If the input is an empty concrete buffer the output will be `Nothing` and I'm not sure if this is the correct semantics (empty vector would make more sense). I don't know if we will ever call this with an empty buffer though since we use it with initcode.

zoep (approved):
Looks good!

PR #390, opened 2023-10-02
Expr: evalProp: a few small solc specific rewrites

d-xo:
## Description

Adds a few rewrites to `evalProp` that simplify solc generated code.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1003,7 +1003,21 @@ evalProp prop =
     go (PNeg (PNeg a)) = a
 
     -- solc specific stuff
+
+    -- iszero(a) -> (a == 0)
+    -- iszero(iszero(a))) -> ~(a == 0) -> a > 0
+    -- iszero(iszero(a)) == 0 -> ~~(a == 0) -> a == 0
+    -- ~(iszero(iszero(a)) == 0) -> ~~~(a == 0) -> ~(a == 0) -> a > 0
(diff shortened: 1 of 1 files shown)


src/EVM/Expr.hs
```diff
@@ -978,6 +978,11 @@ evalProp prop =
     go (PNeg (PBool b)) = PBool (Prelude.not b)
     go (PNeg (PNeg a)) = a
 
+    -- solc specific stuff
+    go (PNeg (PEq (IsZero (IsZero a)) (Lit 0))) = PGT a (Lit 0)
```
msooseth, line 982: We could instead add the rewrite:
```
PEq (IsZero (IsZero a)) (Lit 0) = IsZero a
```

Then `PNeg (PEq (IsZero (IsZero a)) (Lit 0))) ` would become `PNeg (IsZero a)` which we could then rewrite into `PGT a (Lit 0)` with the rule:

```
go (PNeg (IsZero a)) = `PGT a (Lit 0)`
```

That would be more generic I think?

msooseth (approved):
These are actually quite a mind-bending. I was afraid of the `Not` issue we had with `Expr`, but `PNeg` is not the same `Expr.Not`, it's kinda whacky. Anyway, nice! I have only one observation. You can merge, but I think maybe there is a slightly more generic way for one of the rewrites.

PR #393, opened 2023-10-02
SymExec: formatCex decodes calldata if an abi is known

d-xo:
## Description

`formatCex` now takes an optional `Sig`. If this arg is present, then it will decode the raw `ByteString` calldat model into a method call with abi decoded values.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
Nice!

PR #397, opened 2023-10-02
Keccak: update min value assumption to match `readStorage`

d-xo:
## Description

Since https://github.com/ethereum/hevm/pull/367 we will strip writes to a concrete slot where the index is < 256. This PR updates the min value assumptions in Keccak.hs to match.

Fixes https://github.com/ethereum/hevm/issues/394\

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
lgtm

PR #400, opened 2023-10-02
Test: don't generate same name for different types

d-xo:
## Description

We were hitting failures in the prop simplifier fuzz tests due to error messages like the following from the smt solver:

```
Error "error while writing SMT to solver: Solver returned an error:\n(error \"line 86 column 74: ambiguous constant reference, more than one constant with the same sort, use a qualified expression (as <symbol> <sort>) to disambiguate esc_Q\")\nwhile sending the following line: (assert (=> (bvuge (_ bv0 256) (max (_ bv0 256) esc_Q_length)) (= (select esc_Q (_ bv0 256)) (_ bv0 8))))"
```

This was because we could generate the same name for different types (e.g. buffer and word). This changes ensures that we never generate the same name for expressions with different types. We could fix this at the smt layer (as directed by the error message), but I feel like generating this kind of smt during normal execution would be indicative of some deeper problem, so prefer to surface the error if it occurs.

Fixes https://github.com/ethereum/hevm/issues/396

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -3194,7 +3194,7 @@ instance Arbitrary Addr where
 instance Arbitrary (Expr EAddr) where
   arbitrary = oneof
     [ fmap LitAddr arbitrary
-    , fmap SymAddr genName
+    , fmap SymAddr (genName "addr")
     ]
 
 instance Arbitrary (Expr Storage) where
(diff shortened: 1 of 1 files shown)


msooseth (approved):
lgtm

PR #402, opened 2023-10-03
Avoid reading out of buffer limits in `readWordFromBytes`

### LINKED ISSUE #401: `ReadWord` equivalence bug
The result of simplifying the following expression is reported as non-equivalent by the smt solver:

```
(ReadWord (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff) (ConcreteBuf "\163\179s\141\183\197fx\128d'Z\ETX\173\b\ESC`\v\155\141\DC3\SO\DLE\211,\237l\255g\132D\223\&6\218"))
```

To reproduce add the following as a top level function to test.hs and call it from the repl:

```haskell
repro = do
  let p = (ReadWord (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff) (ConcreteBuf "\163\179s\141\183\197fx\128d'Z\ETX\173\b\ESC`\v\155\141\DC3\SO\DLE\211,\237l\255g\132D\223\&6\218"))
  let simplified = Expr.simplify p
  print simplified
  checkEquiv simplified p
```

zoep:
## Description

It adds a check that avoids trying to read a word from a concrete buffer that goes beyond its address space. 

Closes #401.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -279,12 +279,17 @@ readWord i b = readWordFromBytes i b
 -- returns an abstract ReadWord expression if a concrete word cannot be constructed
 readWordFromBytes :: Expr EWord -> Expr Buf -> Expr EWord
 readWordFromBytes (Lit idx) (ConcreteBuf bs) =
-  -- the word we are trying to read is within the address space of the buffer
   if idx <= (maxBound :: W256) - 31
+  -- the word we are trying to read is within the address space of the buffer
   then case toInt idx of
          Nothing -> Lit 0
          Just i -> Lit $ word $ padRight 32 $ BS.take 32 $ BS.drop i bs
-  else ReadWord (Lit idx) (ConcreteBuf bs)
+  -- if the word we are trying to read goes outside of bounds, then
+  -- wraparound and take the remaing bytes from the beginning of the
+  -- buffer
+  else case toInt ((maxBound :: W256) - idx + 1) of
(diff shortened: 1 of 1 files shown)


d-xo:
As discussed in chat: our smt encoding does not match geth (or even our own concrete semantics), if the index being read from memory is > 2^64. In this situation geth always returns zero (matching the behaviour of `readWord` on `main`, but our smt semantics will read data from any index.

The easiest short term fix is probably to constrain BufLength <= 2^64 for all buffers, and update readWord to return 0 for indexes greater than 2^64. longer term we might benefit from encoding bufs as an array with a 64 but address space.

msooseth (approved):
lgtm

msooseth:
>our smt encoding does not match geth (or even our own concrete semantics) if the index being read from memory is > 2^64. In this situation geth always returns zero

While that's true, geth's behaviour is basically undefined for such large idx. The behaviour mentioned above is guarded by a check that the idx is <2^64. When asked, the geth developers said that it doesn't deal with that large indexes, that's gas-out anyway. So I don't know if we should rely on geth's behaviour here. 

This is related to https://github.com/ethereum/hevm/issues/401 where this is explained in a bit more detail.

msooseth:
This has finally been experimented with and partially fixed, with the fuzzer fix as the last resort. Closing.

PR #403, opened 2023-10-03
Abstract gas v2

d-xo:
## Description

This is a second attempt at implementing a partially abstract gas model, based on top of some experiments from @arcz. Instead of modelling gas as a `Maybe Word64` (as in https://github.com/ethereum/hevm/pull/352), here we add a `gas` parameter to the `VM` type and constrain it to be an instance of the `Gas` typeclass.

Two instances of this class are provided:

1. `Word64`: this exactly matches the previous concrete gas semantics
2. `SymGas`: this is a single constructor type and simply acts as a marker that we should ignore all gas accounting

This (combined with a bunch of specilialise and inline pragmas) should keep the overhead for concrete execution relatively low. There are still four failing ethereum tests that need to be fixed before I can confirm this hypothesis for sure, but initial tests by @arcz on an earlier commit seemed promising.

Compared to https://github.com/ethereum/hevm/pull/352 this is maybe not quite as clean (our type parameters continue to proliferate), and it does allow more invalid vm states (e.g. as implemented here the combination of concrete gas with symbolic VM state will output garbage numbers for gas), but it's probably worth it if it minimizes the impact to concrete perf.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -374,13 +374,13 @@ writeByte offset byte src = WriteByte offset byte src
 
 writeWord :: Expr EWord -> Expr EWord -> Expr Buf -> Expr Buf
 writeWord o@(Lit offset) (WAddr (LitAddr val)) b@(ConcreteBuf _)
-  | offset + 32 < maxBytes
+  | offset < maxBytes
   = writeWord o (Lit $ into val) b
 writeWord (Lit offset) (Lit val) (ConcreteBuf "")
-  | offset + 32 < maxBytes
+  | offset < maxBytes
   = ConcreteBuf $ BS.replicate (unsafeInto offset) 0 <> word256Bytes val
 writeWord o@(Lit offset) v@(Lit val) buf@(ConcreteBuf src)
-  | offset + 32 < maxBytes
+  | offset < maxBytes
     = ConcreteBuf $ (padRight (unsafeInto offset) $ BS.take (unsafeInto offset) src)
                  <> word256Bytes val
                  <> BS.drop ((unsafeInto offset) + 32) src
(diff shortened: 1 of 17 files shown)


d-xo:
All tests pass now, but perf unfortunately seems to be very similar to the V1 attempt, running `cabal run bench -- -p "blockchain-tests"` gives:

**main:** 33.712 s ± 199 ms
**here:** 46.897 s ± 462 ms

src/EVM/Expr.hs
```diff
@@ -374,13 +374,13 @@ writeByte offset byte src = WriteByte offset byte src
 
 writeWord :: Expr EWord -> Expr EWord -> Expr Buf -> Expr Buf
 writeWord o@(Lit offset) (WAddr (LitAddr val)) b@(ConcreteBuf _)
-  | offset + 32 < maxBytes
+  | offset < maxBytes
```
msooseth, line 377: Is this an accidental change?

src/EVM/Types.hs
```diff
@@ -681,7 +681,7 @@ data FrameState s = FrameState
   , calldata     :: Expr Buf
   , callvalue    :: Expr EWord
   , caller       :: Expr EAddr
-  , gas          :: {-# UNPACK #-} !Word64
+  , gas          :: !gas
```
msooseth, line 684: The change above had

```
{-# UNPACK #-} !gas
```

Wouldn't that be faster?

d-xo:
added the magic compiler options that @arcz uncovered in https://github.com/ethereum/hevm/pull/427 and now this benchmarks exactly the same as `main`!

*main*: 37.372 s ± 1.63 s
*here*: 36.147 s ± 1.55 s

msooseth:
Wow! Nice.

msooseth:
Yay, Abstract Gasv3 has been merged!! See: https://github.com/ethereum/hevm/pull/427 So I am closing this :)

PR #404, opened 2023-10-06
remove gas

d-xo:
## Description

This pr entirely removes gas accounting from EVM.hs. It significantly simplifies the logic in EVM.hs, and removes a non trivial amount of code. The major downside here is that we can no longer:

1. execute the ethereum test suite
2. fuzz differentially against geth

I'm inclined to merge #352 or #403 instead of this since they allow us to run without gas while still letting us test vs geth. If we do decide to delete gas entirely, we would ideally generate a version of the evm conformance test suite that would represent a gasless semantics. 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
This change has been merged already as part of another amazing PR, thanks for that! Closing this as this is a duplicate :)

PR #405, opened 2023-10-09
Add configuration in a monadic way, for now used only to configure SMT2 debugging

msooseth:
## Description
This is a rather big change. It adds a monad that allows us to pass a global configuration down. This can be extremely helpful, e.g. also for abstraction-refinement, Expr debugging, and a bunch of other things. It required a bit of change here-and-there, but it's not _that_ bad. It will allow us to pass a lot less parameters around for global configs.

This also adds dumping of SMT2 queries. This is done by adding one more field to the SMT2 datatype, the `Expr` that it was generated from. This is essential, otherwise we don't know from that place the SMT2 is coming from. It helps significantly with debugging the SMT2. Passing this `Expr` in any other way is not really possible, not in a clean way as far as I can see. It really does belong with the SMT2, I think.

Configurations supported:
* SMT, Expression, Prop dumping
* abstraction-refinement (revert 3b05c8509fa0a1dd36530cc0060bdb346d27644f in case it's undesired)

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Note, as part of the Keccak PR I am making, I am moving more things to the `Config`. So let's hold off a bit on adding more things to the Config on this PR. I promise to extend `Config` with more options as part of the `Keccak` PR here: https://github.com/ethereum/hevm/pull/382

cli/cli.hs
```diff
@@ -139,11 +143,12 @@ data Command w
       , projectType   :: w ::: Maybe ProjectType        <?> "Is this a Foundry or DappTools project (default: Foundry)"
       , rpc           :: w ::: Maybe URL                <?> "Fetch state from a remote node"
       , number        :: w ::: Maybe W256               <?> "Block: number"
-      , verbose       :: w ::: Maybe Int                <?> "Append call trace: {1} failures {2} all"
       , coverage      :: w ::: Bool                     <?> "Coverage analysis"
       , match         :: w ::: Maybe String             <?> "Test case filter - only run methods matching regex"
       , solver        :: w ::: Maybe Text               <?> "Used SMT solver: z3 (default) or cvc5"
       , smtdebug      :: w ::: Bool                     <?> "Print smt queries sent to the solver"
+      , verbose       :: w ::: Maybe Int                <?> "Verbose printing of internal behaviour (default:0)"
```
d-xo, line 150: wdyt about leaving the old `verbose` alone and calling this option `debug` instead?

src/EVM/Fetch.hs
```diff
@@ -224,27 +227,28 @@ oracle solvers info q = do
...
         Nothing -> pure (continue 0)
         Just (n, url) ->
-         fetchSlotFrom n url addr slot >>= \case
+         liftIO $ fetchSlotFrom n url addr slot >>= \case
            Just x  -> pure (continue x)
            Nothing ->
              internalError $ "oracle error: " ++ show q
 
-type Fetcher s = Query s -> IO (EVM s ())
+type Fetcher m s = (ReadConfig m, MonadUnliftIO m) => Query s -> m (EVM s ())
```
d-xo, line 235:
```suggestion
type Fetcher m s = App m => Query s -> m (EVM s ())
```

d-xo (approved):
just a few small nits, but otherwise looks great!

cc @arcz this is a breaking change that may impact echidna.

cli/cli.hs
msooseth: OK, I have fixed this now :)

PR #406, opened 2023-10-10
Keccak improvements

msooseth:
## Description
This change pushes the Keccak concretization to the very END of the steps we do. We don't concretize keccak (i.e. execute SHA3) during interpretation unless necessary (e.g. to do an RPC call). This allows us to take advantage of the fancy pattern matching in `readStorage` more often, hence making it more effective. In particular, this for example passes the  new `transfer-dapp` test instantly that previously Z3 struggled with.

This improvement required quite a bit of change in the way we handle Keccak. The most hairy parts are:
- We now have a `symbolic` value in our `VMOpts` to know if we are doing concrete or symbolic execution. Concrete exe
- `Keccak.hs` now requires a bit more care, as we need to do some concretization as well
- we now have `concKeccakSimp True/False` to BOTH concretize and simplify (or not) the expression. This is basically a beefed-up `simplify`. This is sometimes needed, rather than pure `simplify`.
- in `EVM.hs` the `accessStorage` function had to be changed slightly so we can check whether we get a `Nothing` either with or without concretization. If we get a Nothing with either of the two, we do an rpc call. This is necessary, or we would never look up the rpc.

I also added a more thorough `Add`+`Sub` simplification. This is more complete than what we had. It was needed for one of the debugging sessions I was doing and so while it's unrelated, I think it should be OK(?) Let me know if you need it separately :)

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Changed
+- Less noisy console output during tracing
+- Minimum distance requirements are now asserted for Keccak function calls. They assert that it's hard to generate two Keccak's that are less than 256 afar.
+- Keccak concretization is now done only after all simplification are performed. This helps with simplification pre-concretization
```
zoep, line 11: small typo
```suggestion
- Keccak concretization is now done only after all simplifications are performed. This helps with simplification pre-concretization
```

test/EVM/Test/Tracing.hs
```diff
@@ -415,7 +415,10 @@ deleteTraceOutputFiles evmtoolResult =
 
 -- Create symbolic VM from concrete VM
 symbolify :: VM s -> VM s
-symbolify vm = vm { state = vm.state { calldata = AbstractBuf "calldata" } }
+symbolify vm = vm {
+                  state = vm.state { calldata = AbstractBuf "calldata" }
+                  , symbolic = True
+                  }
```
zoep, line 421: indentation
```suggestion
symbolify vm = vm { state = vm.state { calldata = AbstractBuf "calldata" }
                                  , symbolic = True
                                  }
```

msooseth: Ayyy good catch! Fixed! :)

src/EVM/SMT.hs
```diff
@@ -249,25 +250,35 @@ assertPropsNoSimp config ps =
...
-    addresses = Set.toList $ Set.unions (fmap referencedWAddrs ps)
-
-    keccakAssumes
+    storageReads = Map.unionsWith (<>) $ fmap findStorageReads toDeclarePs
+    abstractStores = Set.toList $ Set.unions (fmap referencedAbstractStores toDeclarePs)
+    addresses = Set.toList $ Set.unions (fmap referencedWAddrs toDeclarePs)
+
+    -- Keccak assertions: concrete values, distance between pairs, injectivity, etc.
+    keccAssump = keccakAssumptions psPreConc bufVals storeVals
+    keccComp = keccakCompute psPreConc bufVals storeVals
```
zoep, line 274: It think it is worth adding a comment here that this pass will add equations about the concrete values of keccak expressions

msooseth: Good point. Let me do that!

src/EVM/SMT.hs
```diff
@@ -249,25 +250,35 @@ assertPropsNoSimp config ps =
...
-    storageReads = Map.unionsWith (<>) $ fmap findStorageReads ps
-    abstractStores = Set.toList $ Set.unions (fmap referencedAbstractStores ps)
-    addresses = Set.toList $ Set.unions (fmap referencedWAddrs ps)
-
-    keccakAssumes
+    storageReads = Map.unionsWith (<>) $ fmap findStorageReads toDeclarePs
+    abstractStores = Set.toList $ Set.unions (fmap referencedAbstractStores toDeclarePs)
+    addresses = Set.toList $ Set.unions (fmap referencedWAddrs toDeclarePs)
+
+    -- Keccak assertions: concrete values, distance between pairs, injectivity, etc.
```
msooseth, line 272:
```suggestion
    -- Keccak assertions: concrete values, distance between pairs, injectivity, etc.
    --      This will make sure concrete values of Keccak are asserted, if they can be computed (i.e. can be concretized)
```

src/EVM.hs
```diff
@@ -1306,29 +1304,33 @@ accessStorage
   -> (Expr EWord -> EVM s ())
   -> EVM s ()
 accessStorage addr slot continue = do
+  let slotConc = Expr.concKeccakSimpExpr True slot
   use (#env % #contracts % at addr) >>= \case
     Just c ->
+      -- Try first without concretization. Then if we get a Just, with concretization
+      -- if both give a Just, should we `continue`
```
d-xo, line 1311: this does not spark joy, but I think we need a larger refactor of the rpc stuff to handle this properly (out of scope for this pr imo).

msooseth: Good point. I have now explained this in more detail!

msooseth:
OK, I think I managed to improve this a bit :)

src/EVM/Expr.hs
```diff
@@ -1350,35 +1350,20 @@ constFoldProp ps = oneRun ps (ConstState mempty True)
...
-                            else concKeccakSimp simp conc
-
-concKeccakSimpOnePass :: Expr a -> Expr a
-concKeccakSimpOnePass (Keccak (ConcreteBuf bs)) = Lit (keccak' bs)
-concKeccakSimpOnePass orig@(Keccak (CopySlice (Lit 0) (Lit 0) (Lit 64) orig2@(WriteWord (Lit 0) _ (ConcreteBuf bs)) (ConcreteBuf ""))) =
+-- Concretize & simplify Keccak expressions until fixed-point.
+concKeccakSimpExpr :: Expr a -> Expr a
+concKeccakSimpExpr orig = untilFixpoint ((mapExpr concKeccakOnePass) . simplify) orig
+
+concKeccakProps :: [Prop] -> [Prop]
```
d-xo, line 1360:
```suggestion
concKeccakSimpProps :: [Prop] -> [Prop]
```

just for consistency...

d-xo (approved):
just one more small nit, but apart from that lgtm :)

src/EVM/Expr.hs
msooseth: Oh wow, this was a bug actually. There shouldn't have been a `simplifyProps $` in front. So the name was correct, and there ought to be no `Simp` there. I now removed the `simplifyProps $` and the name is now correct :)

PR #414, opened 2023-10-19
Simplifying BufLength via bufLength

msooseth:
## Description
We accidentally removed(?) or never added(?) the `bufLength` call to simplify `BufLength`. Discovered this while writing a concrete fuzzer of Expr. This is a pretty funny one I think! :D Added a unit test to make sure we catch this next time!

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -282,7 +282,12 @@ tests = testGroup "hevm"
     ]
   , testGroup "SimplifierUnitTests"
     -- common overflow cases that the simplifier was getting wrong
-    [ testCase "writeWord-overflow" $ do
+    [ testCase "bufLength-simp" $ do
+      let
+        a = BufLength (ConcreteBuf "ab")
+        simp = Expr.simplify a
+      assertEqual "Must be simplified down to a Lit" simp (Lit 2)
+    , testCase "writeWord-overflow" $ do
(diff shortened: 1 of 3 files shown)


test/test.hs
```diff
@@ -282,7 +282,12 @@ tests = testGroup "hevm"
     ]
   , testGroup "SimplifierUnitTests"
     -- common overflow cases that the simplifier was getting wrong
-    [ testCase "writeWord-overflow" $ do
+    [ testCase "bufLength-simp" $ do
```
zoep, line 285: Would adding tests for more complicated buffer expressions (the involve writes perhaps to abstract buffers) make sense?

zoep (approved):
Nice catch!

I think it would be nice to add more tests for `bufLength`, but maybe it is also testsed by the fuzzer?

test/test.hs
msooseth: Probably :) I wanted to keep this PR short and sweet. I didn't want to dig myself into `bufLength` -- it may be perfect, it may be possible to improve. I just wanted to make sure it's actually working and used :smile: 

Should we write a few more test cases? Perhaps that could be an issue I can tackle in another PR. I don't want to delay this PR too long. What do you think?

msooseth:
I'll add some more tests for other functionalities of `bufLength` in another PR :)

PR #415, opened 2023-10-20
Move Windows build to GHC 9.4

elopez:
## Description

This PR introduces several changes to the Windows CI in order to build with GHC 9.4

GHC 9.4 currently breaks on Windows because it now uses a Clang-based toolchain and the C++ standard library name does no longer match:

  * https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4#windows-uses-a-clang-based-toolchain
  * https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4#link-against-system-cxx-std-lib-instead-of-stdc

Additionally, GHC 9.4 now requires the CLANG64 environment in MSYS2 for native code linking:

  * https://gitlab.haskell.org/ghc/ghc/-/issues/22561

GHC 9.4 introduces a generic way of linking against the system C++ standard library, `build-depends: system-cxx-std-lib`:

  * https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4#link-against-system-cxx-std-lib-instead-of-stdc

This can be used to reliably link against the system's C++ standard library implementation without having to use platform-dependent conditionals. Unfortunately it does not properly work on Nix, so we link manually on Darwin to work around the issue:

  * https://gitlab.haskell.org/ghc/ghc/-/issues/23138

The GitHub runner also has many things preinstalled that tend to cause conflicts during building. This PR switches it to use the minimal MSYS2 path option and adds-in the Haskell tools we need instead.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -58,6 +58,7 @@ common shared
     -Wno-unticked-promoted-constructors
     -Wno-orphans
     -Wno-ambiguous-fields
+    -optc-Wno-ignored-attributes
   default-language: GHC2021
   default-extensions:
     DuplicateRecordFields
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -110,7 +110,7 @@ library
   elif !os(darwin)
     extra-libraries: stdc++
   else
-    extra-libraries: c++
+    -- extra-libraries: c++
     -- https://gitlab.haskell.org/ghc/ghc/-/issues/11829
     ld-options: -Wl,-keep_dwarf_unwind
     ghc-options: -fcompact-unwind
@@ -228,6 +228,8 @@ executable hevm
     optics-core,
     githash                       >= 0.1.6 && < 0.2,
     witch
+  if os(darwin)
+    extra-libraries: c++
   if os(windows)
     buildable: False
 
@@ -237,8 +239,6 @@ common test-base
   import: shared
   hs-source-dirs:
     test
-  extra-libraries:
-    secp256k1
   other-modules:
     Paths_hevm
   autogen-modules:
@@ -303,6 +303,8 @@ common test-common
     EVM.Test.Utils
     EVM.Test.Tracing
     EVM.Test.BlockchainTests
+  if os(darwin)
+    extra-libraries: c++
   if os(windows)
     buildable: False
 
@@ -363,3 +365,5 @@ benchmark bench
     filepath,
     containers,
     mtl
+  if os(darwin)
+    extra-libraries: c++
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -195,6 +195,8 @@ executable hevm
   ghc-options: -threaded -with-rtsopts=-N
   other-modules:
     Paths_hevm
+  if os(darwin)
+    extra-libraries: c++
   build-depends:
     QuickCheck,
     aeson,
@@ -229,8 +231,6 @@ executable hevm
     optics-core,
     githash                       >= 0.1.6 && < 0.2,
     witch
-  if os(darwin)
-    extra-libraries: c++
   if os(windows)
     buildable: False
 
@@ -304,10 +304,10 @@ common test-common
     EVM.Test.Utils
     EVM.Test.Tracing
     EVM.Test.BlockchainTests
-  if os(darwin)
-    extra-libraries: c++
   if os(windows)
     buildable: False
+  if os(darwin)
+    extra-libraries: c++
 
 --- Test Suites ---
 
@@ -349,6 +349,8 @@ benchmark bench
     bench
   ghc-options:
     -O2
+  if os(darwin)
+    extra-libraries: c++
   other-modules:
     Paths_hevm
   autogen-modules:
@@ -366,5 +368,3 @@ benchmark bench
     filepath,
     containers,
     mtl
-  if os(darwin)
-    extra-libraries: c++
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -103,8 +103,17 @@ library
     Paths_hevm
   autogen-modules:
     Paths_hevm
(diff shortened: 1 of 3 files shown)


hevm.cabal
```diff
@@ -371,3 +366,5 @@ benchmark bench
     filepath,
     containers,
     mtl
+  if os(darwin)
+    extra-libraries: c++
```
msooseth, line 370: Can we move this up so the diff looks smaller?

hevm.cabal
```diff
@@ -300,15 +304,10 @@ common test-common
...
-    buildable: False
   if os(darwin)
     extra-libraries: c++
-    -- https://gitlab.haskell.org/ghc/ghc/-/issues/11829
-    ld-options: -Wl,-keep_dwarf_unwind
-    ghc-options: -fcompact-unwind
-  else
-    extra-libraries: stdc++
+  if os(windows)
+    buildable: False
```
msooseth, line 310: Can we move this up so the diff looks smaller? Thanks!

hevm.cabal
```diff
@@ -225,6 +229,8 @@ executable hevm
     optics-core,
     githash                       >= 0.1.6 && < 0.2,
     witch
+  if os(darwin)
+    extra-libraries: c++
```
msooseth, line 233: Can we move this up so the diff looks smaller?

msooseth (commented):
Looks nice! I'm only worried about the MacOS build, it can be finicky. Do you know if this successfully compiles & builds a static, running binary on MacOS, especially on Arm?

hevm.cabal
```diff
@@ -103,8 +104,17 @@ library
     Paths_hevm
   autogen-modules:
     Paths_hevm
-  if os(linux) || os(windows)
+  if impl(ghc >= 9.4) && !os(darwin)
```
d-xo, line 107: not sure we need the ghc version condition here. I'm fine with only supporting 9.4+ unless you need older versions for echidna?

hevm.cabal
```diff
@@ -103,8 +104,17 @@ library
...
+    -- darwin is skipped because it produces this error when building
+    -- > ghc: loadArchive: Neither an archive, nor a fat archive: `/nix/store/l3lkdfm7sg1wwc850451cikqds766h15-clang-wrapper-11.1.0/bin/clang++'
```
d-xo, line 109: I've seen this error before, iirc I was able to fix by adding the config to the executables instead of the library (that's why we have `extra-libaries: c++` duplicated a few times). Maybe that would work too for the `system-cxx-std-lib` stuff too?

elopez:
Hi @msooseth! I tried building a static binary as follows on an M1 Mac and it built fine -- although the binary did not run, as it was not signed. I added an extra commit to implement signing now, and the resulting binary appears to execute fine at first glance. I don't have an Intel Mac to verify if it works over there.

```shell
hevm % nix  build .#redistributable --out-link hevmMacos
hevm % otool -L ./hevmMacos/bin/hevm
./hevmMacos/bin/hevm:
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.60.1)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
        /usr/lib/libiconv.dylib (compatibility version 7.0.0, current version 7.0.0)
hevm % ./hevmMacos/bin/hevm version
0.51.3 [no git revision present]
```

I also moved those blocks around to reduce the diff noise as you requested.

hevm.cabal
elopez: We actually [just moved echidna to 9.4](https://github.com/crytic/echidna/pull/1096) 😄 so if you'd prefer to only support 9.4+ that's fine by me. It's not a lot of code to keep the 9.2 build-ability though (just the `else` block after this bit)

hevm.cabal
elopez: Unfortunately it does not 😞 if I apply this example diff on top of this PR, it stops building correctly. That's why I opted to generalize it for `!darwin` with the new `build-depends: system-cxx-std-lib` but keep the duplicated `extra-libraries: c++` in the executables for `darwin`

```diff
diff --git a/hevm.cabal b/hevm.cabal
index 2ebc62b6..b66c901d 100644
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -196,7 +196,7 @@ executable hevm
   other-modules:
     Paths_hevm
   if os(darwin)
-    extra-libraries: c++
+    build-depends: system-cxx-std-lib
   build-depends:
     QuickCheck,
     aeson,
```

```
       > [1 of 2] Compiling Paths_hevm       ( dist/build/hevm/autogen/Paths_hevm.hs, dist/build/hevm/hevm-tmp/Paths_hevm.o, dist/build/hevm/hevm-tmp/Paths_hevm.dyn_o )
       > [2 of 2] Compiling Main             ( cli/cli.hs, dist/build/hevm/hevm-tmp/Main.o, dist/build/hevm/hevm-tmp/Main.dyn_o )
       > ghc: loadArchive: Neither an archive, nor a fat archive: `/nix/store/l3lkdfm7sg1wwc850451cikqds766h15-clang-wrapper-11.1.0/bin/clang++'
       >
       > <no location info>: error:
       >     loadArchive "/nix/store/l3lkdfm7sg1wwc850451cikqds766h15-clang-wrapper-11.1.0/bin/clang++": failed
       > /nix/store/1k85b7faqm8pc0k3y553zr8bfhb2dv2z-ghc-9.4.6/bin/ghc returned
       > ExitFailure 1
```

d-xo (approved):
Oh the codesigning change is awesome, thanks so much. This looks great in general.

PR #416, opened 2023-10-24
Fuzzing Expr

msooseth:
## Description
This is a early-stage concrete fuzzer for `Expr`. It can, however, be quite useful in case the `Expr` is easy to satisfy. Added is an example use-case, `fuzz-complicated-mul` that takes forever for CVC5/Z3 to satisfy, but is instantaneous with this fuzzer. 

There are a number limitations that this fuzzer currently has. In particular, I would prefer if this was not just a fuzzer, but a local search system. For that it would (1) need to calculate the distance it's from a target solution and (2) need to calculate a vector to get closer to the target.

The system is using the generator from QuickCheck, which allows for e.g. `frequency` and other helper functions, and does not use `IO`, instead uses a deterministic RNG. Note that some tests were incorrect in `test.hs`, they have been fixed.

**NEW** Apparently, we forgot to call `slt` completely, so we never did constant replacement for `SLT` or `SGT`. This has now been fixed. I had to fix it as part of this PR, because my tests didn't pass without it. Oops.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Fuzz.hs
+++ b/src/EVM/Fuzz.hs
@@ -26,8 +26,9 @@ import Test.QuickCheck.Random (mkQCGen)
 
 -- TODO: Extract Var X = Lit Z, and set it
 tryCexFuzz :: [Prop] -> Integer -> (Maybe (SMT.SMTCex))
-tryCexFuzz ps tries = unGen (testVals tries) (mkQCGen 0) 1337
+tryCexFuzz prePs tries = unGen (testVals tries) (mkQCGen 0) 1337
   where
+    ps = Expr.simplifyProps $ Expr.concKeccakProps prePs
     vars = extractVars ps
     bufs = extractBufs ps
     stores = extractStorage ps
@@ -38,8 +39,8 @@ tryCexFuzz ps tries = unGen (testVals tries) (mkQCGen 0) 1337
       bufVals <- getBufs bufs
       storeVals <- getStores stores
       let
-        ret =  filterCorrectKeccak $ map (substituteEWord varvals . substituteBuf bufVals . substituteStores storeVals) ps
-        retSimp =  Expr.simplifyProps ret
+        ret = filterCorrectKeccak $ map (substituteEWord varvals . substituteBuf bufVals . substituteStores storeVals) ps
+        retSimp = Expr.simplifyProps $ Expr.concKeccakProps ret
       if null retSimp then pure $ Just (SMT.SMTCex {
                                     vars = varvals
                                     , addrs = mempty
@@ -103,7 +104,7 @@ findVarProp p = mapPropM go p
     go = \case
       e@(Var a) -> do
         s <- get
-        put $ s {vars = Set.insert (Var a) s.vars}
+        put s {vars = Set.insert (Var a) s.vars}
         pure e
       e@(Lit a) -> do
         s <- get
@@ -135,7 +136,7 @@ findBufProp p = mapPropM go p
     go = \case
       e@(AbstractBuf a) -> do
         s <- get
-        put $ s{bufs=Set.insert (AbstractBuf a) s.bufs}
+        put s {bufs=Set.insert (AbstractBuf a) s.bufs}
         pure e
       e -> pure e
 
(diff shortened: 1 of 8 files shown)


src/EVM/Fuzz.hs
```diff
@@ -0,0 +1,238 @@
...
+                            }
+
+findVarProp :: Prop -> State CollectVars Prop
+findVarProp p = mapPropM go p
+  where
+    go :: forall a. Expr a -> State CollectVars (Expr a)
+    go = \case
+      e@(Var a) -> do
+        s :: CollectVars <- get
+        put CollectVars {vars=Set.insert (Var a) s.vars, vals = s.vals}
```
zoep, line 106: Maybe use record update here:

```suggestion
        put (s { vars = Set.insert (Var a) s.vars })
```

src/EVM/Fuzz.hs
```diff
@@ -0,0 +1,238 @@
...
+extractStorage ps = execState (mapM_ findStorageProp ps) initStorageState
+
+findStorageProp :: Prop -> State CollectStorage Prop
+findStorageProp p = mapPropM go p
+  where
+    go :: forall a. Expr a -> State CollectStorage (Expr a)
+    go = \case
+      e@(AbstractStore a) -> do
+        s <- get
+        put $  CollectStorage {addrs=Set.insert a s.addrs, keys=s.keys, values=s.values}
```
zoep, line 162:
```suggestion
        put $  s { addrs=Set.insert a s.addrs }
```

src/EVM/Fuzz.hs
```diff
@@ -0,0 +1,238 @@
...
+findVarProp :: Prop -> State CollectVars Prop
+findVarProp p = mapPropM go p
+  where
+    go :: forall a. Expr a -> State CollectVars (Expr a)
+    go = \case
+      e@(Var a) -> do
+        s :: CollectVars <- get
+        put CollectVars {vars=Set.insert (Var a) s.vars, vals = s.vals}
+        pure e
+      e@(Lit a) -> do
```
zoep, line 108: Why are literals needed here?

src/EVM/Fuzz.hs
```diff
@@ -0,0 +1,238 @@
...
+    addTxStuff :: Map (Expr EWord) W256 -> Gen (Map (Expr EWord) W256)
+    addTxStuff a = do
+      val <- frequency [ (20, pure 0)
+                       , (1, getRndW256)
+                       ]
+      pure $ Map.insert TxValue val a
+    go :: [Expr EWord] -> Map (Expr EWord) W256 -> Gen (Map (Expr EWord) W256)
+    go [] valMap = pure valMap
+    go (a:ax) valMap = do
+      pickKnown :: Bool <- arbitrary
```
zoep, line 189: I understand now that this is where the known literals are used. But why is it helpful to pick a value from the set of known literals?

src/EVM/Fuzz.hs
```diff
@@ -0,0 +1,238 @@
...
+                                   ,(1, choose (11, 100))
+                                   ]
+      l <- replicateM numElems oneWrite
+      go ax (Map.insert addr (Map.fromList l) addrToValsMap)
+        where
+          oneWrite :: Gen (W256, W256)
+          oneWrite = do
+                    a <- getRndElem storesLoads.keys
+                    b <- frequency [(1, getRndW256)
+                                   ,(3, elements $ Set.toList storesLoads.values)
```
zoep, line 213: Why is it useful to use the values that had been inserted in stores here?

src/EVM/Fuzz.hs
msooseth: Yeah. I wanted to, but I can't. Try and it fails. No idea why. Haskell is mysterious. Maybe you can debug? In the meanwhile, I can't fix it, the error makes no sense, and what you wrote makes 100% sense but Haskell doesn't like it. I'd be happy if you could figure out why.... I really wanna know.

msooseth: It says:

```
src/EVM/Fuzz.hs:106:15: error:
    • Record update is ambiguous, and requires a type signature
    • In the second argument of ‘($)’, namely
        ‘s {vars = Set.insert (Var a) s.vars}’
      In a stmt of a 'do' block:
        put $ s {vars = Set.insert (Var a) s.vars}
      In the expression:
        do s :: CollectVars <- get
           put $ s {vars = Set.insert (Var a) s.vars}
           pure e
    |
106 |         put $ s {vars=Set.insert (Var a) s.vars}
```

src/EVM/Fuzz.hs
msooseth: See above. I just don't understand Haskell. Literally same exact error :cry:

src/EVM/Fuzz.hs
msooseth: What we do here is that we extract known literals and then later, we sometimes set values of variables to these known literals:

```
      pickKnown :: Bool <- arbitrary
      val <- if (not pickKnown) || (null vars.vals) then do getRndW256
                                                    else elements $ Set.toList (vars.vals)
```

So if there is a check like `if (a == 57) then ...` we will pick out `57` as a `Lit` and later try to set variables, e.g. `a` to 57. Basically, we are taking out magic constants from the `Expr`. Magic constant extraction is a known technique in this space. Here's Echidna fuzzer doing it:

https://github.com/crytic/echidna/blob/f52b35af4e646da3d9d86408d9ccda9134e2b9c7/lib/Echidna.hs#L74

and:

https://github.com/crytic/echidna/blob/f52b35af4e646da3d9d86408d9ccda9134e2b9c7/lib/Echidna/Processor.hs#L56

src/EVM/Fuzz.hs
msooseth: It allows us to hop over some tricky ITE-s, like `if (a == 57)` where it'd be hard to figure out 57 (especially out of 2**256). At least that's the idea :) This technique is used in most of fuzzers, and all fuzzers I know allow you to add such magic numbers if you know them in advance for your program. For example, if some parts of the parser may never start if it doesn't see the magic constant `0xfea3fadc`, then you can add that magic constant to the fuzzer and hopefully it'll manage to fuzz more parts of your code.

I check for this in this test:

```
    , test "fuzz-simple-fixed-value3" $ do
      Just c <- solcRuntime "MyContract"
        [i|
        contract MyContract {
          function func(uint a, uint b) public {
            assert(((a != 1337) && (b != 99)));
          }
        }
        |]
```

I hope this helps :)

zoep: oh that's really nice! Thanks for explaining!

msooseth: :)

src/EVM/Fuzz.hs
msooseth: Thanks for helping with this! :heart:

d-xo (approved):
looks great :)

PR #424, opened 2023-11-08
Fixing up BufLength, the use of nubOrd, and adding one more PLT rule

msooseth:
## Description
This is a small fix and 2 improvements. The small fix is that previously:

```haskell
Expr.simplify $ (Max (Lit 0) (BufLength (AbstractBuf "txdata")))
```
became:
```haskell
(Max (Lit 0) (BufLength (AbstractBuf "txdata")))
```

Effectively "ignoring" the rule:
```haskell
    go (Max a b) = case (a, b) of
                    (Lit 0, _) -> b
                    _ -> EVM.Expr.max a b
```

But in fact it was NOT ignoring that rule. Instead, 
```haskell
    go (BufLength buf) = bufLength buf
```

Was in fact creating an infinite loop(!) because of:

```haskell
 bufLengthEnv :: Map.Map Int (Expr Buf) -> Bool -> Expr Buf -> Expr EWord
 bufLengthEnv env useEnv buf = go (Lit 0) buf
   where
     go :: Expr EWord -> Expr Buf -> Expr EWord
     [...]
    go l (AbstractBuf b) = Max l (BufLength (AbstractBuf b))
```
Which added one `Max` every time -- which was then stripped by the `Max` rule. This lead to a "fixedpoint" and so things didn't get simplified. This has now been fixed.

We also added 2 more small rules:
```haskell
    go (PLT (Max (Lit a) b) (Lit c)) | a < c = PLT b (Lit c)
```

and:

```haskell
remRedundantProps p = nubOrd $ collapseFalse . filter (\x -> x /= PBool True) . nubOrd $ p
```

has been changed to have `nubOrd` twice as this way it actually does the work correctly, whereas otherwise it doesn't fully simplify (strangely...)

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


zoep (approved):
Nice catch!

d-xo (approved):

PR #426, opened 2023-11-09
Simplify earlier and don't check reachability for FALSE

msooseth:
## Description
We can do a bit more with `simplifyProps` and we can check for `[PBool False]` during reachability check, so we don't send a `[PBool False]` to the solver

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -567,14 +567,14 @@ verify solvers opts preState maybepost = do
       Nothing -> pure (expr, [Qed ()])
       Just post -> liftIO $ do
         let
-          -- Filter out any leaves that can be statically shown to be safe
-          canViolate = flip filter flattened $
-            \leaf -> case Expr.simplifyProp (post preState leaf) of
-              PBool True -> False
-              _ -> True
+          -- Filter out any leaves from `flattened` that can be statically shown to be safe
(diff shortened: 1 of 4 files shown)


src/EVM/SymExec.hs
```diff
@@ -567,14 +567,14 @@ verify solvers opts preState maybepost = do
...
-          canViolate = flip filter flattened $
-            \leaf -> case Expr.simplifyProp (post preState leaf) of
-              PBool True -> False
-              _ -> True
+          -- Filter out any leaves from `flattened` that can be statically shown to be safe
+          tocheck = flip map flattened $ \leaf -> (toPropsFinal leaf, leaf)
+            where
+              toProps leaf = PNeg (post preState leaf) : assumes <> extractProps leaf
+              toPropsFinal leaf = if opts.simp then Expr.simplifyProps $ toProps leaf
+                                               else toProps leaf
```
d-xo, line 575: Can we put this in the existing `where` block? I find multiple sets of where blocks kind hard to parse

msooseth: Good point! I fixed it now :)

d-xo (approved):
lgtm :)

PR #427, opened 2023-11-17
Abstract gas v3

arcz:
## Description
Another try for abstract gas (see https://github.com/ethereum/hevm/pull/403, https://github.com/ethereum/hevm/pull/404). 

I introduced two types of VMs that we can specialize execution for: `VM Symbolic` and `VM Concrete`. The behavior can be specialized in instances of `VMOps` class and data structures with type families similar to `Gas (t :: VMType)` introduced here. Thanks to `-fspecialize-aggressively -fexpose-all-unfoldings` this version is only slightly (~2%) slower than `main` as all the `VMOps`-dependent functions should be picked at compile time. This opens a whole world of possibilities for optimizing the concrete version of the VM without introducing too many changes.

For instance, having `branch` specialized to `branch (Lit cond) continue = continue (cond > 0)` in concrete VM speeds up ethereum-tests over 10% which puts way ahead compared to the current main. I didn't include further optimizations to limit the scope of this PR to gas.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


cli/cli.hs
```diff
@@ -203,7 +203,7 @@ main = do
     } }
   case cmd of
     Version {} ->putStrLn getFullVersion
-    Symbolic {} -> do
+    Symbolic' {} -> do
```
d-xo, line 206: This breaks the cli interface and turns `hevm symbolic` into `hevm symbolic'`. We should keep this as is

src/EVM.hs
```diff
@@ -2665,6 +2552,219 @@ writeMemory memory offset buf = do
     else
       pure memory
 
-freezeMemory :: MutableMemory s -> EVM s (Expr Buf)
+freezeMemory :: MutableMemory s -> EVM t s (Expr Buf)
 freezeMemory memory =
   ConcreteBuf . BS.pack . VUnboxed.toList <$> VUnboxed.freeze memory
+
+
+class VMOps (t :: VMType) where
```
d-xo, line 2560: It would be nice to pull this out into it's own section at the top of the file (or maybe into `Types.hs`?)

cli/cli.hs
arcz: fixed

src/EVM.hs
arcz: moved to Types, instances have to stay in EVM otherwise it would cause cyclic dependency

arcz:
@d-xo it's ready

d-xo:
Merged. Thanks <3

PR #438, opened 2024-01-14
add support for bitwuzla

d-xo:
## Description

Adds support for Bitwuzla. This is the same as the old PR, but uses a new bitwuzla version (0.3.0), which has support for timeouts This pr also messes with the smt generation less (only real change is replacing `define-const` with `define-fun`). 

It currently can't pass the full test suite (currently 9 failures due to issues support for comparison of constant arrays), but I enabled it for a few tests where it doesn't error out to make sure we're exercising it in CI.

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -160,11 +160,11 @@ declareIntermediates bufs stores =
   where
     compareFst (l, _) (r, _) = compare l r
     encodeBuf n expr =
-      SMT2 ["(define-const buf" <> (fromString . show $ n) <> " Buf " <> exprToSMT expr <> ")\n"]  mempty mempty mempty <> encodeBufLen n expr
+      SMT2 ["(define-fun buf" <> (fromString . show $ n) <> "() Buf " <> exprToSMT expr <> ")\n"]  mempty mempty mempty <> encodeBufLen n expr
     encodeBufLen n expr =
-      SMT2 ["(define-const buf" <> (fromString . show $ n) <>"_length" <> " (_ BitVec 256) " <> exprToSMT (bufLengthEnv bufs True expr) <> ")"] mempty mempty mempty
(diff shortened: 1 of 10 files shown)


src/EVM/Solvers.hs
```diff
@@ -272,21 +276,31 @@ solverArgs solver timeout = case solver of
...
+  let cmd
+        = (proc (show solver) (fmap T.unpack $ solverArgs solver timeout))
+            { std_in = CreatePipe
+            , std_out = UseHandle writeout
+            , std_err = UseHandle writeout
+            }
+  (Just stdin, Nothing, Nothing, process) <- createProcess cmd
   hSetBuffering stdin (BlockBuffering (Just 1000000))
-  let solverInstance = SolverInstance solver stdin stdout stderr process
+  let solverInstance = SolverInstance solver stdin readout process
```
d-xo, line 288: This sends stdout and stderr from the solvers into the same pipe so we can react to both when doing solver interaction...

src/EVM/SMT.hs
```diff
@@ -160,11 +160,11 @@ declareIntermediates bufs stores =
   where
     compareFst (l, _) (r, _) = compare l r
     encodeBuf n expr =
-      SMT2 ["(define-const buf" <> (fromString . show $ n) <> " Buf " <> exprToSMT expr <> ")\n"]  mempty mempty mempty <> encodeBufLen n expr
+      SMT2 ["(define-fun buf" <> (fromString . show $ n) <> "() Buf " <> exprToSMT expr <> ")\n"]  mempty mempty mempty <> encodeBufLen n expr
```
zoep, line 163: Why fun and not const? Is this bitwuzla related?

zoep (approved):
Looks good!

PR #446, opened 2024-01-31
Fix trace source mapping and indexed event args

arcz:
## Description
Source mapping in traces was broken due to the fact that simply matching by code hash is not robust. This is because we strip bytecode metadata before hashing in `SolContract` and contracts have code hash unstripped. The `findSrc` function already handles this. `findSrc` is unfortunately also bugged for contracts with immutables and `SolContract` probably shouldn't strip metadata before hashing but that's for another PR.

This PR also implements indexed event arguments and improves display by showing argument names.

Related issue https://github.com/crytic/echidna/issues/886.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Partial support for dynamic jumps when the jump destination can be computed
   given already available information
 
+## Fixed
+
+- Traces now correctly perform source mapping to display contract details
+- Event traces now correctly display indexed arguments and argument names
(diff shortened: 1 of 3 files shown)


msooseth (commented):
This does look really cool thanks! Unfortunately I am soooo unfamiliar with this part of the codebase. I can try to take another look tomorrow and see if I can figure it out. Otherwise, maybe we can wait for @d-xo , although I probably should try to start doing these by myself. Let me sleep on it and take another look tomorrow.

PR #447, opened 2024-02-01
Limit VMResult cases when concrete

arcz:
## Description
This builds on top of https://github.com/ethereum/hevm/pull/427. Better types using `VMType` allowed to eliminate dead code and optimize concrete execution with specialized `branch`. The speedup is ~10% on ethereum-tests.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1251,32 +1251,11 @@ getCodeLocation vm = (vm.state.contract, vm.state.pc)
 query :: Query t s -> EVM t s ()
 query = assign #result . Just . HandleEffect . Query
 
-choose :: Choose t s -> EVM t s ()
+choose :: Choose s -> EVM Symbolic s ()
 choose = assign #result . Just . HandleEffect . Choose
 
-branch :: forall t s. Expr EWord -> (Bool -> EVM t s ()) -> EVM t s ()
-branch cond continue = do
-  loc <- codeloc
-  pathconds <- use #constraints
-  query $ PleaseAskSMT cond pathconds (choosePath loc)
-  where
-    condSimp = Expr.simplify cond
-    condSimpConc = Expr.concKeccakSimpExpr condSimp
-    choosePath :: CodeLocation -> BranchCondition -> EVM t s ()
-    choosePath loc (Case v) = do
-      assign #result Nothing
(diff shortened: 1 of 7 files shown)


src/EVM.hs
```diff
@@ -1251,32 +1251,11 @@ getCodeLocation vm = (vm.state.contract, vm.state.pc)
 query :: Query t s -> EVM t s ()
 query = assign #result . Just . HandleEffect . Query
 
-choose :: Choose t s -> EVM t s ()
+choose :: Choose s -> EVM Symbolic s ()
```
msooseth, line 1254: Notes to self: so choose is only defined for Symbolic now.

src/EVM.hs
```diff
@@ -2718,6 +2715,10 @@ instance VMOps Concrete where
 
   whenSymbolicElse _ a = a
 
+  partial _ = internalError "won't happen during concrete exec"
```
msooseth, line 2718: Right, so `partial` cannot happen during concrete exec

src/EVM.hs
```diff
@@ -2589,6 +2565,27 @@ instance VMOps Symbolic where
   toGas _ = ()
   whenSymbolicElse a _ = a
 
+  partial e = assign #result (Just (Unfinished e))
```
msooseth, line 2568: `partial` now only needs to be defined for symbolic exec

src/EVM/Stepper.hs
```diff
@@ -46,7 +45,7 @@ data Action t s a where
   Wait :: Query t s -> Action t s ()
 
   -- | Multiple things can happen
-  Ask :: Choose t s -> Action t s ()
+  Ask :: Choose s -> Action Symbolic s ()
```
msooseth, line 48: This is nice actually.

msooseth (approved):
Actually makes the code cleaner and easier to read, too. I like it a lot, thanks!

src/EVM.hs
arcz: In the long run, `partial` should not be part of `VMOps` but it needs a slight refactor to other functions

PR #448, opened 2024-02-07
`CopySlice` simplification, more generic Array/Map slot simplification, fixing `decompose`, `--nodecompose` flag

zoep:
## Description
Changes:
* add simplification of `CopySlice` so that the output of hevm is closer to the Expr output of Act. Expressions of the form `CopySlice 0 0 size src (ConcreteBuf "")` are rewritten to `src` when `size` is equal to the size of the `src` buffer`. 
* `CopySlice` simplification necessitated improvement of both decomposition and `readStorage` improvements
* Decomposition of storage can now be disabled on the command line with `--nodecompose`
* `decompose` can safely skip writes to an `idx` it's not currently reading. This is because in `decompose` we have a very good idea of e.g. no mixing of map/array/other and we know exactly what idx is being read. So we have the safeguards necessary to do this.
* `decompose` is now *only* safe to execute in case everything safely decomposes. This is actually a requirement for `AbstractStore k`, as separate queries will deal with the same `AbstractStore k`. Aalthough we *could* be more lenient and have it decompose for every `AbstractStore k1` and not for `AbstractStore k2` and still be OK to decompose `AbstractStore k1` and not `AbstractStore k2`. However, the strict one works for now.
* We cannot decompose when it's `PEq` of the whole storage -- not with the current setup. We'd need a more sophisticated decompose for that. That kind of check only occurs during equivalence checking. We now check for this and don't perform decomposition on such queries  (rare).

## Checklist
- [X] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -97,6 +97,7 @@ data Command w
       , loopDetectionHeuristic :: w ::: LoopHeuristic <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
       , abstractArithmetic    :: w ::: Bool             <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , abstractMemory    :: w ::: Bool                      <?> "Use abstraction-refinement for Memory. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
+      , nodecompose   :: w ::: Bool               <?> "Don't decompose storage slots into separate arrays"
       }
   | Equivalence -- prove equivalence between two programs
       { codeA         :: w ::: ByteString       <?> "Bytecode of the first program"
@@ -116,6 +117,7 @@ data Command w
       , loopDetectionHeuristic :: w ::: LoopHeuristic <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
       , abstractArithmetic    :: w ::: Bool             <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , abstractMemory    :: w ::: Bool                      <?> "Use abstraction-refinement for Memory. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
+      , nodecompose   :: w ::: Bool             <?> "Don't decompose storage slots into separate arrays"
       }
   | Exec -- Execute a given program with specified env & calldata
       { code        :: w ::: Maybe ByteString  <?> "Program bytecode"
@@ -163,6 +165,7 @@ data Command w
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
       , abstractArithmetic    :: w ::: Bool             <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , abstractMemory    :: w ::: Bool                      <?> "Use abstraction-refinement for Memory. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
+      , nodecompose   :: w ::: Bool                     <?> "Don't decompose storage slots into separate arrays"
       , numCexFuzz    :: w ::: Integer                  <!> "3" <?> "Number of fuzzing tries to do to generate a counterexample (default: 3)"
       , askSmtIterations :: w ::: Integer               <!> "1" <?> "Number of times we may revisit a particular branching point before we consult the smt solver to check reachability (default: 1)"
       }
@@ -207,6 +210,7 @@ main = withUtf8 $ do
     , abstRefineMem = cmd.abstractMemory
     , abstRefineArith = cmd.abstractArithmetic
     , dumpTrace = cmd.trace
+    , decomposeStorage = Prelude.not cmd.nodecompose
     } }
   case cmd of
     Version {} ->putStrLn getFullVersion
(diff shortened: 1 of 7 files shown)


msooseth:
I have now fixed this PR I think, but the more general system that I now created needs a review. Furthermore, I am not sure which order should the `dop`/`take` be, that needs to be checked.

src/EVM/Expr.hs
```diff
@@ -936,6 +971,10 @@ simplify e = if (mapExpr go e == e)
 
     go (WriteByte a b c) = writeByte a b c
 
+    -- eliminate a CopySlice if the resulting buffer is the same as the src buffer
+    go (CopySlice (Lit 0) (Lit 0) (Lit s) src (ConcreteBuf ""))
+      | bufLength src == (Lit s) = src
+
```
msooseth, line 977: The rewrite rule that triggered this whole PR :)

src/EVM/SMT.hs
```diff
@@ -210,7 +210,13 @@ assertProps :: Config -> [Prop] -> SMT2
 assertProps conf ps = assertPropsNoSimp conf (decompose . Expr.simplifyProps $ ps)
   where
     decompose :: [Prop] -> [Prop]
-    decompose props = fromMaybe props (mapM (mapPropM Expr.decomposeStorage) props)
+    decompose props = if conf.decomposeStorage && (isJust $ sequence_ safeList1) && (isJust $ sequence_ safeList2)
+                      then fromMaybe props (mapM (mapPropM Expr.decomposeStorage) props)
+                      else props
+      where
+        -- All in these lists must be a `Just ()` or we cannot decompose
+        safeList1 = map (mapPropM_ Expr.safeToDecompose) props
```
zoep, line 218: `all $ map (foldProp Expr.safeToDecompose) props` would have the same effect as 
`(isJust $ sequence_ (map (mapPropM_ Expr.safeToDecompose) props))` without the use of monads

src/EVM/Traversals.hs
```diff
@@ -233,9 +233,44 @@ mapProp' f = \case
...
+mapPropM' :: forall m . (Monad m) => (Prop -> m Prop) -> Prop -> m Prop
+mapPropM' f = \case
+  PBool b -> f $ PBool b
+  PEq a b -> f $ PEq a b
+  PLT a b -> f $ PLT a b
+  PGT a b -> f $ PGT a b
+  PLEq a b -> f $ PLEq a b
+  PGEq a b -> f $ PGEq a b
+  PNeg a -> do
+    x <-mapPropM' f a
```
zoep, line 246:
```suggestion
    x <- mapPropM' f a
```

src/EVM/Traversals.hs
```diff
@@ -233,9 +233,44 @@ mapProp' f = \case
...
+    f $ PNeg x
+  PAnd a b -> do
+    x <- mapPropM' f a
+    y <- (mapPropM' f b)
```
zoep, line 250: is there a need for parentheses around the second `mapPropM'`?

src/EVM/Traversals.hs
```diff
@@ -233,9 +233,44 @@ mapProp' f = \case
...
+
 mapExpr :: (forall a . Expr a -> Expr a) -> Expr b -> Expr b
 mapExpr f expr = runIdentity (mapExprM (Identity . f) expr)
 
+-- Like mapExprM but allows a function of type `Expr a -> m ()` to be passed
+mapExprM_ ::  Monad m => (forall a . Expr a -> m ()) -> Expr b -> m ()
+mapExprM_ f expr = void ret
+  where
+    ret = mapExprM (fUpd f) expr
+    fUpd :: Monad m => (Expr a -> m ()) -> (Expr a-> m (Expr a))
```
zoep, line 269:
```suggestion
    fUpd :: Monad m => (Expr a -> m ()) -> (Expr a -> m (Expr a))
```

src/EVM/SMT.hs
msooseth: Ah nice! I didn't know about the `all` predicate! So much cleaner, thank you!

src/EVM/Traversals.hs
msooseth: Fixed!

cli/cli.hs
```diff
@@ -163,6 +165,7 @@ data Command w
       , loopDetectionHeuristic :: w ::: LoopHeuristic   <!> "StackBased" <?> "Which heuristic should be used to determine if we are in a loop: StackBased (default) or Naive"
       , abstractArithmetic    :: w ::: Bool             <?> "Use abstraction-refinement for complicated arithmetic functions such as MulMod. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
       , abstractMemory    :: w ::: Bool                      <?> "Use abstraction-refinement for Memory. This runs the solver first with abstraction turned on, and if it returns a potential counterexample, the counterexample is refined to make sure it is a counterexample for the actual (not the abstracted) problem"
+      , nodecompose   :: w ::: Bool                     <?> "Don't decompose storage slots into separate arrays"
```
arcz, line 168: Can we change it to `noDecompose`? The cli switch will be `--no-decompose` which I feel reads better and is similar to other CLI tools https://clig.dev/#arguments-and-flags. At the moment it could be also understood as "node compose".

src/EVM/Effects.hs
```diff
@@ -59,6 +59,7 @@ data Config = Config
    -- and uses the fuzzer ONLY to try to find a counterexample.
    -- Returns Unknown if the Cex cannot be found via fuzzing
   , onlyCexFuzz     :: Bool
+  , decomposeStorage:: Bool
```
arcz, line 62: missing space

src/EVM/Expr.hs
```diff
@@ -777,23 +780,48 @@ getLogicalIdx (GVar _) = internalError "cannot determine addr of a GVar"
...
 
-safeToDecompose :: Expr a -> Bool
-safeToDecompose inp = result /= Mixed
+-- We can't currently decompose cases when the FULL returned state is equated
+-- This is because the decomposition would need to take into account that ALL
+-- maps/arrays/small-slots need to be equivalent. This could be done, but is left
+-- as a TODO. Currently this only affects equivalence checking as there is no
+-- EVM bytecode to compare the FULL state, so such comparison could only be
+-- generated via hevm itself
+safeToDecomposeProp :: Prop -> Maybe ()
```
arcz, line 789: Is there a reason for returning `Maybe ()` instead of `Bool`?

cli/cli.hs
msooseth: Sure, sorry, good point. Fixed!

src/EVM/Effects.hs
msooseth: Sorry. Fixed!

src/EVM/Expr.hs
msooseth: Actually, no! And it's a lot cleaner if it's a Bool. Sorry, fixed :)

I tried doing the same for `safeToDecompose`, but because we need to map it over a Prop, and it takes in an Expr, it'd need one more traversal and I am not sure I should add it... so for the moment, I fixed this one that you highlighted, and left the other one as-is.

msooseth:
Thanks for the review @arcz ! Let me know if you think it's OK now :)

arcz (approved):

cli/cli.hs
```diff
@@ -207,6 +210,7 @@ main = withUtf8 $ do
     , abstRefineMem = cmd.abstractMemory
     , abstRefineArith = cmd.abstractArithmetic
     , dumpTrace = cmd.trace
+    , decomposeStorage = Prelude.not cmd.nodecompose
```
arcz, line 213: `cmd.noDecompose`

msooseth: Hehe, yes, sorry. I fixed it, now it's passing :)

PR #449, opened 2024-02-12
Fix `--root` option for `test` subcommand

msooseth:
## Description

The `--root` option for the `test` subcommand didn't work actually. This is because we `cwd`-d into the directory, then called all the functions with the directory name as a prefix, so the directory was in a sense prefixed twice. This is easy to test -- create a `tmp` subdirectory and `cd` into it, call `git init`, call `forge init --force`, add a solidity file under `src/` run `forge build` then go back with `cd ..` and then run `hevm test --root "tmp/"` it will fail.

With this change, it now succeeds, as expected.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   abstract gas execution but concrete running. In these cases, the interpreter can out-of-heap
   as the only check is that the size allocated is less than $2**{64}$, but that is too large to fit in memory. Now,
   we check more stringently, and still return an IllegalOverflow
+- Fixed `--root` option for the `test` subcommand
 
 ## Added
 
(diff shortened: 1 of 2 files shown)


arcz:
I think this is good. Also, `getCurrentDirectory` returns absolute path but `--root` can be relative so it might be better to use [makeAbsolute](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:makeAbsolute) in `getRoot` so it always returns absolute path: 
```
getRoot cmd = maybe getCurrentDirectory makeAbsolute (cmd.root)
```

msooseth:
> I think this is good. Also, `getCurrentDirectory` returns absolute path but `--root` can be relative so it might be better to use [makeAbsolute](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:makeAbsolute) in `getRoot` so it always returns absolute path:
> 
> ```
> getRoot cmd = maybe getCurrentDirectory makeAbsolute (cmd.root)
> ```

That's a really good point, thanks! I was actually confused when I was debugging what that relative path was that it was referring to. It would have been a lot easier to understand the bug with this already there. I added it in. I'll wait for the checks to finish and then merge it in :)

PR #451, opened 2024-02-15
Fix foundry JSON reading that was dependent on locale

msooseth:
## Description

This was a VERY weird and annoying bug. See full description here: https://github.com/ethereum/hevm/issues/450

We never caught it because nix happens to set the right locale.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog


arcz (approved):
We might also want to enable it for the whole program, more here https://serokell.io/blog/haskell-with-utf8

PR #461, opened 2024-02-26
Small nit in pretty printing writes 

zoep:
## Description

In the formatting of `WriteWord`, `WriteBuf`, and `CopySlice` destination buffers were printed outside of the parentheses, which made it hard to parse the expressions visually. Now they are printed within the parenthesis and in the same indentation block.


msooseth (approved):

msooseth:
Wow, that was bad printing. Thanks!

PR #465, opened 2024-03-04
[DRAFT] Update release.yml to allow for ARM64 binaries in the release

msooseth:
Currently just checking if this will work to release ARM64 binaries for Mac ARM64 as well.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,9 +5,11 @@ on:
       - 'release/[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
-  macosRelease:
-    name: Build MacOS Binary
+  macosRelease-x64:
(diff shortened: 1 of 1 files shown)


d-xo:
Did this work? Would be great to have arm64 binaries.

msooseth:
No, but then someone showed me how to do it!! Thanks for reminding me. Let me create a PR, this is the trick:

https://github.com/ethereum/hevm/pull/486/files

It's absolutely stupid. Version 13 is x86 and 14 is arm. That's it. It's... hilarious.

I'll create a PR in a minute. Thanks for reminding me. Also, sorry for thinking I did this already, it turns out I didn't.

PR #468, opened 2024-03-07
Implement label cheatcode

arcz:
## Description

Closes https://github.com/ethereum/hevm/issues/442.

I did a bit more work and fixed bugs in this PR too:
- There was a bug in cheatcode `CallContext` creation where the calldata was missing the first 4 bytes and was not displayed correctly in traces
- Unified address pretty printing
- Fixed address printing with leading @ when there is no source name / label
- Added "emit" before events, Foundry does this too
- Pretty print if bytes32 are valid string
- Changed colors a bit but I'll be revamping this anyway in a future PR. I want to turn this into prettyprinter Doc so it can be also used with Brick 
- If there is any name (from source or label) don't print address unless creating
- I reformatted other cheat actions, no semantic changes there

Here is how it looks like (the code doesn't make sense, it was just for my testing):
image

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Added
 - More POr and PAnd rules
 - More PEq, PLEq, and PLT rules
+- New `label` cheatcode.
 
 ## Fixed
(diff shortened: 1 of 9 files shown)


msooseth:
Thanks, looks great!

PR #476, opened 2024-03-14
Better docs

msooseth:
## Description
This is an ongoing effort to improve documentation.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog


doc/src/dstest-tutorial.md
```diff
@@ -0,0 +1,130 @@
+# Controlling the Unit Testing Environment
+
+## Writing Tests
+
+Test cases must be prepended with `prove_` and the testing contract must
+inherit from `Test` from (Forge's standard test
+library)[https://book.getfoundry.sh/forge/forge-std]. First, import Test:
+`import {Test} from "forge-std/Test.sol";` and then inherit from it via `... is
+Test`. This allows hevm to discover the test cases it needs to run. Like so:
```
joaovaladares, line 9: What about also talking about the proveFail test cases? They might be useful to some people even though you can do some workarounds with prove_ to test if it fails.

msooseth:
Ah, good point! Thanks, I will add it! :)

doc/src/dstest-tutorial.md
msooseth: Ah, good point, thanks! I have now added a section on this! Do see anything else missing perhaps? Any suggestions?

doc/src/dstest-tutorial.md
```diff
@@ -0,0 +1,243 @@
...
+    }
+}
+```
+
+The postconditions should check the state of the contract after the call(s) are
+complete. In particular, it should check that the changes that the function applied
+did not break any of the (invariants)[https://en.wikipedia.org/wiki/Invariant_(mathematics)] 
+of the contract, such as total number of tokens.
+
+You can read more about testing and cheat codes in the (Foundy
```
joaovaladares, line 78: Typo "Foundy"

doc/src/dstest-tutorial.md
```diff
@@ -0,0 +1,243 @@
...
+Here, hevm provided us with a calldata, where the receiver happens to be the
+zero address, and the value sent is exactly 100. This indeed is the boundary
+condition where the function call fails. The function should have had a `>=`
+rather than a `>` in the `if`. Notice that in this case, while hevm filled in
+the `address` to give a complete call, the address itself is irrelevant,
+although this is not explicitly mentioned.
+
+## Test Cases that Must Always Revert
+
+Hevm assumes that a test case should not always revert. If you have such a test case, Hevm will warn you and return a FAIL. For example this toy contract:
```
joaovaladares, line 122: Change second instance of "Hevm" to "hevm" to keep consistency since it's not on the beginning of a phrase

doc/src/dstest-tutorial.md
joaovaladares: Made some other pontual suggestions. I think the rest is fine. Also if needed I can help in the future on adding portuguese translated documentation :)

joaovaladares: Made some other pontual suggestions. I think the rest is fine. Also if needed I can help in the future on adding portuguese translated documentation :)

doc/src/dstest-tutorial.md
msooseth: Oops, thanks!

doc/src/dstest-tutorial.md
msooseth: Nice, thanks, fixed!

doc/src/getting-started.md
```diff
@@ -0,0 +1,152 @@
+# Getting Started
+
+The hevm project is an implementation of the Ethereum Virtual Machine (EVM)
+focused on symbolic analysis of evm bytecode. This essentially means that hevm
```
zoep, line 4:
```suggestion
focused on symbolic analysis of EVM bytecode. This essentially means that hevm
```

doc/readme.md
```diff
@@ -1,38 +1,40 @@
...
 ```
 
 [mdbook-serve]: https://rust-lang.github.io/mdBook/cli/serve.html
 
 ## Updating syntax highlighting
 
-In order to highlight the Solidity code examples in the book we override mdBook's built-in [highlight.js] with our own.
-To update the highlighting code, run the following commands using [yarn]:
+In order to highlight the Solidity code examples in the book we override
+mdBook's built-in [highlight.js] with our own. To update the highlighting code,
```
zoep, line 26:
```suggestion
mdBook's built-in [highlight.js] with our own. To update the syntax highlighting,
```

doc/src/SUMMARY.md
```diff
@@ -1,18 +1,17 @@
-# Summary
```
zoep, line 1: Maybe add an `# Overview` header and itemize the following three sections

readme.md
```diff
@@ -1,14 +1,15 @@
...
-The `hevm` project is an implementation of the Ethereum virtual machine (EVM) made specifically for
-symbolic execution, unit testing and debugging of smart contracts. The `hevm` command line program
-can symbolically execute smart contracts, run unit tests, interactively debug contracts while
-showing the Solidity source, or run arbitrary EVM code. Computations can be performed using local
-state set up in a `dapp` testing harness, or fetched on demand from live networks using `rpc` calls.
+`hevm` is an implementation of the Ethereum virtual machine (EVM) made for
+symbolic execution, equivalence checking, and unit testing of smart contracts.
+`hevm` can symbolically execute smart contracts, run unit tests, and run
+arbitrary EVM code. It can run on state set up in a [`ds-test` testing
+harness](https://book.getfoundry.sh/forge/forge-std), and/or fetched on demand from
```
zoep, line 7: Is there a reason for the `and/or`?

doc/src/exec.md
```diff
@@ -42,24 +43,29 @@ Available options:
...
 
-Minimum required flags:
-
-`--code` or (`--rpc` and `--address`).
-
-If the execution returns an output, it will be written to stdout.
-
-Exit code indicates whether the execution was successful or errored/reverted.
+Minimum required flags: either you must provide `--code` or you must both pass
+`--rpc` and `--address`. If the execution returns an output, it will be written
```
zoep, line 47:
```suggestion
`--rpc` and `--address`. 

If the execution returns an output, it will be written
```

doc/src/when-to-use.md
```diff
@@ -0,0 +1,162 @@
+# When to use Symbolic Execution
+
+In the cryptocurrency world, it is exceedingly easy to lose a [lot of
+assets](https://chainsec.io/defi-hacks/) due to bugs. This risk can be alleviated
+via a good test approach. Testing is a [discipline of its
```
zoep, line 5:
```suggestion
via a good testing approach. Testing is a [discipline of its
```

doc/src/ds-test-tutorial.md
```diff
@@ -0,0 +1,244 @@
...
+}
+```
+
+The postconditions should check the state of the contract after the call(s) are
+complete. In particular, it should check that the changes that the function applied
+did not break any of the (invariants)[https://en.wikipedia.org/wiki/Invariant_(mathematics)]
+of the contract, such as total number of tokens.
+
+You can read more about testing and cheat codes in the (Foundry
+Book)[https://book.getfoundry.sh/forge/cheatcodes]. Below are the cheat codes
```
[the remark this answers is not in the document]
msooseth: NICE, yess, thanks!

doc/src/ds-test-tutorial.md
```diff
@@ -0,0 +1,206 @@
...
+Which is now the expected outcome.
+
+
+## Supported Cheat Codes
+
+Since hevm is an EVM implementation mainly dedicated to testing and
+exploration, it features a set of `cheat codes` which can manipulate the
+environment in which the execution is run. These can be accessed by calling
+into a contract (typically called `Vm`) at address
+`0x7109709ECfa91a80626fF3989D68f67F5b1DD12D`, which implements the following
```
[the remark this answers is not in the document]
msooseth: Good point! I just added it, thanks!

doc/src/ds-test-tutorial.md
```diff
@@ -0,0 +1,206 @@
...
+methods
+
+| Function | Description |
+| --- | --- |
+|`function prank(address sender) public`| Sets `msg.sender` to the specified `sender` for the next call.|
+|`function deal(uint usr, uint amt) public`| Sets the eth balance of `usr` to `amt`. Note that if `usr` is a symbolic address, then it must be the address of a contract that has already been deployed. This restriction is in place to ensure soundness of our symbolic address encoding with respect to potential aliasing of symbolic addresses.|
```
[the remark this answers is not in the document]
msooseth: Ooops, yes! I think someone caught this and then I accidentally re-introduced this when doing the table. Sorry. Thanks for catching it. Fixed.

msooseth:
> Great update
> 
> I find it a bit curious that there is a circular reference between foundry and hevm. The docs from one cite the other. Not sure if there's a way out

Thanks for your help!! Yeah, not sure there is an out, but I think it's also OK. Not perfect, but maybe an OK start. I will probably improve things as time goes on, hopefully I'll find a way to deal with that as well :) Thanks again for your help,

Mate

readme.md
msooseth: Nope, good point! Replaced with "or" :)

doc/src/when-to-use.md
```diff
@@ -0,0 +1,162 @@
...
+malicious/accidental). It normally consists of a test approach, documenting
+what has been tested and what remains to be tested. It should consist of a
+number of unit tests, module tests, integration tests, user acceptance tests,
+and exploratory testing. Written test cases normally consist of both positive
+tests (i.e. correct input, correct behaviour), and negative tests (i.e.
+incorrect input, no incorrect behaviour).
+
+## Fuzzing versus Symbolic Execution
+
+Fuzzing and symbolic execution fits within the [exploratory
```
zoep, line 18: I wound't say that symbolic execution is a testing approach (also it is not claimed in the referenced wikipedia link).

doc/src/when-to-use.md
```diff
@@ -0,0 +1,162 @@
...
+
+Fuzzers are exceedingly fast and efficient when there are many potential faults
+with a function/contract, or if the faults are of a type that's easy to search
+for (e.g. off-by-one). However, they rarely, if ever, find cases where the bug
+is hidden deep in the branch logic, or needs very specific input parameters.
+Hence, it is best to use fuzzers at first to find easy-to-find bugs, as fuzzers
+are very efficient at that. Then, once the tests pass the fuzzer, it is
+recommended to use a symbolic execution engine such as hevm.
+
+While hevm is similar to [Halmos](https://github.com/a16z/halmos) and
```
zoep, line 128:
```suggestion
hevm is similar to [Halmos](https://github.com/a16z/halmos) and
```

doc/src/getting-started.md
```diff
@@ -0,0 +1,152 @@
...
+can try out _all_ execution possibilities of your contract and see it can
+somehow violate some assertions you have, such as e.g. the total number of
```
zoep, line 6: Maybe also mention other errors as well (overflows etc.)

zoep (approved):
This is a great improvement of the docs! Thank you! I only had a couple of nits to point out.

doc/src/ds-test-tutorial.md
```diff
@@ -0,0 +1,206 @@
...
+|`function store(address c, bytes32 loc, bytes32 val) public`| Sets the slot `loc` of contract `c` to `val`.|
+|`function warp(uint x) public`| Sets the block timestamp to `x`.|
+|`function roll(uint x) public`| Sets the block number to `x`.|
+|`function assume(bool b) public`| Add the condition `b` to the assumption base for the current branch. This functions almost identically to `require`.|
```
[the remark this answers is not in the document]
msooseth: Actually, I also am not 100% sure -- it was worded like that before, so I left it as-is. Looking at the code:

```
  , action "assume(bool)" $
      \sig _ _ input -> case decodeStaticArgs 0 1 input of
        [c] -> modifying #constraints ((:) (PEq c (Lit 1)))
        _ -> vmError (BadCheatCode sig)
```

It looks to me indeed very similar in behaviour. The difference is likely that `require` is built-in and so solidity will be able to understand and maybe optimize it, potentially even optimizing it away. However, `require` is a call out to another contract, which Solidity cannot "see inside", so e.g. it cannot be optimized away.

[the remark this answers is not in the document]
msooseth: For the average developer, the two are incredibly similar, and I'd recommend using `require`. I'll probably add that to the text. Thanks for poking at this! :)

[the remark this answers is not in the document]
msooseth: I hope it's good now. I am closing because I think I managed to explain this more thoroughly to the potential users. Let me know if it's not OK, and then I'll improve :) The final text is:


Add the condition `b` to the assumption base for the current branch. This functions almost identically to `require`. For most users, `require` is preferable. However, in case you wish to understand & modify the internal IR of hevm, you may want to use `assume`.

doc/src/when-to-use.md
msooseth: Ooops, thanks for catching this! I always thought that fuzzing is kind of an automation of exploratory testing, but that does sound bogus. I fixed it to:

```
Fuzzing tests usually have a set of (sometimes implicit) pre- and
postconditions, but the actual action (e.g. function call) is performed by an
external entity, the fuzzer. For C/C++ fuzzing, the implicit postcondition is
often e.g. that the system does not throw a segmentation fault. For EVM
bytecode, postconditions need to be explicit. Let's see an example:
```

What do you think?

doc/src/getting-started.md
msooseth: Added overflow, good point! I don't know if I wanna put too many here, but maybe we can have a section in `when-to-use.md` ?

doc/src/when-to-use.md
```diff
@@ -0,0 +1,162 @@
...
+via a good test approach. Testing is a [discipline of its
+own](https://en.wikipedia.org/wiki/Software_testing) that includes far more
+than simple test cases. Normally, it should include an understanding of what
+needs to be tested (i.e. scope) and for what types of issues (e.g.
+malicious/accidental). It normally consists of a test approach, documenting
+what has been tested and what remains to be tested. It should consist of a
+number of unit tests, module tests, integration tests, user acceptance tests,
+and exploratory testing. Written test cases normally consist of both positive
+tests (i.e. correct input, correct behaviour), and negative tests (i.e.
+incorrect input, no incorrect behaviour).
```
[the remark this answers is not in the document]
msooseth: Actually, really good point. I have no idea why I went on a tangent here, it was unnecessary. Thanks for that. I have now replaced it with:

```
In the cryptocurrency world, it is exceedingly easy to lose a [lot of
assets](https://chainsec.io/defi-hacks/) due to bugs. While fuzz testing can
help find potential issues with digital contracts, it is a tool that can only
execute the program concretely, one execution at a time. In contrast, Symbolic
Execution can execute all potential values in a decision path "in one go",
creating a symbolic expression out of a path, and checking whether it
can trigger a fault. Hence, Symbolic Execution tends to be more efficient at
finding bugs than fuzzing when the bugs are rare, or explicitly (i.e.
maliciously) hidden. Symbolic Execution can also _prove_ that no postcondition
can be violated, increasing the overall confidence in the contract. Note, however,
that Symbolic Execution does not automatically generate postconditions for
well-known bug classes like static code analysis tools do. Instead, these
postconditions, and their sometimes associated preconditions, need to
be explicitly written.
```

Let me know what you think!

doc/src/getting-started.md
```diff
@@ -0,0 +1,152 @@
...
+
+In some sense, hevm is similar to a fuzzer, but instead of only trying with random
+values to trigger faults, it instead _computes_ whether a fault is possible. If
+it is possible, it gives an example call to trigger the fault, and if it isn't
+possible, it mathematically proves it so, and tells the user the contract is
+safe. Note that while great pains have gone into making sure hevm's results can
+be trusted, there can always be bugs in hevm or the libraries and tools it uses.
+
+Hevm can not only be used to find bugs in already written programs, but can
+also help you write a program that from the outside behaves the same as the
```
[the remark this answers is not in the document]
msooseth: OOops, good point! Fixed :)

doc/src/ds-test-tutorial.md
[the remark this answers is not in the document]
msooseth: Hahaha, sorrrryyyy! Pinky promise I'll try to do better!

PR #482, opened 2024-04-08
Solidity: ignore some more kontrol related build artifacts

d-xo:
## Description

Ignores some more kontrol related build output.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Turns out the `build-info` directory is by new forge as well. So I integrated this into: https://github.com/ethereum/hevm/pull/510 and that should also update to newer forge, updates the doc, and adds debug info when requested with `--debug`. So I hope it's OK if I close this?

PR #484, opened 2024-04-17
Fix solver support on Windows

elopez:
## Description

The solver output on Windows has `\r\n` line endings, which the code doesn't account for. We need to trim the extra `\r` for it to work properly. This also cleans up extra concatenations by using the `*Ln` function variants.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
lgtm

PR #489, opened 2024-04-19
Use let expression in copySlice to decrease expression size

### LINKED ISSUE #487: `copySlice` can return giant SMT queries from reasonable-sided expressions
`copySlice`, defined [here](https://github.com/ethereum/hevm/blob/main/src/EVM/SMT.hs#L901-L911), copies its `src` argument `size` times into the resulting SMT string. It can get even worse with nested `CopySlice`s, where we end up getting exponential size. In particular, when `exprToSMT` is called on this expression...
```
(CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x40) (WriteWord (Lit 0x0) (SLoad (Keccak (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x40) (WriteWord (Lit 0x0) (SLoad (Keccak (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x40) (WriteWord (Lit 0x0) (Var "arg1") (ConcreteBuf "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\SOH")) (ConcreteBuf ""))) (ConcreteStore (fromList []))) (ConcreteBuf "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\SOH")) (ConcreteBuf ""))) (ConcreteStore (fromList []))) (ConcreteBuf "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL")) (ConcreteBuf ""))
```
...`exprToSMT` returns a 35-million-character-long value.

Ideally the `prelude` passed to the SMT solver would include a recursive `copySlice` function which handles the functionality that's now done in haskell, but I'm not sure how this would affect SMT performance. This would be ideal because `src` would only need to be typed out once in the SMT expression.

----------------------------

I can go into more detail, if needed, on how this expression came up in practice. Long story short, this expression was involved while testing [a symbolic execution PR for echidna](https://github.com/crytic/echidna/pull/1216) on the following contract:
```
pragma solidity 0.8.13;
contract Challenge {
    mapping(uint256 => uint256[]) private _mapA;
    mapping(uint256 => uint256) private _mapB;
    function challenge(uint256 a) external {
        uint256 val;
        unchecked {
            val = _mapA[_mapB[_mapB[a]]][0];
        }   
        require(val == 0); 
    }   
}
```

samalws-tob:
## Description
This PR changes `copySlice` so that it returns an expression of the form `(let ((src [value of src])) (store (store [value of dst] 0 (select src 0)) 1 (select src 1))`, rather than `(store (store [value of dst] 0 (select [value of src] 0)) 1 (select [value of src] 1))`. This reduces the number of times src has to be written out in the SMT expression from `size` times to only 1 time. This can drastically reduce the size of the resulting SMT expression, especially in the case of nested `CopySlice`s.
Fixes #487 .

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Changed
 - Improved printing of results. Should be more intuitive to understand what hevm found.
 - More complete and precise array/mapping slot rewrite, along with a copySlice improvement
+- Use a let expression in copySlice to decrease expression size
 
 ## Added
 - More POr and PAnd rules
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -900,14 +900,16 @@ propToSMT = \case
 
 -- | Stores a region of src into dst
 copySlice :: Expr EWord -> Expr EWord -> Expr EWord -> Builder -> Builder -> Builder
-copySlice srcOffset dstOffset size@(Lit _) src dst
-  | size == (Lit 0) = dst
-  | otherwise =
-    let size' = (Expr.sub size (Lit 1))
-        encDstOff = exprToSMT (Expr.add dstOffset size')
-        encSrcOff = exprToSMT (Expr.add srcOffset size')
-        child = copySlice srcOffset dstOffset size' src dst in
-    "(store " <> child `sp` encDstOff `sp` "(select " <> src `sp` encSrcOff <> "))"
+copySlice srcOffset dstOffset size0@(Lit _) src dst =
+  "(let ((src " <> src <> ")) " <> internal size0 <> ")"
+  where
+    internal (Lit 0) = dst
+    internal size =
+      let size' = (Expr.sub size (Lit 1))
+          encDstOff = exprToSMT (Expr.add dstOffset size')
+          encSrcOff = exprToSMT (Expr.add srcOffset size')
+          child = internal size' in
+      "(store " <> child `sp` encDstOff `sp` "(select src " <> encSrcOff <> "))"
 copySlice _ _ _ _ _ = internalError "TODO: implement copySlice with a symbolically sized region"
 
 -- | Unrolls an exponentiation into a series of multiplications


arcz (approved):
I think it looks good and it's a great catch. @d-xo @msooseth @zoep would you take a look too?

zoep (approved):
This looks good to me too! But I'd like @d-xo or @msooseth to have a look too

ggrieco-tob:
I'm wondering what is the status of this PR?

samalws-tob:
@ggrieco-tob  PR is ready but waiting for review from @d-xo or @msooseth

d-xo:
In general a positive review from @zoep or @arcz is imo enough to merge.

d-xo:
I am a little confused why the common sub expression elimination pass isn't handling this during smt generation?

zoep:
> I am a little confused why the common sub expression elimination pass isn't handling this during smt generation?

The `copySlice` expansion happens post-CSE, which happens at the Expr level, so this repeating subterm will not be eliminated. 

Regarding SMT encoding, is there any difference between local let and a top-level constant declaration (e.g., in terms of performance)?

PR #498, opened 2024-07-08
Upgrade solc to 0.8.24

blishko:
## Description
Rationale: sold releases prior to 0.8.24 did not contain an executable for arm64 chips. Without Rosetta, the part of `hevm` that requires calling `solc` wouldn't work on arm-based OSX.

Upgrading solc also showed that `hevm` loops if the bytecode contains infinite recursion using only `JUMP` (and not `JUMPI`) instructions. For this reason we now skip a few tests that contain infinite recursion.

## Checklist
- [ x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):
lgtm

PR #500, opened 2024-07-09
Initial support for TSTORE and TLOAD opcodes

samalws-tob:
## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM.hs
```diff
@@ -176,6 +176,7 @@ unknownContract :: Expr EAddr -> Contract
 unknownContract addr = Contract
   { code        = UnknownCode addr
   , storage     = AbstractStore addr Nothing
+  , t_storage   = AbstractStore addr Nothing
```
msooseth, line 179: I think we prefer not using underscore, so maybe tStorage or tempStorage would work better?

src/EVM/Format.hs
```diff
@@ -663,13 +663,15 @@ formatExpr = go
 
       BufLength b -> fmt "BufLength" [b]
 
-      C code store bal nonce -> T.unlines
+      C code store tStore bal nonce -> T.unlines
```
msooseth, line 666: Here you called it tStore, which I think may be OK?

src/EVM/Format.hs
```diff
@@ -663,13 +663,15 @@ formatExpr = go
...
         [ "(Contract"
         , indent 2 $ T.unlines
           [ "code:"
           , indent 2 $ formatCode code
           , "storage:"
           , indent 2 $ formatExpr store
+          , "t_storage:"
```
msooseth, line 673: Here too, I think

src/EVM/Op.hs
```diff
@@ -335,6 +337,8 @@ getOp x = case x of
   0x59 -> OpMsize
   0x5a -> OpGas
   0x5b -> OpJumpdest
+  0x5c -> OpTload
```
msooseth, line 340: I think it may be best calling it `TStore` and `TLoad` instead of `Tstore` and `Tload`

msooseth (changes requested):
Wow, nice! I think the only thing that remains is adding a few tests to check temp storage use and minor changes to some tests for them to pass. What do you think, do you wanna make these changes?

samalws-tob:
Eventually I can work on testing for this, I won't be available to do that for the next month or so though

samalws-tob:
Also any suggestions on where `clearTStorages` should be called from? Is there some transaction related code that we can hook it into?

d-xo:
> Also any suggestions on where `clearTStorages` should be called from? Is there some transaction related code that we can hook it into?

I think [`finalize`](https://github.com/ethereum/hevm/blob/789b6a95ab080ebe79b4744030975c8ee4f78d6e/src/EVM.hs#L1356) would be the spot probably.

If we update `ethereum/tests` to the latest then I think we should get most of the required tests for free?

src/EVM.hs
```diff
@@ -1335,6 +1357,33 @@ accessStorage addr slot continue = do
...
+      --     the expression is of a particular format, which can be destroyed by simplification.
+      --     However, without concretization, it may not find things that are actually in the storage
+      case readStorage slot c.tStorage of
+        Just x -> case readStorage slotConc c.tStorage of
+          Just _ -> continue x
+          Nothing -> continue $ Lit 0
+        Nothing -> continue $ Lit 0
+    Nothing ->
+      fetchAccount addr $ \_ ->
+        accessStorage addr slot continue
```
msooseth, line 1381:
```suggestion
        accessTStorage addr slot continue
```

msooseth: Copy-paste bug ;)

msooseth:
I'm gonna try to create a new PR with some changes, but keeping your history so your contributions remain clear. Let me know how you feel about it :)

samalws-tob:
Sounds good, thanks 👍

msooseth:
Moved to https://github.com/ethereum/hevm/pull/529 and will continue working on it there. Closing!

PR #501, opened 2024-07-11
Enable testing on Windows

elopez:
## Description

This (mostly) fixes hevm (or should I say libff 😅 ) on Windows and enables CI testing.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/scripts/libff.patch
+++ b/.github/scripts/libff.patch
@@ -1,0 +1,215 @@
+diff --git a/libff/algebra/fields/bigint.tcc b/libff/algebra/fields/bigint.tcc
+index d78e94f..0b08431 100755
+--- a/libff/algebra/fields/bigint.tcc
++++ b/libff/algebra/fields/bigint.tcc
+@@ -25,39 +25,32 @@ bigint<n>::bigint(const unsigned long x) /// Initalize from a small integer
+ template<mp_size_t n>
+ bigint<n>::bigint(const char* s) /// Initialize from a string containing an integer in decimal notation
+ {
+-    size_t l = strlen(s);
+-    unsigned char* s_copy = new unsigned char[l];
+-
+-    for (size_t i = 0; i < l; ++i)
+-    {
+-        assert(s[i] >= '0' && s[i] <= '9');
+-        s_copy[i] = s[i] - '0';
+-    }
++    mpz_t k;
++    mpz_init(k);
+ 
+-    mp_size_t limbs_written = mpn_set_str(this->data, s_copy, l, 10);
+ #ifndef NDEBUG
+-    assert(limbs_written <= n);
+-#else
+-    UNUSED(limbs_written);
++    assert(mpz_size(k) <= n);
+ #endif
+ 
+-    delete[] s_copy;
++    mpz_set_str(k, s, 10);
++    for (size_t i = 0; i < n; i++)
++    {
++        data[i] = mpz_getlimbn(k, i);
++    }
++    mpz_clear(k);
+ }
+ 
+ template<mp_size_t n>
+ bigint<n>::bigint(const mpz_t r) /// Initialize from MPZ element
+ {
+-    mpz_t k;
+-    mpz_init_set(k, r);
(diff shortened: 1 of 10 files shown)


test/rpc.hs
```diff
@@ -69,7 +75,7 @@ tests = testGroup "rpc"
     ]
   , testGroup "execution with remote state"
     -- execute against remote state from a ds-test harness
-    [ test "dapp-test" $ do
+    [ ignoreTestWindows $ test "dapp-test" $ do
```
d-xo, line 78: why don't these work on windows?

test/test.hs
d-xo, line 1: curious into why the tests that are ignored on windows here are failing there?

.github/scripts/libff.patch
d-xo, line 1: curious as to the contents of the patch here?

elopez: The gist of the patch is that libff uses mpz_*_ui functions (that return unsigned long) to operate on 64 bits of data at a time. This works on Linux/Max because unsigned long is indeed 64 bits there. But on Windows, unsigned long is 32 bits, so the functions don't work as expected. This patch rewrites them to behave correctly, by mainly using functions that operate on "limbs" which are the right size.

test/test.hs
elopez: I think they all use bitwuzla, which I think is not supported on Windows. If I'm mistaken and it is and there's binaries somewhere, we can install it in CI and run these though.

test/rpc.hs
elopez: Some of the git invocations that hevm did were failing and it was not obvious to me why. There's a comment with the error in the commit that added this ignore.

d-xo: could we just drop a source level comment here to make the reasoning here a little more accesible?

test/test.hs
d-xo: ah, perhaps we can just conditionally use cvc5 or z3 on windows for these tests or something?

.github/scripts/libff.patch
```diff
@@ -0,0 +1,215 @@
...
+ 
+ template<mp_size_t n>
+@@ -127,7 +120,10 @@ size_t bigint<n>::num_bits() const
+         }
+         else
+         {
+-            return ((i+1) * GMP_NUMB_BITS) - __builtin_clzl(x);
++            if(sizeof(x) == sizeof(unsigned long))
++                return ((i+1) * GMP_NUMB_BITS) - __builtin_clzl(x);
++            else
```
msooseth, line 66: I'd probably put an `assert` here to make sure it's `sizeof(unsigned long long)`. But maybe I'm too cautious.

msooseth:
Wow, this is some serious work, thanks! Is this ready or is there a specific reason why it's marked for `work in progress`? Is there something I can help with?

elopez:
@msooseth I assumed you may not want to carry a patch on libff in the repo so I left it marked as WIP, but if you're ok with it I guess we can consider this ready for review 👍

msooseth:
I mean, upstreaming it to libff would probably be best :) But we can have this in the repo as long as it works I guess? Due to nix we don't automatically update libff unless really necessary.

elopez:
I resolved the conflicts and refreshed the branch after the 9.6 merge 😄  I also added some more workflow code to build bitwuzla for Windows, so now those tests also run on Windows CI 🧪 The built binaries are cached so as to not slow down the CI too much.

Test-wise it'd be nice to see why the dapp/foundry Git stuff fails on Windows, then Windows would be at test parity with the other platforms. In the meantime I added a reason string to the ignore function.

I also had to add a new flag to the cabal file, as the newer secp256k1 now requires a new #define to be set so as to link correctly when building hevm against the static library on Windows.

msooseth:
Thanks! Looks nice. If it ever breaks, you may need to help, as none of us are Windows experts. But I hope it won't break too often :) Thank you again,

Mate

PR #503, opened 2024-07-15
[DRAFT] Better cheatcode error texts

### LINKED ISSUE #497: Improve known cheat code errors
We return non-helpful errors when there is a problem with a known cheat code. For instance, [here](https://github.com/ethereum/hevm/blob/main/src/EVM.hs#L1585-L1586) we know we are inside `ffi` but we return `BadCheatCode` with `sig` which doesn't add much debugging value.

Ideally, in the error, we should have the method + args like `ffi(string[])` and a failure reason.

msooseth:
## Description

This PR aims to fix #497 so we can have better cheatcode errors. However, I am still struggling with actually triggering an error :) With some time I will manage. In the meanwhile, I am creating this PR so I can work on it.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Can you maybe trigger one of these faults so we can see if this is working @arcz ? I just couldn't find a way to trigger them :(

blishko (approved):
LGTM!

PR #508, opened 2024-07-22
More Prop simplifications

msooseth:
## Description

This does a number of Prop simplifications. I needed it to debug another issue. Basically:

* `[PAnd (PAnd x y) z]` now becomes `[x, y, z]`, previously it was `[x, PAnd y z]`
* Double negation not only works for `Eq` inside Prop, but also for `LT`, `GT`, etc

This makes it a lot easier to debug some issues (much less deep trees), and accidentally makes things faster, too. Was not the intention, though.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1298,13 +1298,8 @@ simplifyProp prop =
 flattenProps :: [Prop] -> [Prop]
 flattenProps [] = []
 flattenProps (a:ax) = case a of
-  PAnd x1 x2 -> (subFlatten x1) ++ (subFlatten x2) ++ flattenProps ax
+  PAnd x1 x2 -> flattenProps [x1] ++ flattenProps [x2] ++ flattenProps ax
   x -> x:flattenProps ax
-  where
-    subFlatten :: Prop -> [Prop]
-    subFlatten b = case b of
-      PAnd y1 y2 -> flattenProps [y1,y2]
-      _ -> [b]
 
 -- removes redundant (constant True/False) props
 remRedundantProps :: [Prop] -> [Prop]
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -1292,8 +1298,13 @@ simplifyProp prop =
...
 flattenProps [] = []
 flattenProps (a:ax) = case a of
-  PAnd x1 x2 -> x1:x2:flattenProps ax
+  PAnd x1 x2 -> (subFlatten x1) ++ (subFlatten x2) ++ flattenProps ax
   x -> x:flattenProps ax
+  where
+    subFlatten :: Prop -> [Prop]
+    subFlatten b = case b of
+      PAnd y1 y2 -> flattenProps [y1,y2]
+      _ -> [b]
```
blishko, line 1307: Can we do without the helper function? Like this?
```haskell
flattenProps :: [Prop] -> [Prop]
flattenProps [] = []
flattenProps (a:ax) = case a of
  PAnd x1 x2 -> flattenProps [x1] ++ flattenProps [x2] ++ flattenProps ax
  x -> x:flattenProps ax
```

Would the helper function be more efficient for any reason?

src/EVM/Expr.hs
```diff
@@ -1224,7 +1224,6 @@ simplifyProp prop =
     go (PNeg (PNeg a)) = a
 
     -- solc specific stuff
-    go (PEq (IsZero (Eq a b)) (Lit 0)) = PEq a b
```
blishko, line 1227: Why was this removed?

blishko (commented):
Looks good, I just have some questions.

src/EVM/Expr.hs
msooseth: It was only moved lower, to the block about double negation, I think it fits there better?

src/EVM/Expr.hs
msooseth: Oh, of course! Good point :D Fixed!

PR #509, opened 2024-07-22
cli: fix --initial-storage option for hevm symbolic

d-xo:
## Description

We weren't respect the `initial-storage` option when creating the initial contract in `symVmFromCommand`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
I just hand-tested this. Would be nice to have a test actually! But merging :)

PR #510, opened 2024-07-23
Update `forge` to more recent version

msooseth:
## Description
Some people have bumped into the issue where the new `forge` had issues with new forge's build not being parsed. This fixes the issue. We now:
* Use new `forge` and `forge-std` internally
* Don't parse JSON's that are new and not useful for us
* Need `--ast` flag as per:`forge build --ast`. This was the default previously.
* Allow Env to be pushed down into the JSON parsing part so we can debug more effectively using `--debug`

This needs updater to the docs, too, due to `--ast`. This is unfortunate, we only need it to know the `absolutePath` whicih is actually a relative path like `src/multiply-test.sol"

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog


doc/src/ds-test-tutorial.md
```diff
@@ -98,7 +98,7 @@ contract MyContract is Test {
 }
 ```
 
-After compiling with `forge build`, when ran under hevm, we get:
+After compiling with `forge build --ast`, when ran under hevm, we get:
```
d-xo, line 101: I think there is also probably some config file option that forge supports to force this? It might be a nicer workflow to recomend that people set that in the tutorial?

d-xo (approved):
code looks fine. one minor comment about the tutorial, but feel free to merge if you disagree with me there.

doc/src/ds-test-tutorial.md
msooseth: Good point! Thanks, `GPT-40` allowed me to find this config setting! Weirdly enough, the `Phind-70B` model told me it doesn't exist. I am now closing & merging. Thanks again!

msooseth:
Thanks for the review, it was a good catch!

PR #520, opened 2024-08-01
Changing default --max-iterations to 5

msooseth:
## Description
Maximum iterations explored of loops has been set to 5 by default instead of the infinite before. This is to:
- Keep more in line with other symbolic execution systems' defaults (e.g hevm). They tend to a have a fixed default, usually 1. However, 1 is so low that it can cause issues with std-test.
- Not get stuck on a single loop and find potential issues outside of the loop. Otherwise, users would have a very bad experience of getting stuck on one thing where there could be bugs elsewhere.

We now print a better warning, with explanation how to increase the loop bound.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `--debug` flag now dumps the internal expressions as well
 - hevm now uses the forge-std library's way of detecting failures, i.e. through
   reverting with a specific error code unless --assertion-type DSTest is passed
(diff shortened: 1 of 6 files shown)


d-xo:
I think that this is because of some abi encoding stuff that happens in `fail` in `ds-test` unfortunately, shouldn't be too hard to modify `ds-test`, or we could also just switch to these cheatcode based assertions from foundry and move on with our lives probably.

msooseth:
@d-xo yay, now that we moved assertions to foundry, indeed a max value of 5 works fine! I wanna merge this for the 0.54. What do you think?

PR #522, opened 2024-08-01
SymExec: symbolic abi encoding for tuples (+ fuzzing)

msooseth:
## Description
Originally by @d-xo. There was a little bug in there, which I had a hard time finding, but it's fixed now. Basically `y101` is ambiguous. So now it's `y-a-1-a-01` which is not ambiguous.

I left in a bunch of debug code for `symbolic-abi-enc-vs-solidity`. I think that may come handy one day. It certainly helped me understand the encoding and find the bug.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Mergning after review by @ggrieco-tob

PR #525, opened 2024-08-12
Print OpCode on partial exec failure

msooseth:
## Description

Before:
```
       msg: "trying to access a symbolic address that isn't already present in storage"
       program counter: 73
       arguments:
         (SymAddr "caller")
```

Now:
```
       msg: "trying to access a symbolic address that isn't already present in storage"
       opcode: CALL
       program counter: 73
       arguments:
         (SymAddr "caller")
```

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -230,8 +230,8 @@ isCreation = \case
 next :: (?op :: Word8) => EVM t s ()
 next = modifying (#state % #pc) (+ (opSize ?op))
 
-getOpInt :: forall (t :: VMType) s . FrameState t s -> Word8
-getOpInt state = case state.code of
+getOpW8 :: forall (t :: VMType) s . FrameState t s -> Word8
+getOpW8 state = case state.code of
       UnknownCode _ -> internalError "Cannot execute unknown code"
       InitCode conc _ -> BS.index conc state.pc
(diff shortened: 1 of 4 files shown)


src/EVM.hs
```diff
@@ -230,6 +230,17 @@ isCreation = \case
 next :: (?op :: Word8) => EVM t s ()
 next = modifying (#state % #pc) (+ (opSize ?op))
 
+getOpName :: forall (t :: VMType) s . FrameState t s -> [Char]
+getOpName state = intToOpName $ fromEnum $ (?op)
```
blishko, line 234:
```suggestion
getOpName state = intToOpName $ fromEnum (?op)
```

I also don't understand why do you call it `?op`.
From what I read question mark at the beginning is used for implicit parameters, but that is not the case here, no?
Why not call it just `op`?

msooseth: Good catch! Let me try to fix...

blishko (approved):
LGTM!

src/EVM.hs
msooseth: Can you re-review? I cleaned up everything :)

Issue #524, opened 2024-08-10
Unexpected Symbolic Arguments to Opcode (SymAddr "caller")


ethever:
``` bash
     - Unexpected Symbolic Arguments to Opcode
       msg: "trying to access a symbolic address that isn't already present in storage"
       program counter: 69
       arguments: 
         (SymAddr "caller")
```
for this solidity code: 
``` solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import "@openzeppelin/contracts/interfaces/IERC20.sol";

contract ERC20_2 {
    mapping(address => uint256) balanceOf1;
    mapping(address => uint256) balanceOf2;

    mapping(address => uint256) balanceOf;

    address owner = address(0x123456789abcdef);

    function a() public {
        payable(msg.sender).transfer(1234);
    }

}
```
with command: 
``` bash
forge build   &&  hevm symbolic --caller 0x*** --rpc https://eth-mainnet.g.alchemy.com/v2/***   --code $(jq -r '.deployedBytecode.object' out/erc20.sol/ERC20_2.json)  --show-reachable-tree 1>src/hevm/models/s.lisp 
```

msooseth:
Confirmed, I get the same error on main, and also on `fix-fetching-rpc` (a PR that's pending):

```
$ cabal run hevm  -- symbolic --caller 0xaba --rpc $ETH_RPC_URL  --show-reachable-tree --code $(jq -r '.deployedBytecode.object' tmp2/out/bug.sol/ERC20_2.json)                                                                                                                                             impure 
   WARNING: hevm was only able to partially explore the call prefix 0xunknown due to the following issue(s):
     - Unexpected Symbolic Arguments to Opcode
       msg: "trying to access a symbolic address that isn't already present in storage"
       program counter: 73
       arguments:
         (SymAddr "caller")

QED: No reachable property violations discovered

=== Reachable Expression ===

(ITE
  (IsZero
    TxValue
  )
  (ITE
    (LT
      (BufLength
        (AbstractBuf "txdata")
      )
      4
    )
    (Failure
      Error:
        (Revert
          (ConcreteBuf "")
        )
      Assertions:
        (PNeg
          (PLEq
            4
            (BufLength
              (AbstractBuf "txdata")
            )
          )
        )
        (PEq
          TxValue
          0
        )
        (PLT
          (BufLength
            (AbstractBuf "txdata")
          )
          18446744073709551616
        )
    )
    (ITE
[...]

PR #526, opened 2024-08-12
Respect `--caller` option in `symbolic` mode

msooseth:
## Description

As per #524 `--caller` was not respected here:

`cabal run hevm  -- symbolic --caller 0xabab --rpc $ETH_RPC_URL  --show-reachable-tree --code $(jq -r '.deployedBytecode.object' tmp2/out/bug.sol/ERC20_2.json`

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -515,7 +515,9 @@ symvmFromCommand cmd calldata = do
                              )
 
   let
-    caller = SymAddr "caller"
+    caller = case cmd.caller of
+               Nothing -> SymAddr "caller"
+               Just a -> LitAddr a
     ts = maybe Timestamp Lit cmd.timestamp
     callvalue = maybe TxValue Lit cmd.value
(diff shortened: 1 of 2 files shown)


cli/cli.hs
```diff
@@ -515,7 +515,9 @@ symvmFromCommand cmd calldata = do
                              )
 
   let
-    caller = SymAddr "caller"
+    caller = case cmd.caller of
+               Nothing -> SymAddr "caller"
+               Just a -> LitAddr a
```
blishko, line 520: This is good, but to be consistent with the style of the code (see the next couple of lines) we could write the following?
```suggestion
    caller = maybe (SymAddr "caller") LitAddr cmd.caller
```

blishko (commented):
The fix is good!
What do you think about my suggestions?

msooseth:
Wow, nice!!!

PR #532, opened 2024-08-20
Rename method to better reflect its functionality and add tests

blishko:
## Description
This is a cleanup after #530. We agreed the name of the function does not really match what it is doing, so here is my suggestion for a better name.

- [X] tested locally
- [X] added automated tests


msooseth (approved):
Perfect, thanks!

PR #537, opened 2024-08-25
Refactor simplifications for bitwise AND and OR

### LINKED ISSUE #531: Constant propagation is not working
When running sc-eth-comp, the benchmark `./src/unsafe/ds-test/synthetic-manybranch.sol.SyntheticManyBranch/` generates a file `./src/unsafe/ds-test/synthetic-manybranch.sol.SyntheticManyBranch/query-172-abstracted.smt2` when running with `--dumpsmt`, and that has nonsense like this:
```
;(PEq
;  (And
;    1
;    2
;  )
;  2
;)
```

Which is clearly False: `1&2` is bitwise `&` so that's a `1` and that's not equal to `2` -- this query should not have been dispatched to SMT at all.

blishko:
## Description
We unify handling simplification of bitwise AND and OR with other operations, such as bit shifts and aritmetic operations. After handling some specific cases, we defer to `and` and `or` functions from EVM.Expr, which take care of simplifying the operation when both operators are concrete.

Fixes #531.

## Checklist

- [X] tested locally
- [X] added automated tests

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1109,22 +1109,18 @@ simplify e = if (mapExpr go e == e)
       | otherwise = o
 
     -- Bitwise AND & OR. These MUST preserve bitwise equivalence
-    go o@(And (Lit x) _)
-      | x == 0 = Lit 0
-      | otherwise = o
-    go o@(And v (Lit x))
-      | x == 0 = Lit 0
-      | x == maxLit = v
-      | otherwise = o
-    go (And a b) | a == b = a
-    go (And a (Not b)) | a == b = Lit 0
-    go o@(Or (Lit x) b)
-      | x == 0 = b
-      | otherwise = o
-    go o@(Or a (Lit x))
-      | x == 0 = a
-      | otherwise = o
-    go (Or a b) | a == b = a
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -1109,22 +1109,18 @@ simplify e = if (mapExpr go e == e)
...
+      | b == (Not a) || a == (Not b) = Lit 0
+      | a == (Lit 0) || b == (Lit 0) = Lit 0
+      | a == (Lit maxLit) = b
+      | b == (Lit maxLit) = a
+      | otherwise = EVM.Expr.and a b
+    go (Or a b)
+      | a == b = a
+      | a == (Lit 0) = b
+      | b == (Lit 0) = a
+      | otherwise = EVM.Expr.or a b
```
blishko, line 1123: There are several ways how to implement this, but I thought it would be good to follow the style set by other operations, such as `Sub`, `Add`, `SHL` `SHR`, etc.

msooseth:
Nice! Let me approve and merge. BTW, for more serious changes, let's try to add a line to the changelog, too. This is relatively minor, so likely not worth an entry.

PR #542, opened 2024-08-27
Debug flag should dump internal expressions

msooseth:
## Description
As discovered by @blishko the system doesn't dump internal expressions with `--debug` is given. Let's fix that.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
Internal expressions are useful for debugging, I think this makes sense!

PR #545, opened 2024-08-27
Use simplifying constructor for PEq

blishko:
## Description
The idea is to always normalize Prop equalities such that concrete values appear on the left-hand-side.

This follows the example of the helper functions for expressions, that also immediately simplify if they get concrete arguments, and normalize arguments for commutative operations.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Traversals.hs
+++ b/src/EVM/Traversals.hs
@@ -211,7 +211,7 @@ foldExpr f acc expr = acc <> (go expr)
 mapProp :: (forall a . Expr a -> Expr a) -> Prop -> Prop
 mapProp f = \case
   PBool b -> PBool b
-  PEq a b -> PEq (mapExpr f (f a)) (mapExpr f (f b))
+  PEq a b -> mkPEq (mapExpr f (f a)) (mapExpr f (f b))
   PLT a b -> PLT (mapExpr f (f a)) (mapExpr f (f b))
   PGT a b -> PGT (mapExpr f (f a)) (mapExpr f (f b))
   PLEq a b -> PLEq (mapExpr f (f a)) (mapExpr f (f b))
@@ -612,7 +612,7 @@ mapPropM f = \case
   PEq a b -> do
     a' <- mapExprM f a
     b' <- mapExprM f b
-    pure $ PEq a' b'
+    pure $ mkPEq a' b'
   PLT a b -> do
     a' <- mapExprM f a
     b' <- mapExprM f b
--- a/src/EVM/Types.hs
+++ b/src/EVM/Types.hs
@@ -432,6 +432,14 @@ data Prop where
   PBool :: Bool -> Prop
 deriving instance (Show Prop)
 
+mkPEq :: (Typeable a) => Expr a -> Expr a -> Prop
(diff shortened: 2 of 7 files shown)


src/EVM/Types.hs
```diff
@@ -432,6 +432,14 @@ data Prop where
   PBool :: Bool -> Prop
 deriving instance (Show Prop)
 
+mkPEq :: (Typeable a) => Expr a -> Expr a -> Prop
```
blishko, line 435: Not sure what would be the best name.
I originally had `peq`, to follow the example of the helper functions for `Expr`, but I don't like that very much. I think that `mkPEq` better conveys what the function is doing. We could also use full `makePEq`.
@msooseth, opinions?

src/EVM/Traversals.hs
```diff
@@ -211,7 +211,7 @@ foldExpr f acc expr = acc <> (go expr)
 mapProp :: (forall a . Expr a -> Expr a) -> Prop -> Prop
 mapProp f = \case
   PBool b -> PBool b
-  PEq a b -> PEq (mapExpr f (f a)) (mapExpr f (f b))
+  PEq a b -> mkPEq (mapExpr f (f a)) (mapExpr f (f b))
```
msooseth, line 214: This is just a traversal, we should not change things here.

src/EVM/Traversals.hs
```diff
@@ -612,7 +612,7 @@ mapPropM f = \case
   PEq a b -> do
     a' <- mapExprM f a
     b' <- mapExprM f b
-    pure $ PEq a' b'
+    pure $ mkPEq a' b'
```
msooseth, line 615: This is just a traversal, we should not change things here.

src/EVM/Types.hs
msooseth: This sounds like a good name to me.

msooseth (changes requested):
I think traversals should not change things. Otherwise it's actually really good! I also checked, and you left no PEq constructor untouched :)

src/EVM/Traversals.hs
blishko: But then this could break the invariant that were are trying to establish, no?
Because it can transform the second argument to `Lit` and then we would again end up with `PEq` with constant on RHS instead of LHS.

msooseth: Oh, actually, right. I find it a bit strange that it'd change the thing on its own, but at the same time it'd be painful to have to deal with it every time otherwise... So yeah, you are right. Sorry!

msooseth:
OK, merging. Let's make sure to add a CHANGELOG entry next time, though! Also, we should add a fuzz test for this as well as the Expr LHS/RHS ;)

msooseth:
Thanks so much! Really good stuff! We should have done this ages ago!

src/EVM/Traversals.hs
blishko: I guess this is a design question. I could imagine different approaches. Just thought that we would really like to establish this invariant :)

PR #546, opened 2024-08-28
Do not simplify equality immediately on construction

blishko:
## Description
As @zoep pointed out it might be preferable to keep simplification code in the simplification functions and do not simplify equalities directly at construction point.

However, we would still like to normalize equalities, so that constants are at LHS. Therefore we update the simplification rules, which were previously written to expect constant on RHS.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1604,7 +1604,7 @@ cheatActions = Map.fromList
 
   , action "assume(bool)" $
       \sig _ _ input -> case decodeStaticArgs 0 1 input of
-        [c] -> modifying #constraints ((:) (mkPEq (Lit 1) c))
+        [c] -> modifying #constraints ((:) (PEq (Lit 1) c))
         _ -> vmError (BadCheatCode sig)
 
   , action "roll(uint256)" $
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -19,6 +19,7 @@ import Data.List
 import Data.Map.Strict qualified as Map
 import Data.Maybe (mapMaybe, isJust, fromMaybe)
 import Data.Semigroup (Any, Any(..), getAny)
+import Data.Typeable
 import Data.Vector qualified as V
 import Data.Vector (Vector)
 import Data.Vector.Mutable qualified as MV
@@ -199,6 +200,17 @@ sar = op2 SAR (\x y ->
        if msb then maxBound else 0
      else
        fromIntegral $ shiftR asSigned (fromIntegral x))
+
+
+-- Props
+
+peq :: (Typeable a) => Expr a -> Expr a -> Prop
+peq (Lit x) (Lit y) = PBool (x == y)
+peq a@(Lit _) b = PEq a b
+peq a b@(Lit _) = PEq b a -- we always put concrete values on LHS
+peq a b
+  | a == b = PBool True
+  | otherwise = PEq a b
 
 -- ** Bufs ** --------------------------------------------------------------------------------------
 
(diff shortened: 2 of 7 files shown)


zoep (commented):
Thanks Martin! This looks great. Just a small nit: 

Since the logic for argument normalisation for `Expr` operators is in Expr.hs and already has a naming convention, it would make sense to have the same operators for `Prop` in the same file and with the same naming convention. Therefore, I would suggest moving `mkPEq` to Expr.hs and renaming it to `peq` for consistency.

src/EVM/Expr.hs
```diff
@@ -1214,29 +1214,29 @@ simplifyProp prop =
     go (PLEq a (Max _ b)) | a == b = PBool True
     go (PLEq (Sub a b) c) | a == c = PLEq b a
     go (PLT (Max (Lit a) b) (Lit c)) | a < c = PLT b (Lit c)
-    go (PLT (Lit 0) (Eq a b)) = mkPEq a b
+    go (PLT (Lit 0) (Eq a b)) = PEq a b
```
msooseth, line 1217: I don't understand this... it's under `simplifyProp`, why not do `mkPEq` here? It's the most obvious thing to do I think? I must be not getting something. Under `simplify`, which simplifies `Expr`, we have:

```
    go (IsZero (IsZero (Eq x y))) = eq x y
```

Which is exactly the same idea -- `eq` will set the LHS, etc. So we seem to be inconsistent.

I don't mind it too much, but I just don't understand. Can you maybe explain what's the rationale and why is that `eq` ok, but this `mkPEq` not ok?

src/EVM/Expr.hs
```diff
@@ -1256,21 +1256,19 @@ simplifyProp prop =
...
+    go (PEq (Lit 0) (IsZero (GT a b))) = PGT a b
+    go (PEq (Lit 0) (IsZero (LEq a b))) = PLEq a b
+    go (PEq (Lit 0) (IsZero (GEq a b))) = PGEq a b
 
     -- Eq
-    go (PEq (Eq a b) (Lit 0)) = PNeg (mkPEq a b)
-    go (PEq (Eq a b) (Lit 1)) = mkPEq a b
-    go (PEq (Sub a b) (Lit 0)) = mkPEq a b
-    go (PEq (LT a b) (Lit 0)) = PLEq b a
-    go (PEq (Lit l) (Lit r)) = PBool (l == r)
```
msooseth, line 1270: These are great, thank you so much!

src/EVM/SymExec.hs
```diff
@@ -101,7 +101,7 @@ extractCex (Cex c) = Just c
 extractCex _ = Nothing
 
 bool :: Expr EWord -> Prop
-bool e = POr (mkPEq (Lit 1) e) (mkPEq (Lit 0) e)
+bool e = POr (PEq (Lit 1) e) (PEq (Lit 0) e)
```
msooseth, line 104: This I understand, actually, it makes sense not to do it _before_ we simplify!

src/EVM/SymExec.hs
```diff
@@ -420,7 +420,7 @@ getExpr solvers c signature' concreteArgs opts = do
 checkAssertions :: [Word256] -> Postcondition s
 checkAssertions errs _ = \case
   Failure _ _ (Revert (ConcreteBuf msg)) -> PBool $ msg `notElem` (fmap panicMsg errs)
-  Failure _ _ (Revert b) -> foldl' PAnd (PBool True) (fmap (PNeg . mkPEq b . ConcreteBuf . panicMsg) errs)
+  Failure _ _ (Revert b) -> foldl' PAnd (PBool True) (fmap (PNeg . PEq b . ConcreteBuf . panicMsg) errs)
```
msooseth, line 423: :+1:

src/EVM/SymExec.hs
```diff
@@ -483,7 +483,7 @@ flattenExpr = go []
   where
     go :: [Prop] -> Expr End -> [Expr End]
     go pcs = \case
-      ITE c t f -> go (PNeg ((mkPEq (Lit 0) c)) : pcs) t <> go (mkPEq (Lit 0) c : pcs) f
+      ITE c t f -> go (PNeg ((PEq (Lit 0) c)) : pcs) t <> go (PEq (Lit 0) c : pcs) f
```
msooseth, line 486: :+1:

src/EVM/Types.hs
```diff
@@ -460,9 +460,9 @@ x .>= y = PGEq x y
 
 infix 4 .==, ./=
 (.==) :: (Typeable a) => Expr a -> Expr a -> Prop
-x .== y = mkPEq x y
+x .== y = PEq x y
 (./=) :: (Typeable a) => Expr a -> Expr a -> Prop
-x ./= y = PNeg (mkPEq x y)
+x ./= y = PNeg (PEq x y)
```
msooseth, line 465: :+1:

src/EVM/Expr.hs
```diff
@@ -1214,29 +1214,29 @@ simplifyProp prop =
...
 
     -- iszero(a) -> (a == 0)
     -- iszero(a) == 0 -> ~(a == 0)
     -- ~(iszero(a) == 0) -> ~~(a == 0) -> a == 0
-    go (PNeg (PEq (IsZero a) (Lit 0))) = mkPEq (Lit 0) a
+    go (PNeg (PEq (Lit 0) (IsZero a))) = PEq (Lit 0) a
 
     -- a < b == 0 -> ~(a < b)
     -- ~(a < b == 0) -> ~~(a < b) -> a < b
-    go (PNeg (PEq (LT a b) (Lit 0x0))) = PLT a b
```
msooseth, line 1239: Soooo good thanks!! :+1:

src/EVM.hs
```diff
@@ -1604,7 +1604,7 @@ cheatActions = Map.fromList
 
   , action "assume(bool)" $
       \sig _ _ input -> case decodeStaticArgs 0 1 input of
-        [c] -> modifying #constraints ((:) (mkPEq (Lit 1) c))
+        [c] -> modifying #constraints ((:) (PEq (Lit 1) c))
```
msooseth, line 1607: :+1:

src/EVM/Expr.hs
```diff
@@ -1214,29 +1214,29 @@ simplifyProp prop =
...
     -- iszero(iszero(a)) == 0 -> ~~(a == 0) -> a == 0
     -- ~(iszero(iszero(a)) == 0) -> ~~~(a == 0) -> ~(a == 0) -> a > 0
-    go (PNeg (PEq (IsZero (IsZero a)) (Lit 0))) = PLT (Lit 0) a
+    go (PNeg (PEq (Lit 0) (IsZero (IsZero a)))) = PLT (Lit 0) a
 
     -- iszero(a) -> (a == 0)
     -- iszero(a) == 0 -> ~(a == 0)
     -- ~(iszero(a) == 0) -> ~~(a == 0) -> a == 0
-    go (PNeg (PEq (IsZero a) (Lit 0))) = mkPEq (Lit 0) a
+    go (PNeg (PEq (Lit 0) (IsZero a))) = PEq (Lit 0) a
```
msooseth, line 1235: I'd change this to `mkPEq`

src/EVM/Expr.hs
blishko: I agree that here it would be good to simplify immediately. Do you agree as well, @zoep?

zoep: I agree that for consistency it would be better to use `peq`/`mkPEq` here (though I understand that it doesn't matter so much as it will be simplified in the next iteration)

msooseth (approved):
I'm fine either way, using `mkPEq` in `simplifyProps` would make sense I think, but it's fine either way! It's better not to simplify during creation, it's very true!

zoep (approved):
LGTM!

msooseth:
Perfect, thanks, merging!

PR #548, opened 2024-09-04
Bump nixpkgs and ethereum tests, use cabal-install from nix

arcz:
## Description
We can now use cabal-install 3.12 from nix

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,35 +1,18 @@
 {
   "nodes": {
-    "cabal-3-12": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1719273538,
-        "narHash": "sha256-/TECd3cdqGDdlEur88H0KbV7jcE/czc0jecj9921T5c=",
-        "owner": "haskell",
-        "repo": "cabal",
-        "rev": "260ecdc3d848782d4df49e629cb0a5dc9e96ca9e",
-        "type": "github"
-      },
-      "original": {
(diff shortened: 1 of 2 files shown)


msooseth:
Hahah, I just started working on this, so great!! Thanks, let me see if it goes through, if it does, we'll be golden and hopefully it'll fix the issues we are having with the tstore/mcopy PR!! Thanks so much!

msooseth:
BTW, do you know if we still need this?

```
        # workaround for nixpkgs / ghc / macos / cc issue
        # https://gitlab.haskell.org/ghc/ghc/-/issues/23138
        cc-workaround-ghc-23138 =
          pkgs.writeScriptBin "cc-workaround-ghc-23138" ''
            if [ "$1" = "--print-file-name" ] && [ "$2" = "c++" ]; then
                echo c++
            else
                exec cc "$@"
            fi
          '';
```

What they seem to be saying is that:

```
FYI nixpkgs has now been patched: https://github.com/NixOS/nixpkgs/pull/317224
```

What do you think?

arcz:
I'll check if it's still needed

msooseth:
Well, at least I can reproduce it, which is nice. Lexi says it may be due to static vs dynamic compilation using different GHC.

PR #554, opened 2024-09-16
Less crazy long ABI fuzz types

msooseth:
## Description
This is to mitigate the occasionally long runtimes due to ABI encoding fuzzing:

Related to https://github.com/ethereum/hevm/issues/553

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/ABI.hs
+++ b/src/EVM/ABI.hs
@@ -547,11 +547,11 @@ genAbiValue = \case
    AbiStringType ->
      AbiString . BS.pack <$> listOf arbitrary
    AbiArrayDynamicType t ->
-     do xs <- listOf1 (scale (`div` 2) (genAbiValue t))
+     do xs <- listOf1 (scale (`div` 3) (genAbiValue t))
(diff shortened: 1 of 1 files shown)


msooseth:
All that use the `genAbiValue` in test.hs will be affected. As for `scale`, see: https://hackage.haskell.org/package/QuickCheck-2.15.0.1/docs/Test-QuickCheck.html#v:scale Basically, it adjusts the size parameter as we do dynamically packed ABI structs. So array can be say 10 long, but the array inside the array only 5, and the array inside the array inside the array only 2. The point is not to have infinite size ;) Exponential decay helps, but we can pump up the exponent.

msooseth:
Hmmm MCOPY is getting on my nerves, it's doing a CallDataCopy + MCOPY in fuzzing, which leads to a symbolic sized copy operation, which is upsetting the tests passing...

blishko (approved):

Issue #552, opened 2024-09-12
Investigate possible ways to make our test suite run faster


blishko:
Currently the test suite takes around 75 seconds to run on my laptop.

blishko:
I just want to understand what the test actually do. I didn't even know about all these options!
Is this documented somewhere?

I think I am using `cabal run test`.

msooseth:
I actually just discovered that GHC finally started doing things _actually_ parallel! Kinda new and exciting. I am fixing the mult-thread bug so we can run it 3x for free ;) Give me a sec...

msooseth:
Do you think we can close this now? I think the `cabal run -f devel test` fixes this, right?

blishko:
Yes, we can close this now.

PR #555, opened 2024-09-16
Trying to fix multi-threaded test running

msooseth:
## Description
This is to make sure we can run `cabal run -f devel test`, i.e. multi-threaded `test` running. However, some tests use file IO (mainly tracing(?)) so they need to be sequential or they fail. This is to speed up testing as per #552 

Unfortunately, 1.5 of Tasty has not come to nix yet, so `sequentialTestGroup` is not yet available: https://hackage.haskell.org/package/tasty Hence using `after`.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `initial-storage` option of `hevm symbolic` is respected
 - `caller` option of `hevm symbolic` is now respected
 * Thanks to the new simplification rules, we can now enable more conformance tests
+* Multi-threaded running of Tracing.hs was not possible due to IO race. Fixed.
 
 ## [0.53.0] - 2024-02-23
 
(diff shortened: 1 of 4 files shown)


blishko (approved):

msooseth:
Turns out there was a bug, too. Kinda sad and nice at the same time! `concurrently` was not valid in that context, it seems.

src/EVM/SymExec.hs
```diff
@@ -283,7 +283,8 @@ interpret fetcher maxIter askSmtIters heuristic vm =
         evalRight <- toIO $ do
           (rb, vmb) <- liftIO $ stToIO $ runStateT (continue False) vm { result = Nothing }
           interpret fetcher maxIter askSmtIters heuristic vmb (k rb)
-        (a, b) <- liftIO $ concurrently evalLeft evalRight
+        a <- liftIO evalLeft
+        b <- liftIO evalRight
```
blishko, line 287: Just marking this as THE important change.
Apparently, the parallel run was never working properly?

blishko (approved):
LGTM!

src/EVM/SymExec.hs
msooseth: Yes, that's my suspicion. Because the ` -threaded -with-rstopts=-N` was never passed, so this never ran concurrently, until I passed those via `-f devel`, and pooof it then triggered. There is some kind of race condition there for sure. With this change, it never fails.

PR #558, opened 2024-09-17
Respect request to run fewer quickcheck tests

blishko:
## Description
Previosuly, the property test "abi-encoding-vs-solidity-function-pointer" would always generate 20 tests, regardless of the number tests we actually asked to be generated.
It is OK to specify upper bound, but it would be nice to honor a request to generate fewer tests.

## Checklist

- [X] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Wow nice, you managed to find how  to do it!! Wow.

PR #560, opened 2024-09-18
SymExec: fix data race for concurrent exploration

d-xo:
## Description

Fixes the data race due to the sharing of a `vm` instance to both branches of a call to `concurrently` in the symbolic interpreter. The race was caused by the sharing of mutable memory between threads, fixed by copying memory before passing the vm to each branch.

You can check the fix with the following, which should quickly fail on `main`, but run fine on this branch:

```
while cabal run -f devel test -- -p '/access-out-of-bounds-array/'; do :; done
```

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE DeriveAnyClass #-}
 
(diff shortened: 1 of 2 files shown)


arcz:
I think we still need to freeze memory inside frames

msooseth (approved):
lgtm

d-xo:
> I think we still need to freeze memory inside frames

Ah yes of course, nice catch. Should be fixed now.

src/EVM/SymExec.hs
```diff
@@ -248,6 +250,25 @@ loadSymVM x callvalue cd create =
...
+      mem' <- freeze frame.state.memory
+      pure $ (frame :: Frame Symbolic RealWorld) { state = frame.state { memory = mem' } }
+
+    pure (vm :: VM Symbolic RealWorld)
+      { state = state'
+      , frames = frames'
+      }
+  where
+    freeze = \case
+      ConcreteMemory m -> SymbolicMemory . ConcreteBuf . BS.pack . VUnboxed.toList <$> VUnboxed.freeze m
```
d-xo, line 269: This is pretty much a copy paste of `freezeMemory` from `EVM.hs`, but I couldn't figure out how to generalize the type signature there to make it work in both contexts...

Issue #240, opened 2023-03-27
[Bug-Candidate]: Prank uses wrong account when sending ether to payable function


rappie:
I'm using Echidna, which uses HEVM. When using `prank` with a `msg.value`, the balance of the contract making the call is used instead of the address being pranked.

See also: https://github.com/crytic/echidna/issues/998

### Code example to reproduce the issue:

```solidity
function debugStake() public {
        address from = 0x5409ED021D9299bf6814279A6A1411A7e866A631;

        Debugger.log("before balance this", address(this).balance);
        Debugger.log("before balance from", from.balance);

        hevm.prank(from);
        try safEth.stake{value: 1}() {
            Debugger.log("stake success");
        } catch {
            Debugger.log("stake failed");
        }

        Debugger.log("after balance this", address(this).balance);
        Debugger.log("after balance from", from.balance);

        assert(false);
    }
``

### Relevant log output:
![image](https://user-images.githubusercontent.com/1430820/227943868-b61ecd31-db80-46d1-9827-e6add469d3d1.png)

PR #564, opened 2024-09-20
EVM: respect vm.prank during calls to create

d-xo:
## Description

Updates our `prank` implementation to match foundrys. `prank` is now respected during contract creation. 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


test/test.hs
```diff
@@ -1151,6 +1151,35 @@ tests = testGroup "hevm"
...
+                function f() external {
+                  Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+
+                  Owned target = new Owned();
+                  assert(target.owner() == address(this));
+
+                  address usr = address(1312);
+                  vm.prank(usr);
+                  target = new Owned();
+                  assert(target.owner() == usr);
```
d-xo, line 1176: We should check that it changes back afterwards...

src/EVM.hs
```diff
@@ -834,13 +834,18 @@ exec1 = do
...
                 availableGas <- use (#state % #gas)
-                let
-                  (cost, gas') = costOfCreate fees availableGas xSize False
-                newAddr <- createAddress self this.nonce
+                let (cost, gas') = costOfCreate fees availableGas xSize False
+
+                -- handle `prank`
+                let from' = fromMaybe self vm.config.overrideCaller
+                resetCaller <- use (#config % #resetCaller)
+                when (resetCaller) $ assign (#config % #overrideCaller) Nothing
```
blishko, line 842: Can you explain for me what these two lines do?

msooseth (approved):
lgtm

blishko:
I think the current state of this PR is that the original issue has been fixed, but the `badVault` test still exhibits some unexpected behaviour. Is that right, @d-xo?

src/EVM.hs
msooseth: These are no longer there. But they reset the caller after `prank`. Only when `startPrank` should the caller remain the pranked caller. I am closing because this has been fixed as part of the #240 issue. Let me know if you have any questions, I'd be happy to answer. It's perhaps worthwhile reading the associated PR that closed that issue.

msooseth:
This should now pass all the way through. Can you please review @elopez and/or @blishko ?;

src/EVM.hs
```diff
@@ -852,13 +852,18 @@ exec1 = do
...
                 availableGas <- use (#state % #gas)
-                let
-                  (cost, gas') = costOfCreate fees availableGas xSize False
-                newAddr <- createAddress self this.nonce
+                let (cost, gas') = costOfCreate fees availableGas xSize False
+
+                -- handle `prank`
+                let from' = fromMaybe self vm.state.overrideCaller
+                resetCaller <- use (#state % #resetCaller)
+                when resetCaller $ assign (#state % #overrideCaller) Nothing
```
elopez, line 860: just for consistency

```suggestion
                when resetCaller $ do
                  assign (#state % #overrideCaller) Nothing
                  assign (#state % #resetCaller) False
```

src/EVM.hs
```diff
@@ -852,13 +852,18 @@ exec1 = do
...
+
+                newAddr <- createAddress from' this.nonce
```
elopez, line 862: should the nonce be from's ? it might not be that important though if it's only off during pranking.

test/contracts/fail/10_BadVault.sol
```diff
@@ -166,39 +175,4 @@ contract BadVaultTest is Test {
         // console2.log("attacker final balance", address(attacker).balance);
         assert(attacker.balance <= STARTING_BALANCE);
     }
-
```
elopez, line 169: why is the test being removed?

test/test.hs
```diff
@@ -1231,7 +1231,36 @@ tests = testGroup "hevm"
...
```
elopez, line 1256: can we add a check to ensure the reset works correctly on creates?

```suggestion
                  assert(target.owner() == usr);
                  target = new Owned();
                  assert(target.owner() == address(this));
```

test/test.hs
```diff
@@ -1483,7 +1512,7 @@ tests = testGroup "hevm"
     , test "Loop-Tests" $ do
         let testFile = "test/contracts/pass/loops.sol"
         runSolidityTestCustom testFile "prove_loop" Nothing (Just 10) False Nothing Foundry  >>= assertEqualM "test result" True
-        runSolidityTestCustom testFile "prove_loop" Nothing (Just 100) False Nothing Foundry >>= assertEqualM "test result" False
+        runSolidityTestCustom testFile "prove_loop" Nothing (Just 101) False Nothing Foundry >>= assertEqualM "test result" False
```
elopez, line 1515: is this related to this PR?

test/contracts/pass/cheatCodes.sol
```diff
@@ -65,8 +65,10 @@ contract CheatCodes is Test {
 
     function prove_warp_symbolic(uint128 jump) public {
         uint pre = block.timestamp;
-        hevm.warp(block.timestamp + jump);
-        assertEq(block.timestamp, pre + jump);
+        unchecked {
+          hevm.warp(block.timestamp + jump);
+          assertEq(block.timestamp, pre + jump);
+        }
```
elopez, line 71: are these changes related to this PR?

test/contracts/fail/10_BadVault.sol
msooseth: No idea. I didn't do it :D I can add it back.

msooseth: OK, added back the test!

test/contracts/pass/cheatCodes.sol
msooseth: Nope, it's because of `--via-ir`, which is not needed.

test/test.hs
msooseth: Nope, due to `--via-ir`, removing! Thanks for this!

src/EVM.hs
msooseth: I am actually not sure. This: https://book.getfoundry.sh/cheatcodes/prank says it: `Sets msg.sender to the specified address for the next call.` does not talk about nonce. No idea what that means.

msooseth (commented):
@elopez thanks for the feedback! I made it a lot more lean :) I am unsure about the nonce thing. Do you know how we could check?

src/EVM.hs
msooseth: OK, we decided to leave this be for the moment.

PR #567, opened 2024-09-23
Cheatcodes as per new std-forge library

msooseth:
## Goal

The way forge-std is doing things has changed. It no longer sets a value in ds-test, but instead `Revert`-s with a specific message. So we have to match:

```
forge test --match-test "test_set_get()" -vvvv
[...]
  [31205] MateFailContractTest::test_set_get()
    ├─ [22216] MateFailContract::set_mate()
    │   └─ ← [Stop]
    ├─ [268] MateFailContract::get_mate() [staticcall]
    │   └─ ← [Return] 3
    ├─ [0] VM::assertEq(2, 3) [staticcall]
    │   └─ ← [Revert] assertion failed: 3 != 2
    └─ ← [Revert] assertion failed: 3 != 2
```

Instead of the current:

```
Traces:
  [42786] MateFailContractTest::test_set_get()
    ├─ [22216] MateFailContract::set_mate()
    │   └─ ← [Stop]
    ├─ [268] MateFailContract::get_mate() [staticcall]
    │   └─ ← [Return] 3
    ├─ emit log(val: "Error: a == b not satisfied [uint]")
    ├─ emit log_named_uint(key: "      Left", val: 3)
    ├─ emit log_named_uint(key: "     Right", val: 2)
    ├─ [0] VM::store(VM: [0x7109709ECfa91a80626fF3989D68f67F5b1DD12D], 0x6661696c65640000000000000000000000000000000000000000000000000000, 0x0000000000000000000000000000000000000000000000000000000000000001)
    │   └─ ← [Return]
    └─ ← [Stop]
```

The most prominent part of this change is the disappearance of line 202 from `UnitTest.hs`, where we'd check for failure bit being set in the cheat contract (i.e. `0x7109709EC...`)

## Implementation

* Emilio has changed things around already in `EVM.hs` to have a separate frame, which allows this to be coded in a relatively compact way. I have now added a number of required (but not all) cheatcodes. 
* Moved all relevant test cases to `std-forge/Test.hs` and `{Test}` from `ds-test/test.hs` and `{DSTest}`
* Added most of the required `assert...` methods, though some are still missing. The ones still missing would require dynamic ABI handling, such as `assertTrue(uint,string)`, where `string` is dynamic.
* Fixed the function signature generation for types such as tuples. Now `prove_tuple_fail1(uint256[5] memory amounts)` actually works
* Better debug output. Cleaner and more thorough. It should be a lot easier to debug things.

## Future work

I haven't added:
* `assertEq(T[], T[])` type asserts
* `assertNotEq(T[], T[])` type asserts
* `assertXXDecimal`
* `assertApproxXX
* assertEqCall`

Mostly due to the above mentioned issue with dynamic ABI type issue.

## TODO in a future PR

* Move cheatcodes into a new file
* Would be nice to accumulate errors when e.g. we have trouble with cheatcodes

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1696,7 +1696,7 @@ cheat gas (inOffset, inSize) (outOffset, outSize) xs = do
     Nothing -> partial $ UnexpectedSymbolicArg vm.state.pc (getOpName vm.state) "symbolic cheatcode selector" (wrap [abi])
     Just (unsafeInto -> abi') ->
       case Map.lookup abi' cheatActions of
-        Nothing -> vmError (BadCheatCode "cannot understand cheatcode, maybe cheatcode not supported?" abi')
+        Nothing -> vmError (BadCheatCode "Cannot understand cheatcode. " abi')
         Just action -> action input
 
 type CheatAction t s = Expr Buf -> EVM t s ()
@@ -1931,12 +1931,14 @@ cheatActions = Map.fromList
   , action "assertEq(int256,int256)"   $ assertEq (AbiIntType 256)
   , action "assertEq(address,address)" $ assertEq AbiAddressType
   , action "assertEq(bytes32,bytes32)" $ assertEq (AbiBytesType 32)
+  , action "assertEq(string,string)"   $ assertEq (AbiStringType)
   --
   , action "assertNotEq(bool,bool)"       $ assertNotEq AbiBoolType
   , action "assertNotEq(uint256,uint256)" $ assertNotEq (AbiUIntType 256)
   , action "assertNotEq(int256,int256)"   $ assertNotEq (AbiIntType 256)
   , action "assertNotEq(address,address)" $ assertNotEq AbiAddressType
   , action "assertNotEq(bytes32,bytes32)" $ assertNotEq (AbiBytesType 32)
+  , action "assertNotEq(string,string)"   $ assertNotEq (AbiStringType)
   --
   , action "assertLt(uint256,uint256)" $ assertLt (AbiUIntType 256)
   , action "assertLt(int256,int256)"   $ assertLt (AbiIntType 256)
@@ -2326,14 +2328,10 @@ finishFrame how = do
     nextFrame : remainingFrames -> do
 
       -- Insert a debug trace.
-      insertTrace $
-        case how of
-          FrameErrored e ->
-            ErrorTrace e
-          FrameReverted e ->
-            ErrorTrace (Revert e)
-          FrameReturned output ->
-            ReturnTrace output nextFrame.context
+      insertTrace $ case how of
+        FrameReturned output -> ReturnTrace output nextFrame.context
+        FrameReverted e -> ErrorTrace (Revert e)
+        FrameErrored e -> ErrorTrace e
       -- Pop to the previous level of the debug trace stack.
       popTrace
 
@@ -2384,8 +2382,10 @@ finishFrame how = do
             FrameErrored e -> do
               revertContracts
               revertSubstate
-              assign (#state % #returndata) (ConcreteBuf (BS8.pack $ show e))
-              push 0
+              if (isInterpretFailure e) then finishFrame (FrameErrored e)
+              else do
+                assign (#state % #returndata) mempty
+                push 0
         -- Or were we creating?
         CreationContext _ _ reversion subState' -> do
           creator <- use (#state % #contract)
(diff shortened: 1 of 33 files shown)


src/EVM/UnitTest.hs
```diff
@@ -287,14 +291,6 @@ execSymTest UnitTestOptions{ .. } method cd = do
...
   runExpr
 
-checkSymFailures :: VMOps t => UnitTestOptions RealWorld -> Stepper t RealWorld (VM t RealWorld)
-checkSymFailures UnitTestOptions { .. } = do
-  -- Ask whether any assertions failed
-  Stepper.evm $ do
-    popTrace
-    abiCall testParams (Left ("failed()", emptyAbi))
-  Stepper.runFully
-
```
msooseth, line 297: This is not used anywhere, so I'm removing it

src/EVM.hs
```diff
@@ -1942,6 +1985,77 @@ cheatActions = Map.fromList
     stringToByteString = Right . Char8.pack
     stringHexToByteString :: String -> Either ByteString ByteString
     stringHexToByteString s = either (const $ Left "invalid bytes value") Right $ BS16.decodeBase16Untyped . Char8.pack . strip0x $ s
+    paramDecodeErr abitype name abivals = name <> "(" <> (show abitype) <> "," <> (show abitype) <>
+      ")  parameter decoding failed. Error: " <> show abivals
+    revertErr a b comp = frameRevert $ "assertion failed: " <>
+      BS8.pack (show a) <> " " <> comp <> " " <> BS8.pack (show b)
+    assertEq abitype sig input = do
```
d-xo, line 1992: I'm not super bothered by the duplication but if you wanted to cut some lines you could probably merge the implementations into one function that is parametrised on the comparison function and string used in the revert error?

src/EVM.hs
```diff
@@ -2267,10 +2381,10 @@ finishFrame how = do
               push 0
 
             -- Case 3: Error during a call?
-            FrameErrored _ -> do
+            FrameErrored e -> do
               revertContracts
               revertSubstate
-              assign (#state % #returndata) mempty
+              assign (#state % #returndata) (ConcreteBuf (BS8.pack $ show e))
```
d-xo, line 2387: why do we need to do this for non reverting errors?

src/EVM/UnitTest.hs
```diff
@@ -224,6 +227,12 @@ symRun opts@UnitTestOptions{..} vm (Sig testName types) = do
     -- check postconditions against vm
     (e, results) <- verify solvers (makeVeriOpts opts) (symbolify vm') (Just postcondition)
     let allReverts = not . (any Expr.isSuccess) . flattenExpr $ e
+    let fails = filter (Expr.isFailure) $ flattenExpr e
+    unless (null fails) $ liftIO $ do
+      putStrLn $ "      \x1b[33mWARNING\x1b[0m: hevm was only able to partially explore the test " <> Text.unpack testName <> " due to: ";
```
d-xo, line 232: Are failures the same as partial executions?

src/EVM.hs
msooseth: Ahhhh good point! Thanks, fixed :)

src/EVM.hs
msooseth: This has  been removed, thanks!!

msooseth:
This is now ready to re-review, @d-xo  and maybe @blishko in case you want to have a go at it :)

src/EVM/UnitTest.hs
```diff
@@ -189,30 +193,26 @@ runUnitTestContract
...
+        shouldFail = "proveFail" `isPrefixOf` callSig
 
     -- define postcondition depending on `shouldFail`
-    -- We directly encode the failure conditions from failed() in ds-test since this is easier to encode than a call into failed()
-    -- we need to read from slot 0 in the test contract and mask it with 0x10 to get the value of _failed
-    -- we don't need to do this when reading the failed from the cheatcode address since we don't do any packing there
-    let failed store = case Map.lookup cheatCode store of
-          Just cheatContract -> readStorage' (Lit 0x6661696c65640000000000000000000000000000000000000000000000000000) cheatContract.storage .== Lit 1
-          Nothing -> And (readStorage' (Lit 0) (testContract store).storage) (Lit 2) .== Lit 2
-        postcondition = curry $ case shouldFail of
```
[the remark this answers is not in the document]
d-xo: we might wanna keep this around for a while as a compat mode? maybe with some cli flag `--assertion-mode <ds-test|forge-std>`?

PR #570, opened 2024-09-30
Fix bug in SMT encoding of reading multiple consecutive bytes

blishko:
## Description

The current encoding in function `assertReads` asserted, for a read of more than one (and not 32) bytes, that the same expression is equal to multiple different constants. This caused some problems to be trivially unsatisfiable, possibly hiding other bugs.

The new encoding should hopefully fix that.
After running our test suite, it seems like this actually affected some of our tests, because the overall runtime increased noticeably.

This is a follow up on #569.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -404,12 +404,9 @@ assertReads :: [Prop] -> BufEnv -> StoreEnv -> [Prop]
 assertReads props benv senv = concatMap assertRead allReads
   where
     assertRead :: (Expr EWord, Expr EWord, Expr Buf) -> [Prop]
+    assertRead (_, Lit 0, _) = []
     assertRead (idx, Lit 32, buf) = [PImpl (PGEq idx (bufLength buf)) (PEq (ReadWord idx buf) (Lit 0))]
-    assertRead (idx, Lit sz, buf) =
-      fmap
-        -- TODO: unsafeInto instead fromIntegral here makes symbolic tests fail
-        (PImpl (PGEq idx (bufLength buf)) . PEq (ReadByte idx buf) . LitByte . fromIntegral)
-        [(0::Int)..unsafeInto sz-1]
+    assertRead (idx, Lit sz, buf) = [PImpl (PGEq (Expr.add idx $ Lit offset) (bufLength buf)) (PEq (ReadByte (Expr.add idx $ Lit offset) buf) (LitByte 0)) | offset <- [(0::W256).. sz-1]]
     assertRead (_, _, _) = internalError "Cannot generate assertions for accesses of symbolic size"
 
     allReads = filter keepRead $ nubOrd $ findBufferAccess props <> findBufferAccess (Map.elems benv) <> findBufferAccess (Map.elems senv)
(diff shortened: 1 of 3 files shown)


src/EVM/SMT.hs
```diff
@@ -405,11 +405,7 @@ assertReads props benv senv = concatMap assertRead allReads
   where
     assertRead :: (Expr EWord, Expr EWord, Expr Buf) -> [Prop]
     assertRead (idx, Lit 32, buf) = [PImpl (PGEq idx (bufLength buf)) (PEq (ReadWord idx buf) (Lit 0))]
-    assertRead (idx, Lit sz, buf) =
-      fmap
-        -- TODO: unsafeInto instead fromIntegral here makes symbolic tests fail
-        (PImpl (PGEq idx (bufLength buf)) . PEq (ReadByte idx buf) . LitByte . fromIntegral)
-        [(0::Int)..unsafeInto sz-1]
+    assertRead (idx, Lit sz, buf) = [PImpl (PGEq (Add idx $ Lit offset) (bufLength buf)) (PEq (ReadByte (Add idx $ Lit offset) buf) (LitByte 0)) | offset <- [(0::W256).. sz-1]]
```
blishko, line 408: I was not sure, but we can probably use the the simplyfing `Expr.add` function here, no?

msooseth:
Please update the changelog ;)

src/EVM/SMT.hs
msooseth: I don't think we should... I think it's best to keep it being used only inside simplify and its subfunctions? I dunno, I am afraid that it will make it harder to reason about code if we don't... but I am no authority on this.

msooseth (commented):
So I am confused. The EMPTY buffer should work perfectly fine to read 0 from places we didn't write because of:

```
  ConcreteBuf "" -> "((as const Buf) #b00000000)"
```

This is actually non-standard SMTLib, but works on all BV solvers so far. Shouldn't we use the same for buffers that are not empty? I m not challenging this PR, I am just wondering if we have been doing the right thing for a long time :) I also don't really know how this `as const` works, because it's of course undocumented....

src/EVM/SMT.hs
blishko: But it is being used also outside of that, for example in `EVM.exec1`.
It would just compute the concrete offset value in case `idx` is also concrete.
I was also not sure, but I think it should be OK.

msooseth: Good point. Yeah, let's just do it :) Wanna update the PR?

msooseth: I am always hesitant of side-effects. Of course since it's functional programming, it's not supposed to have side-effects :)

msooseth: OK, I did a commit on top with this, let me know what you think! You can always revert!

msooseth:
I think I have managed to make this use `as const` whenever possible, hence we can skip adding those read constraints on fully constant arrays.

src/EVM/SMT.hs
```diff
@@ -831,7 +834,7 @@ exprToSMT = \case
   ReadByte idx src -> op2 "select" src idx
 
   ConcreteBuf "" -> "((as const Buf) #b00000000)"
-  ConcreteBuf bs -> writeBytes bs mempty
+  ConcreteBuf bs -> writeBytes bs (ConcreteBuf "")
```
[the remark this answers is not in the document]
blishko: Yes, `mempty` means `ConcreteBuf ""`, see https://github.com/ethereum/hevm/blob/d0713cb3b6e74b53af049b953ecf8a37161166d0/src/EVM/Expr.hs#L567-L568

No need for this change.

msooseth (approved):
Actually, good. But I really don't like the 4000 constraints for the 4000 long array :S Maybe no way around it.... :(

PR #573, opened 2024-10-01
Remove buggy specializations of CopySlice simplification

blishko:
## Description

The simplification of fully concrete CopySlice was not always taking into account that slicing beyond the size of the destination buffer should extend the buffer first.

Instead of maintaning multiple specializations, the most general one seems sufficient to cover all cases.
I don't think there is any serious performance penalty and keeping only one version of the simplification lessens the cognitive burden (at least for me) and makes the code less error-prone in my opinion.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -324,37 +324,23 @@ maxBytes = into (maxBound :: Word32) `Prelude.div` 8
 
 copySlice :: Expr EWord -> Expr EWord -> Expr EWord -> Expr Buf -> Expr Buf -> Expr Buf
 
--- Copies from empty buffers
-copySlice _ _ (Lit 0) (ConcreteBuf "") dst = dst
-copySlice a b c@(Lit size) d@(ConcreteBuf "") e@(ConcreteBuf "")
-  | size < maxBytes = ConcreteBuf $ BS.replicate (unsafeInto size) 0
-  | otherwise = CopySlice a b c d e
-copySlice srcOffset dstOffset sz@(Lit size) src@(ConcreteBuf "") dst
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -324,38 +324,24 @@ maxBytes = into (maxBound :: Word32) `Prelude.div` 8
...
-copySlice a b c@(Lit size) d@(ConcreteBuf "") e@(ConcreteBuf "")
-  | size < maxBytes = ConcreteBuf $ BS.replicate (unsafeInto size) 0
-  | otherwise = CopySlice a b c d e
-copySlice srcOffset dstOffset sz@(Lit size) src@(ConcreteBuf "") dst
-  | size < maxBytes = copySlice srcOffset dstOffset (Lit size) (ConcreteBuf $ BS.replicate (unsafeInto size) 0) dst
-  | otherwise = CopySlice srcOffset dstOffset sz src dst
-
--- Fully concrete copies
-copySlice a@(Lit srcOffset) b@(Lit dstOffset) c@(Lit size) d@(ConcreteBuf src) e@(ConcreteBuf "")
-  | srcOffset > unsafeInto (BS.length src), size < maxBytes = ConcreteBuf $ BS.replicate (unsafeInto size) 0
```
blishko, line 338: This first case was incorrect, because it did not take into account the destination offset.
The version with nonempty destination buffer actually handles this case correctly.

msooseth (approved):
Yes, I experimented removing it, too, it just makes it even harder to reason about it.

PR #574, opened 2024-10-01
Adding SAT & UNSAT check both

msooseth:
## Description

This adds both a check for SAT and UNSAT, so we don't end up in a situation that @blishko found

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
LGTM!

PR #576, opened 2024-10-01
Fix gas accounting in cheatcodes

elopez:
## Description

While updating hevm in Echidna, I noticed the gas count going negative when calling the ffi function repeatedly. I suspect this is caused by the gas refund that is issued while returning from a call, as we don't burn any gas when we start running a cheatcode. This adds the missing burn, hopefully fixing the issue. A quick test in echidna appears to exhibit sane behavior with this patch but I'd appreciate a second look at this.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
lgtm

PR #577, opened 2024-10-01
Make copyslice of size 0 noop also in buffer length computation

blishko:
## Description

This is a fixup after #573.
We have agreed that `CopySlice` of size 0 is a noop and returns just the destination buffer, unchanged.
This semantics must also be reflected in the computation of buffer length.


msooseth:
Do you know what will happen if it's zero, but symbolically?

blishko:
I don't know, I guess that depends on the symbolic encoding.
But I suspect it will not work correctly at the moment most likely.

PR #580, opened 2024-10-04
More debug printing during test.hs, when required

msooseth:
## Description
During test.hs, more debug printing, when it's turned on.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -3624,12 +3624,16 @@ checkEquiv a b = do
   if a == b then pure True else do
     when (opts.debug) $ liftIO $ putStrLn $ "Checking equivalence of " <> show a <> " and " <> show b
     x <- checkEquivBase (./=) a b True
+    when (opts.debug) $ liftIO $ putStrLn $ "UNSAT check, expect True: " <> show x
     y <- checkEquivBase (.==) a b False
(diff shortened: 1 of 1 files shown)


test/test.hs
```diff
@@ -3624,12 +3624,16 @@ checkEquiv a b = do
   if a == b then pure True else do
     when (opts.debug) $ liftIO $ putStrLn $ "Checking equivalence of " <> show a <> " and " <> show b
     x <- checkEquivBase (./=) a b True
+    when (opts.debug) $ liftIO $ putStrLn $ "UNSAT check, expect True: " <> show x
     y <- checkEquivBase (.==) a b False
+    when (opts.debug) $ liftIO $ putStrLn $ "SAT check, expect False: " <> show x
```
blishko, line 3629:
```suggestion
    when (opts.debug) $ liftIO $ putStrLn $ "SAT check, expect False: " <> show y
```

blishko (changes requested):

test/test.hs
msooseth: Wow, thanks, that was stupid of me

blishko (approved):
LGTM!

msooseth:
Thanks for the review. Really helped actually!

PR #582, opened 2024-10-07
Fix keccak + sha256 to take 2 params: buf + len

### LINKED ISSUE #578: SLoad + SStore + Keccak generates UNSAT expression
```
    , test "storage-slot-single" $ do
        let x = SLoad (Add (Keccak (ConcreteBuf "b")) (Lit 3)) (SStore (Keccak (ConcreteBuf "\NUL")) (Lit 0x44) (SStore (Keccak (ConcreteBuf "\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL\NUL")) (Lit 0x66) (AbstractStore (SymAddr "stuff") Nothing)))
        let simplified = Expr.simplify x
        y <- checkEquivAndLHS x simplified
        assertBoolM "Must be equal" y
```

Is an UNSAT SMT even when x==simp(x).

Maybe the bug is around Keccak this time. I tried minimizing it, but this is as far as I could go.

msooseth:
This changes Keccak and SHA256 to take 2 parameters: buf + length instead of only a buf. This is as per our call last Friday as the (potential) solution the problem highlighted in #578 

I have also changed the way the number of tests for the Keccak simplification is handled. Before, it was capped at 50. [Default is 100](https://www.reddit.com/r/haskell/comments/51cew3/quickcheck_test_count/). So now I set it to half whatever number is passed, so it'll still be 50 when default is passed, but if a large number is passed, e.g. `--quickcheck-tests 10000` it will run 5000, instead of 50.

Closes #578

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -509,9 +509,11 @@ prelude =  SMT2 src mempty mempty mempty
   ; slot -> value
   (define-sort Storage () (Array Word Word))
 
-  ; hash functions
-  (declare-fun keccak (Buf) Word)
-  (declare-fun sha256 (Buf) Word)
+  ; hash functions. They take buffer and a buffer size, and return a Word
+  (declare-fun keccak (Buf Word) Word)
+  (declare-fun sha256 (Buf Word) Word)
+
+  ; helper functions
   (define-fun max ((a (_ BitVec 256)) (b (_ BitVec 256))) (_ BitVec 256) (ite (bvult a b) b a))
 
   ; word indexing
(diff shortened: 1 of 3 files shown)


src/EVM/SMT.hs
```diff
@@ -788,12 +790,14 @@ exprToSMT = \case
...
     "(ite " <> cond `sp` one `sp` zero <> ")"
-  Keccak a ->
-    let enc = exprToSMT a in
-    "(keccak " <> enc <> ")"
-  SHA256 a ->
-    let enc = exprToSMT a in
-    "(sha256 " <> enc <> ")"
+  Keccak a -> let
+      enc = exprToSMT a
+      sz = exprToSMT $ Expr.simplify $ BufLength a
```
blishko, line 795: Shouldn't we use `bufLength` function from `Expr` here?
We do that when computing injectivity constraints in `injProp` in `Keccak`.
Feels like these should be unified...

msooseth: Ah good point. The `Expr.simplify $BufLength x` effectively does **something** like that. If you look at `Expr.simplify`, line 957:

```haskell
    go (BufLength buf) = bufLength buf
```

it actually calls `Expr.bufLength`. However, `Expr.simplify $BufLength x` calls simplify on `x` as well, recursively.

However, `x` should have been simplified at this point, and should need no simplifixation. And also, we should NOT simplify things here -- some of our fuzzing systems rely on simplification NOT taking place:

See line 3650 in `test.hs`:

```haskell
     let smt = assertPropsNoSimp conf [mkprop l r]
```

If simplification takes place, then it's not possible to check simplified vs unsimplified expressions against each other. Thereby making it impossible to check for simplification bugs.

So I am replacing this with `Expr.bufLength`. Thanks for spotting this!

msooseth: I am closing. Please do reopen in case you think my fix is wrong :)

blishko (approved):
LGTM!

PR #588, opened 2024-10-07
EVM: finalize: fix handling of symbolic returndata

d-xo:
We produced a `Partial` node with a message "runtime code cannot have an abstract length" if we could not determine a concrete size for returndata when finalizing a transaction. This change gates that check so it is only made if we are actually in a create transaction (which is the only context where it is valid).

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Wow

msooseth (approved):
Nice, thanks, a lot cleaner, too! LGTM

PR #590, opened 2024-10-08
Adding definition of equivalence to our documentation

msooseth:
## Description
While reading the hevm channel on Matrix, it became apparent we don't really define what we mean by "equivalent" in hevm. So I had a go at it. I have a feeling it may be not correct -- please review and fix.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog

### DIFF
--- a/doc/src/equivalence-checking-tutorial.md
+++ b/doc/src/equivalence-checking-tutorial.md
@@ -5,6 +5,15 @@ circumstances. This allows to e.g. create two functions, one that is known to be
 another that uses less gas, but is hard to check for correctness. Then, with equivalence
 checking, one can check whether the two behave the same.
 
+The notion of equivalence in hevm is defined as follows. Two contracts are equivalent
+if for all possible inputs, after execution has finished, their observable
+storage state is equivalent and they return the same value. In particular, the
+following is NOT checked when checking for equivalence:
+- [Gas](https://ethereum.org/en/developers/docs/gas/) consumption
+- [Events](https://solidity-by-example.org/events/) emitted
+- Maximum stack depth
+- Maximum memory usage
+
 ## Finding Discrepancies
 
 Let's see this toy contract, in file [contract1.sol](code_examples/contract1.sol):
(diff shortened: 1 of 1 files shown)


doc/src/equivalence-checking-tutorial.md
```diff
@@ -5,6 +5,15 @@ circumstances. This allows to e.g. create two functions, one that is known to be
 another that uses less gas, but is hard to check for correctness. Then, with equivalence
 checking, one can check whether the two behave the same.
 
+The notion of equivalence in hevm is defined as follows. Two contracts are equivalent
+if for all possible inputs, after execution has finished, their observable
+storage state is equivalent and they return the same value. In particular, the
```
blishko, line 10: I am a bit confused here because you say it is a notion of an equivalence of _contracts_, but then you talk about execution, inputs and output.

Doesn't the latter implies that you are talking about equivalence of functions, not contracts?

d-xo: contracts are functions with a signature `(world state, calldata) -> (world state, returndata)`

blishko: OK, yes.
That just clashes with the notion of a contract from Solidity where a contract is a collection of such functions, no?

msooseth: Well, the `world state` encompasses the contract state, so I think it's OK? Or in what way do you mean?

msooseth: On this note, maybe I should explain this in the doc... because it's not obvious at all. Do you wanna have a go at improving this so it's a bit more clearly explained relative to calldata/ABI decoding?

blishko: Oh, I see! Thanks for the explanation!
You are right. The "collection of functions" is just a Solidity abstraction over the jump table.

> Do you wanna have a go at improving this so it's a bit more clearly explained relative to calldata/ABI decoding?

I think it is obvious I am not the right person to be explaining this 🙃 

I guess it depends on who is the target audience. Solidity developers might be confused about this, but people used to think in EVM terms will be fine.

msooseth: OK, thanks. I have now added a few more sentences. Can you please check if that makes it better? Hopefully I can then merge :)

blishko (approved):
Perfect!

PR #593, opened 2024-10-11
Take into account possibly overflowing when reading from buffers

blishko:
When reading from a buffer, it is possible to specify offset that would mean reading beyond the buffer's size. That would normally just mean reading 0. However, when reading multiple consecutive indices, there is an edge case when we would start from a very large offset and would overflow the value 2^256 (which is a boundary given by the yellow paper). When the indices overflow, they start again from the beginning of the buffer.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth (approved):
Nice, it's well-defined finally :) I'm liking it! Thanks :)

PR #595, opened 2024-10-21
Fix simplification of CopySlice when overflowing into too short buffer

blishko:
## Description

The problem was that we did not take into account the need to pad the slice when the source buffer was too short for required size.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Nice, thanks for the testcase, too! LGTM

msooseth (approved):

PR #597, opened 2024-10-23
Revert wraparound semantics for reading from buffers

blishko:
## Description

Recently, we have changed the semantics of reading beyond index `2^256` from a buffer. Previously, any read beyond buffer length would return all zeroes. However, to match the semantics of SMT, we changed that so that if reading from multiple consecutive indices would overflow the index (at `2^256`), the indices would wrap around and start from `0` again. Unfortunately, it turns out that the interpreters (such as evmtool) do not behave this way, they follow the original "just return all zeroes" strategy.

In order to match the behaviour of real interpreters, we revert the recent changes and return to the original behaviour.

This also introduces the problem of incompatibility with our SMT encoding, which should be mitigated by modifying the fuzzer to not generate these corner cases.

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -705,6 +705,11 @@ tests = testGroup "hevm"
         y <- checkEquivAndLHS x simplified
         assertBoolM "Must be equal" y
     ]
+  {- NOTE: These tests were designed to test behaviour on reading from a buffer such that the indices overflow 2^256.
+           However, such scenarios are impossible in the real world (the operation would run out of gas). The problem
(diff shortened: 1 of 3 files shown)


test/test.hs
```diff
@@ -736,13 +741,8 @@ tests = testGroup "hevm"
           s = Expr.simplify e
         equal <- checkEquiv e s
         assertEqualM "Must be equal" True equal
-    , test "copy-slice-overflowing-into-too-short-buffer" $ do
```
blishko, line 739: This test was basically a duplicate of copy-slice-overflow-beyond-source-into-nonempty

PR #598, opened 2024-10-23
Fuzzer: Generate smaller indices in read expressions

blishko:
## Description

When generating expressions that read multiple consecutive bytes (ReadWord and CopySlice expressions), we need to avoid generating indices close to max value of Word256, as these would cause overflow in indices. Such overflows have different semantics in SMT and in EVM, so the fuzzer would report unhelpful failures when fuzzing our simplifier.

The proposed solution is to trim concrete Lit expressions in as read index.

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -32,7 +32,7 @@ import Data.Tuple.Extra
 import Data.Tree (flatten)
 import Data.Typeable
 import Data.Vector qualified as V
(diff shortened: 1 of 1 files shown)


blishko:
`ReadWord` and `CopySlice` :)

msooseth:
Yes, oops!

PR #599, opened 2024-10-23
Flake: cleanup and expose dynamic output

d-xo:
## Description

Cleans up the nix flake as follows:

- conform to modern nixpkgs standards by removing the `with` statements
  and making scoping explicit
- removing `flake-compat` (which was broken on macos due to
  https://github.com/edolstra/flake-compat/issues/53).

Additionally reworks the structure a little and makes the default flake
output dynamically linked, making it easier to consume from other nix
flakes (e.g. act).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -17,22 +17,6 @@
         "type": "github"
       }
     },
-    "flake-compat": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1696426674,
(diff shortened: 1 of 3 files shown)


msooseth (approved):
So nice! On my machine, I will use nix-shell instead of `nix develop`, because it drops me into `zsh` instead of `bash`, but for others, I'll recommend `nix develop`, because it's a lot more stable it seems. Thanks again!

PR #602, opened 2024-10-25
Fix condition in min-distance keccak axiom

blishko:
## Description

If keccak arguments are not the same then the values must be at least 256 apart.
However, the condition of "not same" must include both the buffer *and* its size.

While fixing the condition, I made a cleanup pass over the generation of keccak axioms.
Notably, I have removed the axioms regarding the concrete values in `keccakAssumptions`, because there is another function for that `keccakCompute`, which is better because it tries to simplify the buffer before checking if the result is a concrete buffer.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


d-xo (approved):

msooseth:
Oh wow of course. Thanks!

Issue #608, opened 2024-11-28
hevm `startPrank` overrides msg.sender at all call depth levels, making it inconsistent vs foundry and causing unintended operations


GalloDaSballo:
## Impact

I have written this repo which shows a full repro

https://github.com/Recon-Fuzz/prank-echidna-vs-foundry

Fundamentally, when using `startPrank` and calling a contract that performs a `transferFrom`, echidna (using hevm under the hood), will perform the `transferFrom` with the pranked address instead of the router, causing the transfer to fail

## Mitigation

I believe that the prank should only be enforced at the current call level, a nested call should not longer have it's msg.sender altered

msooseth:
Right, so `overrideCaller` is set from `Nothing` to the address when `startPrank` is called. This also sets `resetCaller` to `False`. Then, when `stopPrank` is called, `overrideCaller` is set to `Nothing`, and `resetCaler` is set to `True`.

So what you are saying would I think mean that when we do another `call`, `overrideCaller` should be reset to `Nothing`, along with `resetCaller` being reset to `True`. Let me make a PR like that, and maybe you can then check if that fixes it. Would that be OK?

msooseth:
So I have a new branch, `fix_prank` that may fix this. I'm trying to check if it does now :)

elopez:
@GalloDaSballo I opened a draft PR on Echidna that uses @msooseth's hevm branch, if you can try it out and see if it behaves as expected now, that'd be great: https://github.com/crytic/echidna/pull/1331

msooseth:
OK, seems to be fixed with `311b4115aed46e3f2db87acaa69fd73b5503bc14`. The corresponding echidna is:

https://github.com/msooseth/echidna/pull/1

You can cone that, go into `fix_prank` branch, then `nix shell`, then `git clone https://github.com/Recon-Fuzz/prank-echidna-vs-foundry` then `cd prank-echidna-vs-foundry` and then `echidna . --contract CryticTester --config echidna.yaml` and it should all pass.

GalloDaSballo:
Have tested this release candidate by @elopez :
https://github.com/crytic/echidna/actions/runs/12081766889

I ran it for 1 Million tests with no reverts
<img width="1129" alt="Screenshot 2024-11-29 at 18 05 39" src="https://github.com/user-attachments/assets/53f94b17-4d7c-47ae-b8eb-2e896aa5d412">


In contrast see the previous version failing in less than 50k tests
[...]

PR #609, opened 2024-11-28
Fixing prank: no longer override the sender on lower frames

msooseth:
## Description
As per #608 there is an issue with prank. It keeps overriding the sender on lower frames as well. This should not happen.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -878,6 +878,7 @@ exec1 = do
                               assign #caller from'
                               assign #contract callee
                               assign #overrideCaller Nothing
+                              assign #resetCaller False                              
                             touchAccount from'
                             touchAccount callee
                             transfer from' callee xValue
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -898,6 +898,7 @@ exec1 = do
                         assign #callvalue xValue
                         assign #caller $ fromMaybe self overrideC
                         assign #overrideCaller Nothing
+                        assign #resetCaller False
                       touchAccount self
             _ ->
               underrun
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -2015,6 +2015,7 @@ delegateCall this gasGiven xTo xContext xValue xInOffset xInSize xOutOffset xOut
                     assign #returndata mempty
                     assign #calldata calldata
                     assign #overrideCaller Nothing
+                    assign #resetCaller False
                   continue xTo
 
 -- -- * Contract creation
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -84,7 +84,7 @@ blankState = do
     , returndata   = mempty
     , static       = False
     , overrideCaller = Nothing
-    , resetCaller  = True
+    , resetCaller  = False
     }
 
 -- | An "external" view of a contract's bytecode, appropriate for
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1820,6 +1820,7 @@ cheatActions = Map.fromList
       \_ _ -> do
         doStop
         assign (#state % #overrideCaller) Nothing
+        assign (#state % #resetCaller) False
 
   , action "createFork(string)" $
       \sig input -> case decodeBuf [AbiStringType] input of
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -83,6 +83,8 @@ blankState = do
     , gas          = initialGas
     , returndata   = mempty
     , static       = False
+    , overrideCaller = Nothing
+    , resetCaller  = True
     }
 
 -- | An "external" view of a contract's bytecode, appropriate for
@@ -145,6 +147,8 @@ makeVm o = do
       , gas = o.gas
       , returndata = mempty
       , static = False
+      , overrideCaller = Nothing
+      , resetCaller = True
       }
     , env = env
     , cache = cache
@@ -153,8 +157,6 @@ makeVm o = do
     , iterations = mempty
     , config = RuntimeConfig
       { allowFFI = o.allowFFI
-      , resetCaller = True
-      , overrideCaller = Nothing
       , baseState = o.baseState
       }
     , forks = Seq.singleton (ForkState env block cache "")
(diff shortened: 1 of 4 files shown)


msooseth:
@elopez could you maybe review?  Maybe it'd be useful to add more tests in `tests.hs`, perhaps you could have a go at that?

src/EVM.hs
```diff
@@ -83,6 +83,8 @@ blankState = do
     , gas          = initialGas
     , returndata   = mempty
     , static       = False
+    , overrideCaller = Nothing
+    , resetCaller  = True
```
elopez, line 87:
```suggestion
    , resetCaller  = False
```
to match with the override = Nothing

src/EVM.hs
```diff
@@ -145,6 +147,8 @@ makeVm o = do
       , gas = o.gas
       , returndata = mempty
       , static = False
+      , overrideCaller = Nothing
+      , resetCaller = True
```
elopez, line 151:
```suggestion
      , resetCaller = False
```
to match with the override = Nothing

src/EVM.hs
```diff
@@ -866,17 +868,16 @@ exec1 = do
...
-                            let from' = fromMaybe self vm.config.overrideCaller
+                            let from' = fromMaybe self overrideC
                             zoom #state $ do
                               assign #callvalue xValue
                               assign #caller from'
                               assign #contract callee
-                            do
-                              resetCaller <- use (#config % #resetCaller)
-                              when (resetCaller) $ assign (#config % #overrideCaller) Nothing
+                              assign #overrideCaller Nothing
```
elopez, line 880:
```suggestion
                              assign #overrideCaller Nothing
                              assign #resetCaller False                              
```

src/EVM.hs
```diff
@@ -889,14 +890,13 @@ exec1 = do
...
+                    overrideC <- use $ #state % #overrideCaller
                     delegateCall this gas xTo self xValue xInOffset xInSize xOutOffset xOutSize xs $ \_ -> do
                       zoom #state $ do
                         assign #callvalue xValue
-                        assign #caller $ fromMaybe self vm.config.overrideCaller
-                      do
-                        resetCaller <- use (#config % #resetCaller)
-                        when (resetCaller) $ assign (#config % #overrideCaller) Nothing
+                        assign #caller $ fromMaybe self overrideC
+                        assign #overrideCaller Nothing
```
elopez, line 899:
```suggestion
                        assign #overrideCaller Nothing
                        assign #resetCaller False
```

src/EVM.hs
```diff
@@ -978,17 +978,16 @@ exec1 = do
...
                           zoom #state $ do
                             assign #callvalue (Lit 0)
-                            assign #caller $ fromMaybe self (vm.config.overrideCaller)
+                            assign #caller $ fromMaybe self overrideC
                             assign #contract callee
                             assign #static True
-                          do
-                            resetCaller <- use (#config % #resetCaller)
-                            when (resetCaller) $ assign (#config % #overrideCaller) Nothing
+                            assign #overrideCaller Nothing
```
elopez, line 990:
```suggestion
                            assign #overrideCaller Nothing
                            assign #resetCaller False
```

src/EVM.hs
```diff
@@ -1101,7 +1100,9 @@ callChecks this xGas xContext xTo xValue xInOffset xInSize xOutOffset xOutSize x
     accessMemoryRange xOutOffset xOutSize $ do
       availableGas <- use (#state % #gas)
       let recipientExists = accountExists xContext vm
-      let from = fromMaybe vm.state.contract vm.config.overrideCaller
+      let from = fromMaybe vm.state.contract vm.state.overrideCaller
+      resetCaller <- use $ #state % #resetCaller
+      when resetCaller $ assign (#state % #overrideCaller) Nothing
```
elopez, line 1105: This should probably also set resetCaller to False at the same time

elopez: having this on callChecks feels a bit odd to me, should it maybe live on delegateCall instead?

src/EVM.hs
```diff
@@ -1799,26 +1797,26 @@ cheatActions = Map.fromList
       \sig input -> case decodeStaticArgs 0 1 input of
         [addr]  -> case wordToAddr addr of
           Just a -> do
-            assign (#config % #overrideCaller) (Just a)
             doStop
```
elopez, line 1800: Can you add a comment explaining why doStop is done first? I imagine this lets you edit the parent's `#state` which is the one who should hold the prank, but it's non-obvious.

src/EVM.hs
```diff
@@ -1799,26 +1797,26 @@ cheatActions = Map.fromList
...
+            assign (#state % #resetCaller) False
           Nothing -> vmError (BadCheatCode "startPrank(address), could not decode address provided" sig)
         _ -> vmError (BadCheatCode "startPrank(address) parameter decoding failed" sig)
 
   , action "stopPrank()" $
       \_ _ -> do
-        assign (#config % #overrideCaller) Nothing
-        assign (#config % #resetCaller) True
         doStop
+        assign (#state % #overrideCaller) Nothing
```
elopez, line 1819:
```suggestion
        assign (#state % #overrideCaller) Nothing
        assign (#state % #resetCaller) False
```

src/EVM.hs
```diff
@@ -2014,6 +2011,7 @@ delegateCall this gasGiven xTo xContext xValue xInOffset xInSize xOutOffset xOut
                     assign #memorySize 0
                     assign #returndata mempty
                     assign #calldata calldata
+                    assign #overrideCaller Nothing
```
elopez, line 2014:
```suggestion
                    assign #overrideCaller Nothing
                    assign #resetCaller False
```

elopez (commented):
Overall it looks good to me! I left you a few suggestions to make resetCaller state more consistent

src/EVM.hs
msooseth: This has been fixed now :)

src/EVM.hs
msooseth: Hmmm... yeah, we could set `resetCaller` in `delegateCall`. However, the `let from = ..` needs to be here of course. Wanna have a go at putting it in `delegateCall`? Just create a PR on top of this PR :)

msooseth: Thanks, good point! I fixed it :) Let me know what you think!

PR #612, opened 2024-12-05
Cleanups in preparation of GHC 9.8

elopez:
## Description

This PR does a few cleanups to reduce the number of warnings/errors when compiling with GHC 9.8

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/scripts/install-libsecp256k1.sh
+++ b/.github/scripts/install-libsecp256k1.sh
@@ -3,7 +3,7 @@ set -eux -o pipefail
 
 ## The following script builds and installs libsecp256k1 to ~/.local/lib
 
-INSTALL_VERSION=0.6.0
+INSTALL_VERSION=0.5.1
 
 if [[ "$(uname -s)" =~ ^MSYS_NT.* ]]; then
     echo "This script is only meant to run on Windows under MSYS2"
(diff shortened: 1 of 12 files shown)


msooseth (approved):
Wow, nice, thanks! LGTM -- did you check the warning about 

```
Module ‘Data.Aeson’ does not export ‘json’
```

elopez:
@msooseth Yeah, older aeson versions export `json` and we have a variable called that all around the file, so it needs hiding. But the newer versions don't have it anymore, so it produces that warning. I can't remove the `hiding ...` or it would break the build for older versions of aeson. Maybe we can bump the min aeson version to something higher? I think I tried that and it failed for a different reason but I can't recall the specifics now; give it a try if you can 😄

elopez:
I was hoping to update nixpkgs further as well but currently z3 is broken: https://github.com/NixOS/nixpkgs/pull/327052

msooseth:
Ahhhh okay, that makes sense regarding the `hiding`. Sad that Z3 is broken. Thanks, merging!

numinit:
I'm hoping that should be merged soon. :-)

PR #619, opened 2024-12-17
Fixing missing `concKeccakSimpExpr` for `wordToAddr`, `maybeLitByte`, etc.

### LINKED ISSUE #618: `STATICCALL` Error "Unable to determine a call target"
I'm encountering an issue when running a symbolic execution test case with `hevm` for my contract. Specifically, the following error message appears when trying to explore the symbolic execution path for the `prove_Swap(uint256)` function:
```bash
[RUNNING] prove_Swap(uint256)
   WARNING: hevm was only able to partially explore the call prefix 0xde6fb24b due to the following issue(s):
       msg: "Unable to determine a call target"
       opcode: STATICCALL
       program counter: 13155
       arguments: 
         (And
           1461501637330902918203684832716283019655932542975
           (And
             1461501637330902918203684832716283019655932542975
             (SLoad
               slot:
                 (Keccak
                   (ConcreteBuf
                     Length: 64 (0x40) bytes
                     0000:   00 00 00 00  00 00 00 00  00 00 00 00  d9 53 22 74   .............S"t
                     0010:   58 65 82 27  19 16 4b 1f  c1 67 93 07  54 c2 48 de   Xe.'..K..g..T.H.
                     0020:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 00   ................
                     0030:   00 00 00 00  00 00 00 00  00 00 00 00  00 00 00 4a   ...............J
                   )
                 )
               storage:
                 (ConcreteStore
                   vals:
                     (0x40,0x12309ce54000)
                     (0x41,0x7)
                     (0x42,0x53b1daa2fd98364)
                     (0x43,0x493b9f4e9de52f7)
                     (0x44,0x5af3107a4000)
                     (0x45,0xde0b6b3a7640000)
                     (0x46,0x470de4df820000)
                     (0x47,0xacab)
                     (0x48,0x2beec76ec7ffde5ddec5c7f7bc28375a139d8dec)
                     (0x49,0x5)
                     (0xbcbfa19733cec6d55a5a287f85260feea2d177a81ed4201be8f853017b387b5,0x1)
                     (0x13897ad23775ee38cfc3ba4e9fdb836d9e8cc936c144e95e360abde83213ce33,0x1)
                     (0x18f40bf710eef8eb8b14a1cd644b7b8f6abe1c497e5380f59172f91f726b5ca9,0x1)
                     (0x33c01989199dc6bcc08e31fbeca7c654030b9d3fd3bda7c618f691ba2fc13e71,0x1)
                     (0x343bdc6eebb233b371e49d1a36a68177caf342e418df1d089bac9bf75e021781,0x1)
                     (0x37e472f504e93744df80d87316862f9a8fd41a7bc266c723bf77df7866d75f55,0xd95322745865822719164b1fc167930754c248de)
                     (0x37e472f504e93744df80d87316862f9a8fd41a7bc266c723bf77df7866d75f56,0x6eecb916b78765d6b277bff328684db91f4f6a5d)
                     (0x37e472f504e93744df80d87316862f9a8fd41a7bc266c723bf77df7866d75f57,0x8840f3a24be3971c313f435453c9c749d3f936c)
                     (0x37e472f504e93744df80d87316862f9a8fd41a7bc266c723bf77df7866d75f58,0xf64a197b987fd4400a0887909c067a37ae209e1a)
                     (0x37e472f504e93744df80d87316862f9a8fd41a7bc266c723bf77df7866d75f59,0xe663b9e12c29924565ee36fbb8c4a3191971f4da)
                     (0x42776d27e37c472cdec0856a9d58feaee2fde7e868fb53f36330a4a42e3a58a9,0x1)
                     (0x46eda384002bf299f519d3a9ef467d4225b0a66aa70e94ae3d558d3d71330e2c,0x5d51444bdd2bc621788319959c657af0f9d619c5)
                     (0x58b5e753403da07442b646a1311cfc1ac7ee398f664e8ba00f86755a18aafbf4,0x4)
                     (0x7f38886520bb8e417575a1c3c82c9404d0055e706357f640ecf1c94b462da90c,0xbb8daedddc5bb0483a4f44bfa628cc5daa72e2d1)
                     (0x9016d09d72d40fdae2fd8ceac6b6234c7706214fd39c1cd1e609a0528c199300,0xacab)
                     (0x9b779b17422d0df92223018b32b4d1fa46e071723d6817e2486d003becc55f00,0x2)
                     (0x9f1fd62955e083dd560a43035289fc6cb1df5e1e76c50e41f8a46ed187b86f8a,0xc2bbaea76ec6b85a0b90f04de8ce4476d6e27374)
                     (0xae6c11380398076a1cf99b4510e55da6b3062b4c9f858fa16277e672f2a5bd79,0x841c03e9bf1c008fe90b2dedc429d497795301a5)
                     (0xba962bc6bb35f1f8f35f5da32ced64e0c82d793a2d77283935f688b01f0cb163,0x0)
                     (0xcd5ed15c6e187e77e9aee88184c21f4f2182ab5827cb3b7e07fbedcd63f03300,0x0)
                     (0xebd33f63ba5dda53a45af725baed5628cdad261db5319da5f5d921521fe1161d,0x5842cfa0028e51f1b74c9dca4a59ef0e05544163)
                     (0xf0c57e16840df040f15088dc2f81fe391c3923bec73e23a9662efc9c229c6a00,0x1)
                     (0xf4344425ee2bee17a4c8a465e6fb8acc37882812cb3ff3774a8123a9a581eb50,0x3)
                     (0xf524c08d9fb11ff0e21253f37caa0ffe703c4ad2cf61a0f6771265f91f5907a8,0x2)
                 )
             )
           )
         )
```
This error indicates that the symbolic execution engine is unable to determine the call target during the execution of a `STATICCALL` operation. This prevents `hevm` from fully exploring the path.

Here is my test file snippet
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.20;

import { Test } from "forge-std/Test.sol";
import { console2 } from "forge-std/Test.sol";
import { OldPool } from "../src/pool/OldPool.sol";


contract OldPoolTest is Test {
    OldPool public pool;

    function setUp() public {
        // setup logic
    }

    function testFuzz_Swap(uint256 amount) public {
        vm.assume(amount > 10_000 * 1e6);
        vm.assume(amount < 150_000 * 1e6);

        _sequence(amount);
        assertGe(amount, USDC.balanceOf(address(this)));
    }

    function prove_Swap(uint256 amount) public {
        if (amount != 0) {
            _sequence(amount);
            assertGe(amount, USDC.balanceOf(address(this)));
        }
    }

    function _sequence(uint256 amount) private {
        // test logic
    }
}
```
Note that I use `_sequence` function for both forge fuzz test and hevm, and fuzz working just fine

msooseth:
## Description
It turns out that as per #618 we ran `simplify` instead of run `concKeccakSimpExpr` on `wordToAddr`. Furthermore, we should run `concKeccakSimpExpr` on `maybeLitByte`, `maybeLitAddr` etc. so more things can run concretely.

Also added `hexStringToByteString` to test.hs, which makes it a LOT easier to create `ConcreteBuf`-s via e.g. `ConcreteBuf (hexStringToByteString "0a0056")`. It's been really painful to try to debug without this.

Added a test case to verify that indeed the caase in #618 is fixed.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (commented):
Are you sure the `maybe` methods are the place where simplification should happen?
Shouldn't these methods stay trivial, and shouldn't the simplification happen in somewhere else?

msooseth:
@blishko Hmmmm I found it cleaner this way. I would need to add it to, at least:
* forceConcrete
* forceConcrete2
* OpSstore
* OpReturn
* precompiledContract
* decodeBuf
...

And that's just `EVM.hs` and `maybeLitWord`. I do see your point, though, i.e. to be more explicit in its use. However, these `maybe...` functions either force the execution to error out or to do something less convenient when they return a `Nothing`, so I think we'd need to put the `concKeccakSimp` almost all the places they are used.

But maybe I'm wrong. I just see it to be a larger diff, and hence more clutter. But it's not such a big deal, I can be convinced the other way. If you think it's better, I'll adjust this PR to use `concKeccakSimp` almost everywhere -- I think there'd be 2-3 places where it wouldn't be needed -- in `Format.hs` for example, though still maybe nice to output the concrete value, too, if possible.

msooseth:
Yeah, renaming of the functions is a very good idea, actually, thanks! Misleading without it. 

`forceConcrete` is definitely a good place where we really ought to try to simplify as much as possible. In fact, it should even try to run an SMT solver. For example at `wordToAddr` @d-xo wrote:

```haskell
-- TODO: make this smarter, probably we will need to use the solver here?
wordToAddr :: Expr EWord -> Maybe (Expr EAddr)
wordToAddr e = case (concKeccakSimpExpr e) of
[...]
```

And I actually agree with this idea In fact, we'd need that here: #581 In order to remove that symbolic jump issue, we'd need to run an SMT solver, and get out all possible jump destinations, and explore the execution with all. That's a bit of a more involved case. For that to work, we will need to rewrite some of the places where these functions are used, because if the number of solution is limited (like in #581 where it's limited to the number of functions in the contracts), then we should try all -- and of course in these cases neither `simplify`, nor `concKeccakSimpExpr` will be able to get out a single concrete value, since there are multiple possible concrete values.

Let me rename the functions, let's see if that helps :)

src/EVM/Expr.hs
```diff
@@ -1616,3 +1614,25 @@ checkLHSConstProp a = isJust $ mapPropM_ lhsConstHelper a
 -- Commutative operators should have the constant on the LHS
 checkLHSConst :: Expr a -> Bool
 checkLHSConst a = isJust $ mapExprM_ lhsConstHelper a
+
+maybeLitByte :: Expr Byte -> Maybe Word8
```
blishko, line 1618:
```suggestion
maybeLitByteSimp :: Expr Byte -> Maybe Word8
```

blishko (commented):
I guess also `maybeLitByte` should have the `Simp` prefix?

src/EVM/Expr.hs
msooseth: Ayyyy you are so good. I'm sorry for wasting your time. Let me fix.

msooseth:
It's about holidays time, but when you are back, I'd be grateful for a review, @blishko  :) Happy holidays in the meanwhile!

msooseth:
> I guess also `maybeLitByte` should have the `Simp` suffix?

Yesss, thanks, I added it! Can you perhaps review in the coming days? Also, if you could review #625 that's on top of this that'd be nice :)

blishko (approved):
Seems OK now. Although, you might need to rebase to resolve the conflicts.

PR #622, opened 2024-12-23
Fixing missing override on EVM DeletegateCall instruction? I think it's missing...

msooseth:
## Description

I think the DelegateCall was missing the override check? @elopez what do you think?

Notice that the override will be reset by the `delegateCall` function:

```
        \xGas -> do
          resetCaller <- use $ #state % #resetCaller
          when resetCaller $ assign (#state % #overrideCaller) Nothing
```

But it does not assign the `from` caller to the value, instead, we do that in e.g.:

```haskell
        OpStaticcall ->
          case stk of
            xGas:xTo:xInOffset:xInSize:xOutOffset:xOutSize:xs ->
[...]
                Just xTo' ->
                  case gasTryFrom xGas of
                    Left _ -> vmError IllegalOverflow
                    Right gas -> do
                      overrideC <- use $ #state % #overrideCaller
                      delegateCall this gas xTo' xTo' (Lit 0) xInOffset xInSize xOutOffset xOutSize xs $
                        \callee -> do
                          zoom #state $ do
                            assign #callvalue (Lit 0)
                            assign #caller $ fromMaybe self overrideC

[...]
```

And similarly in other places. But somehow OpDelegateCall is missing this.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -952,6 +952,8 @@ exec1 = do
                   case gasTryFrom xGas of
                     Left _ -> vmError IllegalOverflow
                     Right gas ->
+                      -- NOTE: we don't update overrideCaller in this case because
+                      -- forge-std doesn't. see: https://github.com/foundry-rs/foundry/pull/8863
                       delegateCall this gas xTo' self (Lit 0) xInOffset xInSize xOutOffset xOutSize xs $
                         \_ -> touchAccount self
             _ -> underrun


elopez:
@msooseth `DELEGATECALL` is special as it normally preserves the context (ie the caller remains the same as it was before the call), I'm not sure if pranking should be touching the caller here. Have you looked at Foundry's behavior to see what it does in this particular case?

msooseth:
Ah, interesting. So they have a version of `prank` with a 2nd argument that, if set to `true`, will also affect `delegateCall`. But it does not affect `delegateCall` otherwise.

Discussion: https://github.com/foundry-rs/foundry/pull/8863
The PR that got merged: https://github.com/foundry-rs/foundry/pull/8863/files

So it's good as-is. I am changing this PR to add a note instead so it's documented what's going on.

PR #625, opened 2025-01-02
Allow dealing with abi.encodeWithSelector

msooseth:
## Description
This helps to deal with `abi.encodeWithSelector`. Adds a corresponding simplification rules.

Fixes the test case added:

```solidity
contract C {
function prove_warp_symbolic(uint128 jump) public {
        uint pre = block.timestamp;
        address hevm = 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D;
        (bool success, ) = hevm.call(abi.encodeWithSelector(bytes4(keccak256("warp(uint256)")), block.timestamp+jump));
        require(success, "Call to hevm.warp failed");
        assert(block.timestamp == pre + jump);
    }
}
```

Meant to be used together with #619 and works only on top of #619 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -992,6 +992,20 @@ simplify e = if (mapExpr go e == e)
             where simplifiedBuf = BS.take (unsafeInto (n+sz)) buf
     go (CopySlice a b c d f) = copySlice a b c d f
 
+    go (JoinBytes (LitByte a0)  (LitByte a1)  (LitByte a2)  (LitByte a3)
+      (LitByte a4)  (LitByte a5)  (LitByte a6)  (LitByte a7)
+      (LitByte a8)  (LitByte a9)  (LitByte a10) (LitByte a11)
+      (LitByte a12) (LitByte a13) (LitByte a14) (LitByte a15)
+      (LitByte a16) (LitByte a17) (LitByte a18) (LitByte a19)
+      (LitByte a20) (LitByte a21) (LitByte a22) (LitByte a23)
+      (LitByte a24) (LitByte a25) (LitByte a26) (LitByte a27)
+      (LitByte a28) (LitByte a29) (LitByte a30) (LitByte a31)) =
+        let b =  map fromIntegral [a0, a1, a2, a3 ,a4, a5, a6, a7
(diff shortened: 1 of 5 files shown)


src/EVM/Format.hs
```diff
@@ -471,7 +471,7 @@ formatPartial = \case
   JumpIntoSymbolicCode pc idx -> "Encountered a jump into a potentially symbolic code region while executing initcode. pc: " <> pack (show pc) <> " jump dst: " <> pack (show idx)
 
 formatSomeExpr :: SomeExpr -> Text
-formatSomeExpr (SomeExpr e) = formatExpr e
```
msooseth, line 474: I added this change, because it's waaay nicer for the user, and we should be getting cleaner issues, too.

msooseth:
NOTE: This is failing currently only because #619 is not yet merged :)

src/EVM/Expr.hs
```diff
@@ -1418,6 +1432,11 @@ indexWord i@(Lit idx) e@(And (Lit mask) w)
     fullWordMask = (2 ^ (256 :: W256)) - 1
     unmaskedBytes = fromIntegral $ (countLeadingZeros mask) `Prelude.div` 8
     isByteAligned m = (countLeadingZeros m) `Prelude.mod` 8 == 0
+-- This pattern happens in Solidity for function selectors. Since Lit 0xfff... (28 bytes of 0xff)
+-- is masking the function selector, it can be simplified to just the function selector bytes. Remember,
+-- indexWord takes the MSB i-th byte when called with (indexWord i).
+indexWord (Lit a) (Or funSel (And (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffff) _)) | a < 4 =
+  indexWord (Lit a) funSel
```
blishko, line 1439: Do I understand it correctly that this simplification is correct because the four most significant bytes of `(And (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffff) _)` will always be zero?

blishko (approved):
LGTM!

PR #626, opened 2025-01-02
Faster test running by default

msooseth:
## Description
Let's run 3 solver processes by default. Should make testing a lot faster.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
Why do you think 3 is the right magic number? :)

msooseth:
Hahaha not sure! I was afraid with more processes thing will run out of memory... Maybe it should be more....

msooseth:
Merging for now, may update it later :)

PR #628, opened 2025-01-06
More Prop simplification rules

msooseth:
## Description
Three more prop simplification rules.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1220,7 +1220,9 @@ simplifyProp prop =
     go (PLEq _ (Lit x)) | x == maxLit = PBool True
     go (PLT  (Lit val) (Var _)) | val == maxLit = PBool False
     go (PLEq (Var _) (Lit val)) | val == maxLit = PBool True
+    go (PLT a b) | a == b = PBool False
     go (PLT (Lit l) (Lit r)) = PBool (l < r)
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -1232,6 +1234,9 @@ simplifyProp prop =
     go (PNeg (PBool b)) = PBool (Prelude.not b)
     go (PNeg (PNeg a)) = a
 
+    -- Empty buf
+    go (PEq (Lit 0) (BufLength k)) = peq k (ConcreteBuf "")
```
blishko, line 1238: Are you sure this is a simplification?
Is it better on the SMT level to compare arrays instead of computing buffer length and comparing it to `0`?

msooseth:
Not entirely sure… the Props look nicer to my human eyes though 😂 I can
remove it if you think it’s all the same/worse. What do you think?

Mate

blishko:
I think it's OK. I was just thinking that if ever `hevm` would couple buffers with length, then asking for length would be much easier.
But I guess that is not going to happen any time soon.

blishko (approved):
LGTM!

msooseth:
Thanks for the review!

PR #632, opened 2025-01-15
Ooops, staticcall was wrong. This fixes it.

msooseth:
## Description
Wow, just discovered this

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


test/test.hs
```diff
@@ -3944,7 +3948,7 @@ tests = testGroup "hevm"
                 False -> recursiveList ax (a:b)
           recursiveList [] b = pure b
         files <- liftIO $ recursiveList fullpaths []
-        let filesFiltered = filter (\file -> not $ any (`List.isSubsequenceOf` file) ignoredTests) files
+        let filesFiltered = filter (\file -> not $ any (`List.isInfixOf` file) ignoredTests) files
```
blishko, line 3951: `isInfixOf` is more strict than `isSubsequenceOf`, right?
But this is comparing paths, no?
Is it going to make any difference here?

blishko:
> Turns out our staticcall abstraction messed up the stack. It actually left too many elements on the stack. This fixes that.

This is related to the recent changes you made? The abstraction of the static calls? Was that before or after the last release?

test/test.hs
msooseth: It's a huge difference :) The `isSubsequenceOf` actually does something _very_ different that what we wanted!

`isSubsequenceOf` will say YES, when "abc" is in the list of exclusions, and the file checked is "sadfasdf-aaaab-bbba-aacc-ccccaa-abbb-bsdf-asdf"

`isInfixOf` will say NO. It will only say YES to: `something-abc-somethingelse` or `abc` or `something-abc` or `abc-something-abc`

This was a bug -- it was meant to be `isInfixOf`. Not `isSubsequenceOf`!

msooseth:
It's related to staticcall abstraction. The stack was messed up, it didn't actually remove the elements from the stack that it was supposed to remove. I had to fix the test filtering too, because our test filtering was wrong, and it filtered out the test that actually catches this bug! :)

blishko:
I was asking mostly to know if the problem with staticcall abstraction has been there for a while or if it was something you introduced in the recent changes.
Because in the latter case I would say there should not be a new changelog entry about it, since it would not be something that is fixed compared to the last release.
Anyway, that's just nitpicking...

msooseth:
> I was asking mostly to know if the problem with staticcall abstraction has been there for a while or if it was something you introduced in the recent changes. Because in the latter case I would say there should not be a new changelog entry about it, since it would not be something that is fixed compared to the last release. Anyway, that's just nitpicking...

Fair :) I fixed the changelog now:

```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 53661689..7b5c734d 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - More simplification rules for Props
 - Using the SMT solver to get a single concrete value for a symbolic expression
   and continue running, whenever possible
+- STATICCALL abstraction is now performed in case of symbolic arguments

 ## Fixed
 - We now try to simplify expressions fully before trying to cast them to a concrete value
   This should improve issues when "Unexpected Symbolic Arguments to Opcode" was
   unnecessarily output
-- STATICCALL abstraction left incorrect stack, fixed
 - Not all testcases ran due to incorrect filtering, fixed
 - Removed dead code related to IOAct in the now deprecated and removed debugger
 ```

Do you think we can merge this PR?

blishko (approved):
LGTM!

msooseth:
Thanks so much!

PR #636, opened 2025-01-22
Forgot to set dumpEndStates for --debug

msooseth:
## Description
This should be default for `--debug` as it's not noisy at all and is very helpful

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -211,6 +211,7 @@ main = withUtf8 $ do
   let env = Env { config = defaultConfig
     { dumpQueries = cmd.smtdebug
     , debug = cmd.debug
+    , dumpEndStates = cmd.debug
     , dumpExprs = cmd.debug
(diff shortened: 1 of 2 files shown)


cli/cli.hs
```diff
@@ -47,6 +47,7 @@ import EVM.Types qualified
 import EVM.UnitTest
 import EVM.Effects
 import EVM.Expr (maybeLitWordSimp, maybeLitAddrSimp)
+import EVM.Effects (Config(dumpEndStates))
```
blishko, line 50: I think the bot is right, no? There is already `import EVM.Effects` two lines above.

msooseth:
Yes, sorry, I’ll fix. It was a mistake :( Debugging another issue at the
moment :)

blishko (approved):
LGTM!

PR #641, opened 2025-01-28
Update more precise smt address encoding PR

msooseth:
## Description
This is an update to #376  that makes it work for the test cases added to that PR. As per @d-xo 's description:

`This adds some more assertions during encoding around the values of symbolic addresses. Still needed is a pass that asserts that all symbolic addresses are pairwise distinct with all known concrete addresses.`

I have added the part that they are pairwise distinct. It is also working now because it checks the addresses _before_ they get erased through the revert that `VMFailure` does. So the generation of the constraints is done _before_ the `VMFailure`'s revert. This fixes the underlying issue that I think may have been created since that PR.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -436,7 +436,7 @@ declareVars names = SMT2 (["; variables"] <> fmap declare names) cexvars mempty
 
 -- Given a list of variable names, create an SMT2 object with the variables declared
 declareConstrainAddrs :: [Builder] -> SMT2
-declareConstrainAddrs names = SMT2 (["; concretea and symbolic addresseses"] <> fmap declare names <> fmap assume names) cexvars mempty
+declareConstrainAddrs names = SMT2 (["; concrete and symbolic addresses"] <> fmap declare names <> fmap assume names) cexvars mempty
   where
     declare n = "(declare-fun " <> n <> " () Addr)"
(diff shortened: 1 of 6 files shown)


src/EVM/SMT.hs
```diff
@@ -435,10 +435,12 @@ declareVars names = SMT2 (["; variables"] <> fmap declare names) cexvars mempty
     cexvars = (mempty :: CexVars){ calldata = fmap toLazyText names }
 
 -- Given a list of variable names, create an SMT2 object with the variables declared
-declareAddrs :: [Builder] -> SMT2
-declareAddrs names = SMT2 (["; symbolic addresseses"] <> fmap declare names) cexvars mempty
+declareConstrainAddrs :: [Builder] -> SMT2
+declareConstrainAddrs names = SMT2 (["; concretea and symbolic addresseses"] <> fmap declare names <> fmap assume names) cexvars mempty
```
blishko, line 439:
```suggestion
declareConstrainAddrs names = SMT2 (["; concrete and symbolic addresses"] <> fmap declare names <> fmap assume names) cexvars mempty
```

blishko (commented):
Apart from the typo in the comment. I think this is good! :)

msooseth:
Nice, thanks for the review!

PR #642, opened 2025-01-29
Constant propagation over Props

msooseth:
## Description

Performs constant propagation on the Prop before sending it to the solver

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1222,7 +1222,7 @@ simplifyProps ps = if cannotBeSat then [PBool False] else simplified
     simplified = if (goOne ps == ps) then ps else simplifyProps (goOne ps)
     cannotBeSat = PBool False `elem` simplified
     goOne :: [Prop] -> [Prop]
-    goOne newPs = remRedundantProps . map simplifyProp . constPropagate . flattenProps $ newPs
+    goOne = remRedundantProps . map simplifyProp . constPropagate . flattenProps
 
 
 -- | Evaluate the provided proposition down to its most concrete result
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -1219,8 +1219,11 @@ simplify e = if (mapExpr go e == e)
 simplifyProps :: [Prop] -> [Prop]
 simplifyProps ps = if cannotBeSat then [PBool False] else simplified
   where
-    simplified = remRedundantProps . map simplifyProp . flattenProps $ ps
-    cannotBeSat = isUnsat simplified
+    simplified = if (goOne ps == ps) then ps else simplifyProps (goOne ps)
+    cannotBeSat = PBool False `elem` simplified
+    goOne :: [Prop] -> [Prop]
+    goOne newPs = remRedundantProps . map simplifyProp . constPropagate . flattenProps $ newPs
```
blishko, line 1225: Would this work? Is this more or less readable?

```suggestion
    goOne = remRedundantProps . map simplifyProp . constPropagate . flattenProps
```

src/EVM/Expr.hs
```diff
@@ -1567,38 +1570,54 @@ data ConstState = ConstState
   }
   deriving (Show)
 
--- | Checks if a conjunction of propositions is definitely unsatisfiable
-isUnsat :: [Prop] -> Bool
-isUnsat ps = Prelude.not $ oneRun ps (ConstState mempty True)
+-- | Performs constant propagation
```
blishko, line 1573: It's cool that you managed to generalize this function!

blishko (approved):
I think this is good!

src/EVM/Expr.hs
msooseth: Haha that's nice! I love this point-free version. I should do/learn point-free programming....

src/EVM/Expr.hs
msooseth: Thanks! I was pretty happy about myself actually. Kinda weird it worked so easily.

PR #643, opened 2025-01-30
Early multi-solutions system

msooseth:
## Description
Making progress towards symbolic jumpdestinations via SMT solving. This PR allows us to branch out to all potential jumpdests.

The system can restrict the returned value's number of relevant bytes to a fixed value. So if we are looking for a JUMP, it ignores high bytes, because it's impossible to have a PC that's more than 2^16 (more than 65K instructions). Similarly for addresses. This matters, because it can happen that there's some junk that's not restricted and we could get thousands of solutions, but they are actually all the same address if we ignore the bits other than the 160b that an address has.

Currently, it launches a new query for each new solution, which is very slow. This needs to be improved in a new PR. That will require a slight re-engineering of Solvers.hs, so we can query the system for more than one solution at a time.

Number of potential concrete is set to 10 by default. This is actually kinda low. However, we need to be a bit cautious, as things can blow up quickly, so I set 10. It can be changed via `--max-branch K`.

Currently, it does NOT use incremental solving to get multiple solutions. This is not optimal. Hence, it's called "early", as a next PR will fix that.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -810,7 +810,7 @@ exec1 = do
         OpJump ->
           case stk of
             x:xs ->
-              burn g_mid $ forceConcrete x "JUMP: symbolic jumpdest" $ \x' ->
+              burn g_mid $ forceConcreteLimitSz x 2 "JUMP: symbolic jumpdest" $ \x' ->
                 case tryInto x' of
                   Left _ -> vmError BadJumpDestination
                   Right i -> checkJump i xs
@@ -818,7 +818,7 @@ exec1 = do
 
         OpJumpi ->
           case stk of
-            x:y:xs -> forceConcrete x "JUMPI: symbolic jumpdest" $ \x' ->
+            x:y:xs -> forceConcreteLimitSz x 2 "JUMPI: symbolic jumpdest" $ \x' ->
               burn g_high $
                 let jump :: Bool -> EVM t s ()
                     jump False = assign (#state % #stack) xs >> next
@@ -887,7 +887,7 @@ exec1 = do
                       Left _ -> vmError IllegalOverflow
                       Right gas -> do
                         overrideC <- use $ #state % #overrideCaller
-                        delegateCall this gas xTo xTo xValue xInOffset xInSize xOutOffset xOutSize xs $
+                        delegateCall this gas xTo xTo xValue xInOffset xInSize xOutOffset xOutSize xs unknownCodeFallback $
                           \callee -> do
                             let from' = fromMaybe self overrideC
                             zoom #state $ do
@@ -908,7 +908,7 @@ exec1 = do
                   Left _ -> vmError IllegalOverflow
                   Right gas -> do
                     overrideC <- use $ #state % #overrideCaller
-                    delegateCall this gas xTo self xValue xInOffset xInSize xOutOffset xOutSize xs $ \_ -> do
+                    delegateCall this gas xTo self xValue xInOffset xInSize xOutOffset xOutSize xs unknownCodeFallback $ \_ -> do
                       zoom #state $ do
                         assign #callvalue xValue
                         assign #caller $ fromMaybe self overrideC
@@ -963,7 +963,7 @@ exec1 = do
                     Right gas ->
                       -- NOTE: we don't update overrideCaller in this case because
                       -- forge-std doesn't. see: https://github.com/foundry-rs/foundry/pull/8863
-                      delegateCall this gas xTo' self (Lit 0) xInOffset xInSize xOutOffset xOutSize xs $
+                      delegateCall this gas xTo' self (Lit 0) xInOffset xInSize xOutOffset xOutSize xs unknownCodeFallback $
                         \_ -> touchAccount self
             _ -> underrun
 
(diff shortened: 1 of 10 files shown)


src/EVM.hs
```diff
@@ -3004,16 +3012,34 @@ instance VMOps Symbolic where
...
         assign #result Nothing
         continue Nothing
+    where
+      runMore vals leftOrRight = do
+        case length vals of
+          -- if 2, we run both, otherwise, we run 1st and run ourselves with the rest
+          2 -> if leftOrRight then runOne $ head vals
+               else runOne (head $ tail vals)
+          _ -> if leftOrRight then runOne $ head vals
+               else choose . PleaseChoosePath ewordExpr $ runMore (tail vals)
```
blishko, line 3037: Can you explain this a bit? I just probably don't understand `choose` and `PleaseChoosePath`, but what does the parameter `leftOrRight` represents and where is it actually passed in?

For the case of `2`, if `leftOrRight` is `false`, isn't this just ignoring the first value? I don't see where the "we run both" from the comment actually happens.

msooseth: Yeah, sorry. It's actually not CHOOSING anything. It's running both. It USED to be choosing things, because there was a debugger! But no longer. Let me rename that thing and then it'll be lot easier to understand :)

src/EVM.hs
```diff
@@ -1012,8 +1012,8 @@ exec1 = do
                             touchAccount self
                             touchAccount callee
               where
-                fallback :: EVM t s ()
-                fallback = do
+                fallback :: a -> EVM t s ()
+                fallback _ = do
```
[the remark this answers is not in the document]
msooseth: Let me make it a BIT more sane, but it's basically that we use this function here, above:

`delegateCall this gas xTo' xTo' (Lit 0) xInOffset xInSize xOutOffset xOutSize xs fallback $`

where `delegateCall`-s last argument is a function of the type: `(Expr EAddr -> EVM t s ())`. But we don't need the first argument. I wanted to be able to pass this function to `delegateCall` and also use it here:

`                      False -> fallback undefined`
and here:
`                Nothing -> fallback undefined`

But yeah, it's annoying. I'll make a helper function `strip1st :: (x) ->(a->x)` that makes a Monadic function that takes no arguments into a Monadic function that takes 1 argument, and ignores it. Then it will be maybe cleaner!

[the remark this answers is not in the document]
msooseth: Ha nice, OK, it's fixed now, using `const`. Thanks, this was useful. I tried figuring out this function, I also didn't like it, but I couldn't find it. I now used LLMs because, why not. It told me about `const`. Can you check if it's better now?

src/EVM.hs
msooseth: Yay, OK, it should be a lot cleaner now, renamed all relevant functions and datatypes. Can you re-review?

msooseth:
BTW, I also added a test:

```diff
+     , test "symbolic-to-concrete-multi" $ do
+        Just c <- solcRuntime "MyContract"
+            [i|
+            interface Vm {
+              function deal(address,uint256) external;
+            }
+            contract MyContract {
+              function fun(uint160 a) external {
+                Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+                uint160 c = 10 + (a % 2);
+                address b = address(c);
+                vm.deal(b, 10);
+              }
+             }
+            |]
+        let sig =Just (Sig "fun(uint160)" [AbiUIntType 160])
+        (e, [Qed _]) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        assertBoolM "The expression is not partial" $ Prelude.not (Expr.containsNode isPartial e)
```

Which failed before.

test/test.hs
```diff
@@ -2098,6 +2098,24 @@ tests = testGroup "hevm"
...
+            contract MyContract {
+              function fun(uint160 a) external {
+                Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+                uint160 c = 10 + (a % 2);
+                address b = address(c);
+                vm.deal(b, 10);
+              }
+             }
+            |]
+        let sig =Just (Sig "fun(uint160)" [AbiUIntType 160])
```
blishko, line 2116:
```suggestion
        let sig = Just (Sig "fun(uint160)" [AbiUIntType 160])
```

msooseth:
(sorry, had to update above description to be more precise)

blishko:
Hmm, I can give it another look in the evening. But it's fine if you want to merge this and move on right away.

blishko (approved):

msooseth:
There is no rush actually :) Take your time!

msooseth:
Ahhh, I see you approved... OK, let me merge. You can ask me about issues if have some. I feel like it's a non-trivial change, mostly because `Ask` hasn't been working for a long while due to the debugger being stripped away. The code now reflects this conceptual change better.

PR #652, opened 2025-02-10
Faster multi-solution system

msooseth:
## Description
Using SMT solver's incremental solving capability to vastly speed up obtaining multiple solutions

NOTE: if the number of solutions is zero, i.e. there are no valid solutions, we REVERT. This is not _perfect_ -- what that is telling us is that in fact the whole branch we are in is nonsense. We should stop the EVM execution completely. Not sure how to do that, though, as things expect a return value, etc. So the best I could come up with is a revert.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
LGTM!

Very nice!

PR #655, opened 2025-02-12
Two simplifications that will be needed to deal with @zoep's issues regarding initcode equivalence

msooseth:
## Description
One more simplification of `readByte`. This one works ALSO when `CopySlice` is not called with a `Lit` as a `srcOffset`. So it's more general, but cannot simplify some of the things that the more specific one can. It definitely helps in some cases, as per Zoe's issue

Note that the SMT overflow vs how the EVM actually works is slightly different on CopySlice. We need this rewrite rule, but it means that in some cases, the SMT will differ from the simplification system, when writing to extremely large offsets -- which always leads to over-gas-limit, so it's not a relevant case. However, the rewrite rule actually triggers, in real-world scenarios, and is useful.

NOTE: this is a *cherry-pick from Zoe's changeset* so it will apply perfectly when this is merged, appearing as less of a changeset for Zoe's PR.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Multiple solutions are allowed for a single symbolic expression
 - Aliasing works much better for symbolic and concrete addresses
 - Constant propagation for symbolic values
+- One more simplification rule for `ReadByte` when the `CopySlice` after it
+  is writing after the position being read, so the `CopySlice` can be ignored
 
 ## Fixed
 - We now try to simplify expressions fully before trying to cast them to a concrete value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Multiple solutions are allowed for a single symbolic expression
 - Aliasing works much better for symbolic and concrete addresses
 - Constant propagation for symbolic values
-- One more simplification rule for `ReadByte` when the `CopySlice` after it
-  is writing after the position being read, so the `CopySlice` can be ignored
+- Two more simplification rules: `ReadByte` & `ReadWord` when the `CopySlice`
+  it is reading from is writing after the position being read, so the
+  `CopySlice` can be ignored
 
 ## Fixed
 - We now try to simplify expressions fully before trying to cast them to a concrete value
--- a/test/test.hs
+++ b/test/test.hs
@@ -524,7 +524,7 @@ tests = testGroup "hevm"
       assertEqualM "readByte simplification" simp (ReadByte (Lit 0x0) (AbstractBuf "dst"))
     , test "simp-readByte2" $ do
       let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
-          size =(Lit 0x1)
+          size = (Lit 0x1)
           src = (AbstractBuf "stuff2")
           e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
           simp =Expr.simplify e
--- a/test/test.hs
+++ b/test/test.hs
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
         let e = BufLength (CopySlice (Lit 0x2) (Lit 0x2) (Lit 0x1) (ConcreteBuf "") (ConcreteBuf ""))
         b <- checkEquiv e (Expr.simplify e)
         assertBoolM "Simplifier failed" b
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      assertEqualM "readByte simplification" simp (ReadByte (Lit 0x0) (AbstractBuf "dst"))
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(Lit 0x1)
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      res <- checkEquiv e simp
+      assertEqualM "max-buflength rules"  res True
     , test "simp-max-buflength" $ do
       let simp = Expr.simplify $ Max (Lit 0) (BufLength (AbstractBuf "txdata"))
       assertEqualM "max-buflength rules" simp $ BufLength (AbstractBuf "txdata")
--- a/test/test.hs
+++ b/test/test.hs
@@ -517,7 +517,7 @@ tests = testGroup "hevm"
         assertBoolM "Simplifier failed" b
     , test "simp-readByte" $ do
       let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
-          size =(ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
+          size = (ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
           src = (AbstractBuf "stuff2")
           e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
           simp =Expr.simplify e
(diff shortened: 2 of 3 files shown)


test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
...
         b <- checkEquiv e (Expr.simplify e)
         assertBoolM "Simplifier failed" b
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      assertEqualM "readByte simplification" simp (ReadByte (Lit 0x0) (AbstractBuf "dst"))
+    , test "simp-readByte" $ do
```
blishko, line 525: Should this test use a different name than the previous one?

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
...
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(Lit 0x1)
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      res <- checkEquiv e simp
+      assertEqualM "max-buflength rules"  res True
```
blishko, line 532: The message is should something else, no?

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
...
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      assertEqualM "readByte simplification" simp (ReadByte (Lit 0x0) (AbstractBuf "dst"))
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(Lit 0x1)
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
+      res <- checkEquiv e simp
```
blishko, line 531: Wouldn't it be stronger to assert what the result of the simplification is?
I don't quite understand the purpose of this second test.

test/test.hs
```diff
@@ -562,7 +577,9 @@ tests = testGroup "hevm"
         a = BufLength (ConcreteBuf "ab")
         simp = Expr.simplify a
       assertEqualM "Must be simplified down to a Lit" simp (Lit 2)
-    , test "CopySlice-overflow" $ do
+    -- This test no longer works or needed, due to ReadWord simplification. Notice that it's writing
```
blishko, line 580: Wait, so does this fail now or is it simply no longer needed?
I think the original purpose of the test was to check this weird overflow behaviour when the copy slice actually wraps back to start reading from the beginning again.
But that is such a can of worms and the SMT semantics seems to be different than EVM semantics.

Maybe we should just remove the test completely?

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
         let e = BufLength (CopySlice (Lit 0x2) (Lit 0x2) (Lit 0x1) (ConcreteBuf "") (ConcreteBuf ""))
         b <- checkEquiv e (Expr.simplify e)
         assertBoolM "Simplifier failed" b
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
```
blishko, line 523:
```suggestion
          simp = Expr.simplify e
```

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
         let e = BufLength (CopySlice (Lit 0x2) (Lit 0x2) (Lit 0x1) (ConcreteBuf "") (ConcreteBuf ""))
         b <- checkEquiv e (Expr.simplify e)
         assertBoolM "Simplifier failed" b
+    , test "simp-readByte" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
```
blishko, line 520:
```suggestion
          size = (ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
```

blishko (commented):
The simplification is sound, but only if writing to a buffer does not overflow.
In that case we have a conflict between the SMT semantics and EVM semantics.
I guess we have agreed to ignore this weird edge case, right?

blishko:
Also, there seem to be some test failures on windows:
```
FAIL
      Exception: Maybe.fromJust: Nothing
        error, called at libraries\base\Data\Maybe.hs:150:21 in base:Data.Maybe
        fromJust, called at test\EVM\Test\Tracing.hs:880:30 in hevm-0.54.2-inplace-test:EVM.Test.Tracing
      Use -p '/calldata-wraparound-1/' to rerun this test only.
    calldata-wraparound-2:                                        Error converting trace! exit code:ExitFailure 127
```

test/test.hs
msooseth: I guess I wanted to prove that the simplification and the SMT solver agree... maybe I was a bit too eager to add 2 tests. It does prove that the SMT solver agrees, which is nice. Let me know what you think. Maybe it can be removed, honestly.

test/test.hs
msooseth: Good point! Yeah, removing!

msooseth:
The windows build issue is unrelated, unfortunately. It's occurring here too:

https://github.com/ethereum/hevm/actions/runs/13283178489/job/37085893025

which is also unrelated. I'll merge this with that issue, and I have already asked the Echidna folks to look into the Windows issue. I think it's something small, or at least I'm hoping :) Probably some shell script thing.

msooseth:
> The simplification is sound, but only if writing to a buffer does not overflow. In that case we have a conflict between the SMT semantics and EVM semantics. I guess we have agreed to ignore this weird edge case, right?

Yep, we did :)

test/test.hs
blishko: Right, I see. I guess it is good to have a test for that as well!

blishko: So, keep the test!
But please fix the spaces around `=` also in this test, as you've already done for the first one :)

msooseth: Oh man, I'm so sorry. I reviewed this PR and then I was like... let me add some tests. And then of course all the mess is from the tests... Sorry.

msooseth:
@blishko beware -- I added the same simplification to `readWord` as well, which I now noticed is missing.

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
...
+    , test "simp-readByte1" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size = (ReadWord (Lit 0x1) (AbstractBuf "stuff2"))
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp = Expr.simplify e
+      assertEqualM "readByte simplification" simp (ReadByte (Lit 0x0) (AbstractBuf "dst"))
+    , test "simp-readByte2" $ do
+      let srcOffset = (ReadWord (Lit 0x1) (AbstractBuf "stuff1"))
+          size =(Lit 0x1)
```
blishko, line 527:
```suggestion
          size = (Lit 0x1)
```

test/test.hs
```diff
@@ -515,6 +515,21 @@ tests = testGroup "hevm"
...
+          src = (AbstractBuf "stuff2")
+          e = ReadByte (Lit 0x0) (CopySlice srcOffset (Lit 0x10) size src (AbstractBuf "dst"))
+          simp =Expr.simplify e
```
blishko, line 530:
```suggestion
          simp = Expr.simplify e
```

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Multiple solutions are allowed for a single symbolic expression
 - Aliasing works much better for symbolic and concrete addresses
 - Constant propagation for symbolic values
+- One more simplification rule for `ReadByte` when the `CopySlice` after it
```
blishko, line 22: Please mention also the added simplification for `ReadWord` in the entry.

blishko (commented):
Maybe you should also add a test for the `ReadWord` simplification?

CHANGELOG.md
msooseth: Oh, good catch. Fixed!

msooseth:
> Maybe you should also add a test for the `ReadWord` simplification?

Yes.. sorry. I just added two :)

blishko (approved):
LGTM!

Issue #664, opened 2025-02-17
Simplification is not equivalent, but it should be


msooseth:
```
      let e = PEq (BufLength (AbstractBuf "esc_buf_CugleWOwEWXP")) (Lit 0)
      let simp = Expr.simplifyProp e
      let simpExpected = PEq (AbstractBuf "esc_buf_CugleWOwEWXP") (ConcreteBuf "")
      assertEqualM "buflen-to-empty" simp simpExpected
      ret <- checkEquivPropAndLHS e simpExpected
      assertBoolM "Must be equivalent" ret
```

Somehow we get a CEX, even though this simplification is good. I'll look into this. Triggered at: https://github.com/ethereum/hevm/actions/runs/13362022051/job/37313197906?pr=663 in PR #663 by @elopez

msooseth:
Oh wow, so this somehow is SAT:

```
   1   │ ;(PNeg
   2   │ ;  (PAnd
   3   │ ;    (PImpl
   4   │ ;      (PEq
   5   │ ;        (BufLength
   6   │ ;          (AbstractBuf "mybuf")
   7   │ ;        )
   8   │ ;        0
   9   │ ;      )
  10   │ ;      (PEq
  11   │ ;        (AbstractBuf "mybuf")
  12   │ ;        (ConcreteBuf "")
  13   │ ;      )
  14   │ ;    )
  15   │ ;    (PImpl
  16   │ ;      (PEq
  17   │ ;        (AbstractBuf "mybuf")
  18   │ ;        (ConcreteBuf "")
  19   │ ;      )
  20   │ ;      (PEq
  21   │ ;        (BufLength
  22   │ ;          (AbstractBuf "mybuf")
  23   │ ;        )
  24   │ ;        0
  25   │ ;      )
  26   │ ;    )
  27   │ ;  )
  28   │ ;)
```

which sounds wrong. buflength=0 should imply and empty buf, and an empty buf should emply buflength=0.

blishko:
Not sure what has changed, but both examples, from #801 and from this issue, work for me now on main.

msooseth:
Oh wow. OK, thanks! Let me close this issue then. I'd prefer a shorter issue list... this way we can't see the trees from the forest.

blishko:
Looks like this is actually still an issue. I am reopening.

PR #665, opened 2025-02-17
[DRAFT no review] Triage & fix the bug #664

msooseth:
## Description
Trying to triage and fix the bug as reported in #644 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
@blishko what do you think of this? I think we need it, as per #664 It's a kind of a small change, but it also requires some thought. Let me know what you think.

blishko:
I don't think this is entirely correct. Maybe we can have a short call to discuss this?
But at least the second implication seems suspicious to me. I can have a non-empty buffer where all values are zero, no?
Then it would be represented in SMT as `((as const Buf) #b00000000)` but the length would be non-empty.

msooseth:
Yes, you are right. It's a bit unclear to me, too, hence me asking your advice :) It also depends on SMT semantics. And this `as const` is not even defined right now :S Call would be good actually. Let me think about this a bit more then I'll ask you for a timeslot :) Thanks so much!

msooseth:
Issue seems to be gone, closing.

PR #667, opened 2025-02-18
New CopySlice rule that helps to avoid symbolic copyslice in case of overapproximation

msooseth:
## Description
This allows us to skip the symbolic copyslice. Basically it's copying over a EMPTY concretebuf values that have been copied from an abstractbuf and ONLY those values. So the result is just the abstractbuf! This helps a lot in avoiding sybmolic copyslice.

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Two more simplification rules: `ReadByte` & `ReadWord` when the `CopySlice`
   it is reading from is writing after the position being read, so the
   `CopySlice` can be ignored
-- One more simplification rule that helps avoid symbolic copyslice in case of
+- More simplification rules that help avoid symbolic copyslice in case of
   STATICCALL overapproximation
 - Test to make sure we don't accidentally overapproxmate a working, good STATICCALL
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `CopySlice` can be ignored
 - More simplification rules that help avoid symbolic copyslice in case of
   STATICCALL overapproximation
-- Test to make sure we don't accidentally overapproxmate a working, good STATICCALL
+- Test to make sure we don't accidentally overapproximate a working, good STATICCALL
 
 ## Fixed
 - We now try to simplify expressions fully before trying to cast them to a concrete value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Two more simplification rules: `ReadByte` & `ReadWord` when the `CopySlice`
   it is reading from is writing after the position being read, so the
   `CopySlice` can be ignored
+- One more simplification rule that helps avoid symbolic copyslice in case of
+  STATICCALL overapproximation
+- Test to make sure we don't accidentally overapproxmate a working, good STATICCALL
 
 ## Fixed
 - We now try to simplify expressions fully before trying to cast them to a concrete value
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -954,9 +954,9 @@ simplify e = if (mapExpr go e == e)
 
     -- redundant CopySlice
     go (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x0) _ dst) = dst
-    go (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k1))
-      (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k2)) (AbstractBuf k3) _)
-      (ConcreteBuf "")) | k1 == k2 && k2 == k3 = (AbstractBuf k1)
+    go (CopySlice (Lit 0) (Lit 0) (BufLength k1)
+      (CopySlice (Lit 0) (Lit 0) (BufLength k2) k3 _)
+      (ConcreteBuf "")) | k1 == k2 && k2 == k3 = k1
 
     -- simplify storage
     go (SLoad slot store) = readStorage' slot store
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -954,9 +954,12 @@ simplify e = if (mapExpr go e == e)
 
     -- redundant CopySlice
     go (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x0) _ dst) = dst
-    go (CopySlice (Lit 0) (Lit 0) (BufLength k1)
-      (CopySlice (Lit 0) (Lit 0) (BufLength k2) k3 _)
-      (ConcreteBuf "")) | k1 == k2 && k2 == k3 = k1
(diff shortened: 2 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -954,6 +954,9 @@ simplify e = if (mapExpr go e == e)
 
     -- redundant CopySlice
     go (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x0) _ dst) = dst
+    go (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k1))
+      (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k2)) (AbstractBuf k3) _)
+      (ConcreteBuf "")) | k1 == k2 && k2 == k3 = (AbstractBuf k1)
```
blishko, line 959:
```suggestion
    go (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k))
      (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k)) (AbstractBuf k) _)
      (ConcreteBuf ""))  = (AbstractBuf k)
```

Also, it feels like this is just a consequence of more general simplification rules.
I will try to write them down here.

msooseth: This... actually doesn't work:

image

msooseth: It's... a Haskell thing. Don't ask me why :D So the k1..k3 and making them all equal is the only way this can be made to work, AFAIK.

msooseth: However... this inspired me to make it more generic!

src/EVM/Expr.hs
```diff
@@ -954,6 +954,9 @@ simplify e = if (mapExpr go e == e)
 
     -- redundant CopySlice
     go (CopySlice (Lit 0x0) (Lit 0x0) (Lit 0x0) _ dst) = dst
+    go (CopySlice (Lit 0) (Lit 0) (BufLength k1)
+      (CopySlice (Lit 0) (Lit 0) (BufLength k2) k3 _)
+      (ConcreteBuf "")) | k1 == k2 && k2 == k3 = k1
```
blishko, line 959: So, I have a suggestion to split this into two, more general rules.

First is to remove redundant nested `CopySlice`:
`(CopySlice srcOff2 dstOff2 size2 (CopySlice srcOff1 dstOff1 size1 src _) dst) | size1 == size2 && srcOff2 == dstOff1 = CopySlice srcOff1 dstOff2 size1 src dst)`

The second one the CopySlice into empty buffer:
`(CopySlice (Lit 0) (Lit 0) (BufLength src) src (ConcreteBuf "")) = src`

Your rule should be a consequence of these two.

msooseth: Thanks! I have now added this :) Looking great!

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Two more simplification rules: `ReadByte` & `ReadWord` when the `CopySlice`
   it is reading from is writing after the position being read, so the
   `CopySlice` can be ignored
+- One more simplification rule that helps avoid symbolic copyslice in case of
```
blishko, line 29:
```suggestion
- More simplification rules that help avoid symbolic copyslice in case of
```

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  STATICCALL overapproximation
+- Test to make sure we don't accidentally overapproxmate a working, good STATICCALL
```
blishko, line 31:
```suggestion
- Test to make sure we don't accidentally overapproximate a working, good STATICCALL
```

test/test.hs
```diff
@@ -2291,6 +2291,100 @@ tests = testGroup "hevm"
...
+         }
+        |]
+      let sig2 = Just (Sig "retFor(address)" [AbiAddressType])
+      (expr, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig2 [] defaultVeriOpts
+      putStrLnM $ "successfully explored: " <> show (Expr.numBranches expr) <> " paths"
+      assertBoolM "The expression is NOT error" $ not $ any isError ret
+    -- NOTE: below used to be symbolic copyslice copy error before new copyslice simplification:
+    --       go (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k1))
+    --         (CopySlice (Lit 0) (Lit 0) (BufLength (AbstractBuf k2)) (AbstractBuf k3) _)
+    --         (ConcreteBuf "")) | k1 == k2 && k2 == k3 = (AbstractBuf k1)
```
blishko, line 2318: This comment should be updated.

test/test.hs
```diff
@@ -2291,6 +2291,100 @@ tests = testGroup "hevm"
...
+    , test "overapproximates-undeployed-contract" $ do
```
blishko, line 2319: Can this test be merged with the previous one?
It is doing the same work, no?
Only checks on the result are different.

test/test.hs
msooseth: Good point, fixed!

test/test.hs
msooseth: Actually, they are different, weirdly enough! The one above actually works without the simplifications. But this one does not. It's because Target was not deployed in this case, so it's abstracted over. Since we are not running a forge test, we don't have a setUp() and nothing is set up, so mm is actually not deployed. Weird, but true.

blishko: Oh, I see now.
Yeah, that's pretty weird.
The state variable `mm` is never assigned.
Won't the variable have always value `0`? Shouldn't we that call `mm.get` is always trying to call function on address `0`?

msooseth: Oh good point, it is trying to call address 0. But there is nothing at 0, so it overapproximates. Which is OK in this case, I think. Although we could emit a warning for call to address zero. Let me create an issue to track this. Would be hepful to emit a warning probably. It can happen, but in this case, we can't really do much about it. However, it is likely unintended.

In the meanwhile, I'm merging, with a note that this is address 0, and the contract is obviously not deployed there...

msooseth: Documented here: https://github.com/ethereum/hevm/issues/671

msooseth:
Thanks for the review! Let me think about my idea to track that zero address as an issue to be addressed separately and then we can hopefully merge this :)

test/test.hs
blishko: One last note, sorry for distracting you.
But, if we know the call is to zero address, we know that it will revert, no?

I don't know if I understand correctly how EVM works, but the call will lead to revert, so the assertion will not be reached, no?
It seems to me from your description that the counterexamples show how the assertion can be reached, but these are spurious, no?

Let me know if I am misunderstanding something.

msooseth: Yeah, good point. This is getting to be bigger than I wanted... can we merge this and document that as part of #671 ?

msooseth: It's getting to be too much to keep in mind and work on at the same time...

blishko: OK!

blishko (approved):
LGTM!

PR #668, opened 2025-02-18
Fix printing during tests, add proper asserts

msooseth:
## Description
This fixes printing during test case running, and also adds asserts where there were none. So it's actually cleaner and better test running.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
LGTM!

PR #673, opened 2025-03-03
More symbolic overapproximation when the remote contract cannot be fetched, adding CodeHash SMT representation

msooseth:
## Description
This allows us to overapproximate when the contract is real, but cannot be fetched. Meant to be on top of #658 i.e. `symbolic-extcodesize`. It allows us to overapproximate also when the contract cannot be fetched with RPC for opcodes:
* Balance
* CodeHash

I also added `CodeHash`  SMT representation so concrete addresses that cannot be fetched can be abstracted away and e.g. equivalence-checked. Tests have been added that exercise this very functionality.

See opcodes for reference:
* https://www.evm.codes/?fork=cancun#31
* https://www.evm.codes/?fork=cancun#3f

Debug code has also been added to show when the system is trying to fetch data via RPC. This helps in e.g. debugging. We probably at one point want to collect this data, which would allow us to fully and correctly replay things in tests. But that's a bit far away...

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   but allows much more thorough exploration of the contract
 - Allow controlling the max buffer sizes via --max-buf-size to something smaller than 2**64
   so we don't get too large buffers as counterexamples
+- More symbolic overapproximation for Balance and ExtCodeHash opcodes, fixing
+  CodeHash SMT representation
 
 
 ## Fixed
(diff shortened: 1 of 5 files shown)


blishko (approved):
Looks OK to me, though again, I still lack the knowledge to understand the fallbacks.

msooseth:
@blishko thanks for the review! @ggrieco-tob has kindly reviewed the EVM part of it, and it seems sane to him, so I'm merging :)

PR #674, opened 2025-03-04
Limiting the branching factor and the depth of branching

msooseth:
## Description

This is a rewrite of the branch width limitation, along with a new branch depth limitation. It should make things a _lot_ cleaner and tidier. Less noise, more sense. Added some tests, too.

## Checklist

- [x] tested locally
- [x] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -325,13 +325,13 @@ interpret fetcher maxIter askSmtIters heuristic vm =
         when (length vals < 2) $ internalError "PleaseRunAll requires at least 2 branches"
         frozen <- liftIO $ stToIO $ freezeVM vm
         let newDepth = vm.exploreDepth+1
-        threads <- withRunInIO $ \runInIO -> mapConcurrently (runInIO . runOne frozen newDepth) vals
-        pure $ goITE (zip vals threads)
+        ends <- withRunInIO $ \runInIO -> mapConcurrently (runInIO . runOne frozen newDepth) vals
+        pure $ goITE (zip vals ends)
         where
           goITE :: [(W256, Expr End)] -> Expr End
           goITE [] = internalError "goITE: empty list"
-          goITE [(_, e)] = e
-          goITE ((v,e):ps) = ITE (Eq expr (Lit v)) e (goITE ps)
+          goITE [(_, end)] = end
+          goITE ((val,end):ps) = ITE (Eq expr (Lit val)) end (goITE ps)
           runOne :: App m => VM 'Symbolic RealWorld -> Int -> W256 -> m (Expr 'End)
           runOne frozen newDepth v = do
             (ra, vma) <- liftIO $ stToIO $ runStateT (continue v) frozen { result = Nothing, exploreDepth = newDepth }


src/EVM.hs
```diff
@@ -1010,7 +1011,7 @@ exec1 conf = do
                 let cost = if acc then 0 else g_cold_account_access
                     funds = this.balance
                     recipientExists = accountExists xTo vm
-                branch (Expr.iszero $ Expr.eq funds (Lit 0)) $ \hasFunds -> do
+                branch (?conf.maxExplore) (Expr.iszero $ Expr.eq funds (Lit 0)) $ \hasFunds -> do
```
blishko, line 1014: What's `?conf.maxEplore`? What does the question mark mean?

src/EVM/Effects.hs
```diff
@@ -46,6 +46,7 @@ data Config = Config
   , maxBranch     :: Int
   , promiseNoReent   :: Bool
   , maxBufSize       :: Int
+  , maxExplore     :: Maybe Int
```
blishko, line 49: Can you explain what is the difference between `maxBranch` and `maxExplore`?

test/test.hs
```diff
@@ -1517,8 +1519,58 @@ tests = testGroup "hevm"
...
-      test "symbolic-copyslice" $ do
+      , testCase "limit-num-explore-hit-limit" $ do
+        let conf = testEnv.config {maxExplore = Just 3}
+        let myTestEnv :: Env = (testEnv :: Env) {config = conf :: Config}
+        runEnv myTestEnv $ do
+          Just c <- solcRuntime "C"
+            [i|
+            contract C {
+                function checkval(uint256 a, uint256 b, uint256 c) public {
+                  if (a ==b) {
```
blishko, line 1530:
```suggestion
                  if (a == b) {
```

test/test.hs
```diff
@@ -1517,8 +1519,58 @@ tests = testGroup "hevm"
...
+      -- below we don't hit the limit of the depth of the symbolic execution tree
+      , testCase "limit-num-explore-no-hit-limit" $ do
+        let conf = testEnv.config {maxExplore = Just 4}
+        let myTestEnv :: Env = (testEnv :: Env) {config = conf :: Config}
+        runEnv myTestEnv $ do
+          Just c <- solcRuntime "C"
+            [i|
+            contract C {
+                function checkval(uint256 a, uint256 b, uint256 c) public {
+                  if (a ==b) {
```
blishko, line 1556:
```suggestion
                  if (a == b) {
```

src/EVM.hs
blishko: I guess this is the implicit parameter thing, right?

msooseth: Yes! It's the implicit parameter thing. Kinda annoying, but easier than passing `maxExplore` everywhere else...

src/EVM/Effects.hs
msooseth: Good point actually! Thanks, you are really good at this. This needs clarification, and likely some rewrite, you are right. It's a mish-mash. Let me think about it a bit!

msooseth: OK, I now explained it better in `--help`  and I wrote a description in the `common-options.md` as well. Hopefully it's clearer now!

msooseth:
@blishko First of all, thanks for the review! I think it helped making this better :) Can you maybe take another peek at this? I tried clarifying the issues you found, hopefully it's in a better state now.

doc/src/common-options.md
```diff
@@ -48,3 +48,19 @@ Loops in the code cause a challenge to symbolic execution framework. In order
...
+encountered. For example, if a JUMP instruction is called with a symbolic
+expression, the system will explore all possible valid jump destinations,
+which may be too many. This option limits the branching factor in these cases.
+
+## Maximum Exploration Limt, ``--max-explore``
+
+Limits the branching factor and branching depth of symbolic exploration. This
+is a heuristic way to prevent the exploration from running for too long. Useful
+in scenarios where you use e.g. both symbolic execution and fuzzing, and don't
+want the symbolic execution to run for too long. It will often read to WEARNING-s
```
blishko, line 64:
```suggestion
want the symbolic execution to run for too long. It will often read to WARNING-s
```

doc/src/common-options.md
```diff
@@ -48,3 +48,19 @@ Loops in the code cause a challenge to symbolic execution framework. In order
...
 number can be set via the `--ask-smt-iterations` flag.
+
+## Maximum Branch Limit, ``--max-branch``
+
+Limits the number of branches that are explored in case a symbolic value is
+encountered. For example, if a JUMP instruction is called with a symbolic
+expression, the system will explore all possible valid jump destinations,
+which may be too many. This option limits the branching factor in these cases.
+
+## Maximum Exploration Limt, ``--max-explore``
```
blishko, line 59:
```suggestion
## Maximum Exploration Limit, ``--max-explore``
```

blishko (commented):
Looks OK to me, though I would like to see some documentation on why you sometime pass the limit to `branch`, and sometime you explicitly disable it by passing `Nothing`.
How should one decide when do one or the other?

msooseth:
> Looks OK to me, though I would like to see some documentation on why you sometime pass the limit to `branch`, and sometime you explicitly disable it by passing `Nothing`. How should one decide when do one or the other?

Actually, very good point. Thank you for that! It turns out that I did that only because I didn't have `conf.maxExplore` in context, so I cheated... It's just wrong. I'll fix it. Thank you, this has been super-helpful, as always!

blishko (approved):
I still lack sufficient knowledge to have a good picture how the exploration proceeds.
But I don't see any obvious issues, so we can try to proceed :)

msooseth:
> I still lack sufficient knowledge to have a good picture how the exploration proceeds. But I don't see any obvious issues, so we can try to proceed :)

I re-wrote it so hopefully it will be easier to review. It really ought to be easy to review, if not, I likely messed something up... so I decided to rewrite the multi-solution system to be much more clear and straightforward. This should make it so much easier to follow the control flow. The patterns of use of branching should also be much clearer and more consistent now. Honestly, previously, it was a mess, and likely also wrong. We now have a much cleaner --max-width and --max-depth flags, and they do what they say on the lid.

src/EVM/Types.hs
```diff
@@ -609,6 +611,10 @@ data Query t s where
 data RunBoth s where
   PleaseRunBoth    :: Expr EWord -> (Bool -> EVM Symbolic s ()) -> RunBoth s
 
+-- | Execution could proceed down one of two branches
```
blishko, line 614:
```suggestion
-- | Execution could proceed down one of several branches
```

src/EVM/SymExec.hs
```diff
@@ -321,13 +321,29 @@ interpret fetcher maxIter askSmtIters heuristic vm =
         conf <- readConfig
         (r, vm') <- liftIO $ stToIO $ runStateT (exec conf) vm
         interpret fetcher maxIter askSmtIters heuristic vm' (k r)
+      Stepper.ForkMany (PleaseRunAll expr vals continue) -> do
+        when (length vals < 2) $ internalError "PleaseRunAll requires at least 2 branches"
+        frozen <- liftIO $ stToIO $ freezeVM vm
+        let newDepth = vm.exploreDepth+1
+        threads <- withRunInIO $ \runInIO -> mapConcurrently (runInIO . runOne frozen newDepth) vals
+        pure $ goITE (zip vals threads)
```
blishko, line 329: Is `threads` the best name? The type of the elements is `Expr End`, right?
So aren't these really `branches`?

Also, can you comment for my understanding what is going to happen in Haskell runtime?
Is it going to spawn all threads immediately, or is there some thread pool with a limit on the number of threads that can run concurrently?

blishko (commented):
I like the new design! Still have some minor questions, though.

src/EVM/SymExec.hs
msooseth: Yes, true. They are branches. I'll rename.

It will spawn them all immediately. Unfortunately, there is no thread pool. Yes, this can actually lead to issues, but this has been known already. I mentioned this to Lexi, but they were quite sure it's fine. And honestly, this thread issue has not caused any problems (they could, eating up memory), so Lexi was right :)

PR #675, opened 2025-03-04
Reading code plain from a file instead of JSON

msooseth:
## Description
It's actually better to read plain code rather than JSON. Let's just do that. A simple `jq '.deplayedBytecode.object file.json > file.txt'` will do the trick. In order to do that, I have now stripped: empty space, newline, and '"'. This means that the above `jq` will just work.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warnings now lead printing FAIL. This way, users don't accidentally think that
   their contract is correct when there were cases/branches that hevm could not
   fully explore. Printing of issues is also now much more organized
+- We now read file as bytecode rather than JSON, as it's more compatible
+  with different input formats
 
 ## [0.54.2] - 2024-12-12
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Warnings now lead printing FAIL. This way, users don't accidentally think that
   their contract is correct when there were cases/branches that hevm could not
   fully explore. Printing of issues is also now much more organized
+- We now read file as bytecode rather than JSON, as it's more compatible
+  with different input formats
```
blishko, line 51: Instead of adding a "changed" entry, I think we should update the "added" entry about reading from JSON file.

blishko (changes requested):
Apart from the changelog, I think this is good!

CHANGELOG.md
msooseth: Good point. Sorry. Let me fix it :)

msooseth: Just fixed it, thanks! :)

PR #676, opened 2025-03-04
Warn user on zero address use

msooseth:
## Description
This attempts to fix https://github.com/ethereum/hevm/issues/671 Still needs work, because that test in particular doesn't fire. We need to add a verbosity flag, otherwise libraries that use us will have all sorts of junk printed on the console. Default is zero (is it's seamless for libraries), but we set it to 1 in all our use-cases (cli/testing) for the moment.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Effects.hs
+++ b/src/EVM/Effects.hs
@@ -46,6 +46,7 @@ data Config = Config
   , maxBranch     :: Int
   , promiseNoReent   :: Bool
   , maxBufSize       :: Int
+  , verb        :: Int
   }
   deriving (Show, Eq)
 
@@ -62,6 +63,7 @@ defaultConfig = Config
   , maxBranch = 100
   , promiseNoReent = False
   , maxBufSize = 64
+  , verb = 0
   }
 
 -- Write to the console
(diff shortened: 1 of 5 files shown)


src/EVM/Effects.hs
```diff
@@ -46,6 +46,7 @@ data Config = Config
   , maxBranch     :: Int
   , promiseNoReent   :: Bool
   , maxBufSize       :: Int
+  , verb        :: Int
```
blishko, line 49: Should this be aligned with the other options? Also `maxBranch` seems like it is standing out.

blishko (approved):
Looks good to me!

src/EVM/Effects.hs
msooseth: OOops, yes, let me fix.

msooseth:
@blishko  Oh wow, I just realized I went down the wrong path here -- there was already a `verbose` flag. Let me try  to do that instead. I'm closing this and re-opening a new one.

PR #684, opened 2025-03-13
Canonicalize all commutative operators in Expr

msooseth:
Canonicalizes all commutative operators in Expr. Much more powerful than some of the rewrite rules we had before.

This helps us with Eq x1 x1. In cases when x1 == x2 modulo commutativity, the Eq x1 x2 could actually be dispatched to the SMT solver. However, if we canonicalize x1 and x2, then this will not happen. This significantly improves solving. Note that this canonicalization is 100% compatible with us wanting Lit as the 1st argument. Lit is the lowest in the order, and so will always be the 1st.

This also means that some old rewrite rules need to change. The Add (Add ((Lit x) y) (Lit z)) can no longer stay that way, as Lit is smaller than Add, so it becomes Add (Lit x) (Add (Lit z) y. This is a good thing, because it brings all Lit-s to the LHS. This is now done for all commutative operators. The new setup is cleaner and also more principled than before. I adjusted the tests, and they showcase quite well that we are doing good indeed.

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


src/EVM/Expr.hs
```diff
@@ -708,7 +718,7 @@ pattern MappingSlot idx key = Keccak (WriteWord (Lit 0) key (ConcreteBuf idx))
 -- storage slots for arrays are determined by (keccak(bytes32(id)) + offset)
 -- note that `normArgs` puts the Lit as the 2nd argument to `Add`
```
blishko, line 719: Isn't this comment outdated? `Lit` should be the first argument, no? Not the second.

blishko (approved):
LGTM!

src/EVM/Expr.hs
msooseth: Oh wow, thanks for spotting this! Yes, definitely wrong. Fixing.

PR #686, opened 2025-03-18
Use single `conf.verb`, warn users on zero-address use

msooseth:
## Description
This attempts to fix https://github.com/ethereum/hevm/issues/671

We need to add a general verbosity flag, otherwise libraries that use us will have all sorts of junk printed on the console. Default is zero (is it's seamless for libraries), but we set it to 1 when running from the command line.

Note: `hevm test`'s `--verbosity` flag has been rolled into this, and it is now part of `App` rather than `UnitTestOptions`. This makes it more general, and makes it such that global options are truly in a single place (`App`).

The `hevm test --verbose X` seems to have been broken. When it was `Nothing`, then nothing extra was printed, but when it was `Just` then stuff got printed. However, the description seems to have talked about a difference between `1` and `2` levels. This doesn't seem to exist. Now it's set to print more failure info when it's `2` or above. This preserves the default "silent" operation when running from the command line (where verbosity is 1 by default).

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/doc/src/test.md
+++ b/doc/src/test.md
@@ -2,7 +2,7 @@
 
 ```
 Usage: hevm test [--root STRING] [--project-type PROJECTTYPE] [--rpc TEXT]
-                 [--number W256] [--verbose INT] [--coverage] [--match STRING]
+                 [--number W256] [--coverage] [--match STRING]
                  [--solver TEXT] [--num-solvers NATURAL] [--smtdebug] [--debug]
                  [--trace] [--ffi] [--smttimeout NATURAL]
                  [--max-iterations INTEGER]
@@ -16,7 +16,6 @@ Available options:
   --project-type PROJECTTYPE Foundry or CombinedJSON project
   --rpc TEXT               Fetch state from a remote node
   --number W256            Block: number
-  --verbose INT            Append call trace: {1} failures {2} all
   --coverage               Coverage analysis
   --match STRING           Test case filter - only run methods matching regex
   --solver TEXT            Used SMT solver: z3 (default), cvc5, or bitwuzla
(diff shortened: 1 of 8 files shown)


src/EVM/UnitTest.hs
```diff
@@ -298,37 +297,20 @@ indentLines n s =
...
-  in if (v > 1) then
-    mconcat
-      [ "Success: "
-      , fromMaybe "" (stripSuffix "()" testName)
-      , "\n"
-      , if (v > 2) then indentLines 2 (showTraceTree dapp vm) else ""
-      , indentLines 2 (formatTestLogs dapp.eventMap vm.logs)
-      , "\n"
-      ]
-    else ""
```
msooseth, line 316: Not used anywhere, removing.

test/EVM/Test/Utils.hs
```diff
@@ -59,7 +59,6 @@ testOpts solvers root buildOutput match maxIter allowFFI rpcinfo = do
     , askSmtIters = 1
     , smtTimeout = Nothing
     , solver = Nothing
-    , verbose = Just 1
```
msooseth, line 62: This is now `conf` and in cli.hs is set to by default 1. In API library use mode it's set to 0 by default.

doc/src/test.md
```diff
@@ -2,7 +2,7 @@
 
 ```
 Usage: hevm test [--root STRING] [--project-type PROJECTTYPE] [--rpc TEXT]
-                 [--number W256] [--verbose INT] [--coverage] [--match STRING]
+                 [--number W256] [--coverage] [--match STRING]
```
blishko, line 5: Should `verb` be added here? Or why is that not necessary?

msooseth: I'm actually thinking about removing this part entirely. It's trivial to get via `--help` and it's likely also to be more accurate. What do you think? I wanna do that as a follow-up PR. I don't think this reference is very helpful. The tutorial, and other parts are way more useful. The value-add here is minimal. I have seen people navigate to this and then be confused how it's helpful....

blishko (approved):
LGMT!

doc/src/test.md
msooseth: I have now written a PR to significantly improve our documentation, here #690

PR #687, opened 2025-03-19
fixed maxBufSize breaking bug

scottbuckley:
## Description
Recent changes that introduced the maxBufSize flag broke the `version` and `exec` modes, due to a naive bounds check on the input, where that input (nor a default value) does not exist for the above modes. `maxBufSize` is a partial selector being used on a sum type, and it's being evaluated on non-matching types during CLI argument parsing.

The way this bug presented was the error `hevm: No match in record selector maxBufSize`, when running `hevm version` or `hevm exec ...`, using the `hevm` binary I installed on my OSX machine (under Rosetta) via Nix pointing at this repo.

I fixed this bug by adding a total function `getMaxBufSize` which returns the default value 64 for `version` and `exec` modes. This is now used instead of `.maxBufSize` in three places. There will certainly be other ways to fix the same problem; I just did it this way to get it to work for me, so I'm sharing the result.

I was a bit surprised to see that your tests didn't catch this error, but I didn't want to dive into this to figure it out. I'd recommend adding tests that would have caught this kind of bug.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/cli/cli.hs
+++ b/cli/cli.hs
@@ -221,13 +221,18 @@ getFullVersion = showVersion Paths.version <> " [" <> gitVersion <> "]"
       Right val -> "git rev " <> giBranch val <>  "@" <> giHash val
       Left _ -> "no git revision present"
 
+getMaxBufSize :: Int -> Command Options.Unwrapped -> Int
+getMaxBufSize def (Version {}) = def
+getMaxBufSize def (Exec {}) = def
+getMaxBufSize _ cmd = cmd.maxBufSize
+
 main :: IO ()
 main = withUtf8 $ do
   cmd <- Options.unwrapRecord "hevm -- Ethereum evaluator"
-  when (cmd.maxBufSize > 64) $ do
+  when (getMaxBufSize 64 cmd > 64) $ do
     putStrLn "Error: maxBufSize must be less than or equal to 64. That limits buffers to a size of 2^64, which is more than enough for practical purposes"
     exitFailure
-  when (cmd.maxBufSize < 0) $ do
+  when (getMaxBufSize 64 cmd < 0) $ do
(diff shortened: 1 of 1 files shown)


msooseth:
Wow, that was a bad. Thank you so much. The fact that there are no tests for the executable (at all :cry:) is something that has been bugging me for a while. I wanted to add them before. I will now prioritize trying to figure out how to do it well. I wanted to use python `lit`, which is what llvm is using (and I have been using in many of my projects). I'll create a PR with it and see if it's acceptable.

Thank you again!

msooseth:
@scottbuckley BTW, hevm has improved its symbolic execution part significantly over the past months. You may wanna give it a go :) It really is a different beast!

scottbuckley:
> @scottbuckley BTW, hevm has improved its symbolic execution part significantly over the past months. You may wanna give it a go :) It really is a different beast!

I will give it a try for sure! I'm trying to move from PL / formal methods research into blockchain verification so I'm just familiarising myself with the existing tools lately. Actually this is a little off-topic, but while I have you - can you point me to documentation on how to run the interactive debugger? I was following the documentation from the dapptools repo on how to interactively debug a live transaction, with `hevm exec ... --debug` (with other args pulled in via `seth`), but on this repo's hevm this doesn't seem to trigger interactive debugging. Is there documentation somewhere about this change and about how I could access such a feature on the current version? Sorry, I know this isn't the venue for such a question, but it was the reason I wanted to get hevm running in the first place :)a

msooseth:
Sorry, the debugger has been removed. We may add it back... one day? The changelog says: `- The visual debugger has been removed`. It was actually very cool. But no longer. It was a bit hard to maintain but I actually kinda loved it.

msooseth:
When you give `hevm` a go, please let us know of your thoughts. Would be nice to hear some interesting feeback!

Issue #688, opened 2025-03-19
`hevm equivalence` reports no discrepancies on two bytecodes that compute different values


konnov:
Hi! Perhaps I am expecting too much of hevm. I feed it two bytecodes: One computes a value via `SAR`, and another one computes a value via `SDIV`. They do compute slightly different values, but `hevm` reports no discrepancies.

```sh
$ cat bytecode1
7fe7f7981242e0f50016a573e27a65b63c8fd75305b4b504a70fd5c21f7d9ea4fd60161d
$ cat bytecode2
7fe7f7981242e0f50016a573e27a65b63c8fd75305b4b504a70fd5c21f7d9ea4fd7f00000000000000000000000000000000000000000000000000000000004000009005
$ hevm equivalence --code-a `cat bytecode1` --code-b `cat bytecode2`
Found 1 total pairs of endstates
Asking the SMT solver for 1 pairs
Reuse of previous queries was Useful in 0 cases
No discrepancies found
```

When I run the two bytecodes on [evm.codes](https://evm.codes), they compute two different values. This behavior seems to be consistent with [`bitwise_sar`](https://github.com/ethereum/execution-specs/blob/1a587803e3e698407d204888b02342393f8b4fe5/src/ethereum/cancun/vm/instructions/bitwise.py#L213) and [`sdiv`](https://github.com/ethereum/execution-specs/blob/1a587803e3e698407d204888b02342393f8b4fe5/src/ethereum/cancun/vm/instructions/arithmetic.py#L145) of EELS.

msooseth:
Thanks! Definitely reproducible on `main`. However:

```
cabal run exe:hevm equivalence -- --code-a "7f" --code-b "7fe7"
```

Also reports equivalent. But:

```
cabal run exe:hevm equivalence -- --code-a "600060015260406000f3" --code-b "600a60015260406000f3"
```

Reports fail. There reason for this is that we compare what would happen when the contract is called. The contract you put forth doesn't return a value.

If I add an MSTORE&RETURN to your bytecodes, we get:

```
cabal run exe:hevm equivalence -- --code-a "7fe7f7981242e0f50016a573e27a65b63c8fd75305b4b504a70fd5c21f7d9ea4fd60161d60205260406000f3" --code-b "7fe7f7981242e0f50016a573e27a65b63c8fd75305b4b504a70fd5c21f7d9ea4fd7f0000000000000000000000000000000000000000000000000000000000400000900560205260406000f3"
[...]
[...]

PR #690, opened 2025-03-19
[DRAFT] Improve docs

msooseth:
## Description
An attempt at improving the documentation

Related to #688 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Ah, that's actually a good point. Let me add that :)

msooseth:
Hahaha I can't... Options.Command is not made for human consumption. I can add it, but only to the main `--help`, not the sub-help of e.g. `symbolic`, and it also removes the detailed help... weird.

msooseth:
> Thanks a lot for improving the documentation. The explanation of what constitutes equivalence definitely helps! It would be also great, if `--help` had a pointer to the documentation. It would remind people like me, who are trying to check something in a rush, to check the docs.

That's actually a great idea, and I have tried adding that line and it's been 3h of struggling with `Options`. We use `Options` in a rather special way. And that doesn't seem to allow that. I think I will have to rewrite our cli.hs and then it will work. For the moment, I'll leave it as-is and will create a new PR to add that help line at the bottom :)

doc/src/equivalence.md
```diff
@@ -1,53 +1,149 @@
...
-from solidity contracts can be done via, e.g.:
+to prove equivalence between their outputs and storages. For a full listing of
+options, see `hevm equivalence --help`.
 
+## Simple example usage
+
+```shell
+$ solc --bin-runtime "contract1.sol" | tail -n1 > a.bin
+$ solc --bin-runtime "contract2.sol" | tail -n1 > b.bin
+$ hevm equivalence --code-a-file a.txt --code-b-file b.txt
```
blishko, line 20:
```suggestion
$ hevm equivalence --code-a-file a.bin --code-b-file b.bin
```

doc/src/equivalence.md
```diff
@@ -1,53 +1,149 @@
...
+
+## Calldata size limits
+
 If `--sig` is given, calldata is assumed to take the form of the function
 given. If `--calldata` is provided, a specific, concrete calldata is used. If
 neither is provided, a fully abstract calldata of at most `2**64` byte is
 assumed. Note that a `2**64` byte calldata would go over the gas limit, and
-hence should cover all meaningful cases.
+hence should cover all meaningful cases. You can limit the buffer size via
+`--max-buf-size`, which sets the exponent of the size, i.e. 10 would limit the
```
blishko, line 30: Should we call this option `--max-calldata-buf-size`?

doc/src/equivalence.md
```diff
@@ -1,53 +1,149 @@
...
-  --loop-detection-heuristic LOOPHEURISTIC
-                           Which heuristic should be used to determine if we are
-                           in a loop: StackBased (default) or Naive
-                           (default: StackBased)
+```plain
+Usage: hevm equivalence [--code-a TEXT] [--code-b TEXT] [--code-a-file STRING]
+                        [--code-b-file STRING] [--sig TEXT] [--arg STRING]...
+                        [--calldata TEXT] [--smttimeout NATURAL]
+                        [--max-iterations INTEGER] [--solver TEXT]
+                        [--num-solvers NATURAL] ...
```
blishko, line 8: What does `...` mean here?

doc/src/symbolic.md
```diff
@@ -1,83 +1,24 @@
...
-                           Number of times we may revisit a particular branching
-                           point before we consult the smt solver to check
-                           reachability (default: 1) (default: 1)
-  --num-solvers NATURAL    Number of solver instances to use (default: number of
-                           cpu cores)
-  --loop-detection-heuristic LOOPHEURISTIC
-                           Which heuristic should be used to determine if we are
-                           in a loop: StackBased (default) or Naive
-                           (default: StackBased)
+                     [--number W256] [--timestamp W256] [--basefee W256] ..
```
blishko, line 7: I guess you forgot one more dot? :)

blishko (commented):
Nice! Just left a few questions.

doc/src/equivalence.md
msooseth: Ah wait... this actually applies to other buffers too, like here:

```haskell
symCalldata :: App m => Text -> [AbiType] -> [String] -> Expr Buf -> m (Expr Buf, [Prop])
symCalldata sig typesignature concreteArgs base = do
  conf <- readConfig
  let
    args = concreteArgs <> replicate (length typesignature - length concreteArgs) "<symbolic>"
    mkArg :: AbiType -> String -> Int -> CalldataFragment
    mkArg typ "<symbolic>" n = symAbiArg (T.pack $ "arg" <> show n) typ
    mkArg typ arg _ =
      case makeAbiValue typ arg of
        AbiUInt _ w -> St [] . Lit . into $ w
        AbiInt _ w -> St [] . Lit . unsafeInto $ w
        AbiAddress w -> St [] . Lit . into $ w
        AbiBool w -> St [] . Lit $ if w then 1 else 0
        _ -> internalError "TODO"
    calldatas = zipWith3 mkArg typesignature args [1..]
    (cdBuf, props) = combineFragments calldatas base
    withSelector = writeSelector cdBuf sig
    sizeConstraints
      = (Expr.bufLength withSelector .>= cdLen calldatas)
      .&& (Expr.bufLength withSelector .< (Lit (2 ^ conf.maxBufSize)))
```

and also the data that an unknown contract sends back to us. We should explain this...

msooseth: OK, I am gonna talk about this more in the docs and try to give a better explanation in the `--help`, too.

doc/src/equivalence.md
msooseth: I was trying to imply that there are more options... maybe I should remove? I don't want to list them all, I think it's not that useful? What do you think?

doc/src/symbolic.md
msooseth: Hahah fixed :)

msooseth:
Let me write a `common-options.md` to document some of the common options such as `maxBufSize`.

doc/src/equivalence.md
blishko: Why not document all the options?
It seems unusual to omit some of them.
Unless you can argue they belong to separate categories, like "for users" and "for developers"?
And you want to list only those important for the users.

msooseth: Well, we keep adding some and then not documenting them in the MD docs.  I am wondering what' the best thing to do here. Of course I can run `--help` and copy-paste but I'm not sure that's the best. I am trying to improve the documentation, not make it perfect :D Currently, it's quite out of date. I am not sure what's best. The full `--help` is also way too verbose and I don't think it's too helpful. They can always run it themselves. What do you think? Should I just copy-paste the `--help` and then explain the options as it is now (with common-options.md to explain the common ones)?

blishko:
It's definitely an improvement :)

I don't have any strong opinion on how to document the options, as I am not great with documentation myself.

blishko (approved):
LGTM!

PR #691, opened 2025-03-20
CLI upgrade for more control and less duplication

msooseth:
## Description
We are now using `Options.Applicative` for CLI. This allows us to have a lot more control, e.g. separate footers for each subcommand. It also allows us to skip having repeated code that was becoming pretty unwieldy. The common commands are now much more visible/reasonable to edit and improve.

I also added some very simple test cases for the CLI, just to make sure I didn't break some basic CLI functionality. We probably should extend this test more for some basic checks to make sure new options work.

Notes:
* `solver` in `UnitTestOptions` was completely unused (oops!). Removed.
* I am using as little copy-paste as possible, but some options are special, and only appear in some subcommands. These are duplicated, though their parser is factored out
* I updated the authors to include all of us (I don't care about the order, so I put myself last. If you feel like you should be earlier in the list, please feel free to rearrange)
* The `functionAbi f` function is actually not very user-friendly. When I give a signature of `--sig "stuff"` it gives an `internalError`, because it's missing the `()`. It's now a _little_ bit better, but it should be rewritten to return an `Either` and give better feedback to the user.
* Updated build.yml to run `cli-test`
* I set `default-language` everywhere in the cabal file. This seems to be needed now, with the new `cli-test` depending on `hevm`. Weird, but otherwise `cabal check` complains. This will be needed anyway in the future when we upgrade to newer cabal.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (commented):
Looks pretty cool!

I have not spotted any problem. I will try to play with the CLI as well :)

hevm.cabal
```diff
@@ -185,11 +187,13 @@ library
     split                             >= 0.2.3 && < 0.3,
     template-haskell                  >= 2.19.0 && < 3,
     extra                             >= 1.7.14 && < 2,
+    hspec                             >= 2.9.0 && < 2.13.0,
```
blishko, line 190: I am getting a warning from the compiler that these packages are not needed.
This is config the `hevm` library, right? The CLI changes should not be affecting the library, no?

blishko (commented):
Looks good! Can you check if these new package dependencies should really be dependencies for the library target?

hevm.cabal
msooseth: Nice catch, as always! Sorry, we disabled these warnings to become errors, because of an annoying warning that can't be fixed (` - system-cxx-std-lib-1.0 (exposed by flag -package-id system-cxx-std-lib-1.0`) and so I just normalized these warnings. Well-spotted!

Fixed :) I am merging, it seems to be working for my checks. I have read and re-read the code a few times, it seems fine to me.

PR #692, opened 2025-03-27
Symbolic block number is now allowed in library mode

msooseth:
## Description
This change only affects library mode. In this mode, we now allow the block number to be symbolic. This is helpful for Echidna. It's a very minor change, and should not affect any use-case beyond being used as a library.

Built on top of #691 . Needs #691 to be merged before it makes any sense.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog


blishko (approved):
I don't really understand much, but it seems OK :)

blishko:
My review is not much better than Copilot's 😅

msooseth:
Hahah, no, I think yours is actually helpful, indicates I need to do better. So I decided to add a test case, thanks to @gustavo-grieco who contributed it. Let me add it and then merge :)

msooseth:
Haha, the test case highlighted I forgot to change `roll()` to accept a symbolic block number. Now I added a test to check for this, and it sets the block number to a symbolic value, then tests it against a symbolic value. Now I feel like it's time to close :)

PR #693, opened 2025-03-27
[DRAFT] Better structuring of Keccak rewrites

msooseth:
## Description
There are some serious issues with our rewrite system. With this improvement, we get way better expressions. But it fails sometimes.

TODOs:
* Fix the failing test -- figure out what's going on
* Add the `cabal run exe:hevm test -- --solver bitwuzla --root tmp --match ".*prove_nested_append.*" --debug --smtdebug` test so we can test for this rewrite breaking. That test takes 100s of seconds without this rewrite
* add the full `simplify` for every place we use `readStorage` -- I think. This change MAY make these places less effective... but may make them more effective. I think we should be running `concKeccakSimp` in these places anyway.

nested_array.sol:

```
pragma solidity ^0.8.17;

contract MyContract is Test {
  uint[][] arr2;
  function prove_nested_append(uint v, uint w) public {
    arr2.push([v,w]);
    arr2.push();
    arr2.push();

    arr2[1].push(arr2[0][0]);

    arr2[2].push(w);
    // TODO: this is the problematic line!
    // changing this to arr2[1].push(1), and updating the assertion for the
    // value of arr2[1][1] to reflect the change makes this test pass...
    arr2[1].push(w);

    assert(arr2.length == 3);

    assert(arr2[0].length == 2);
    assert(arr2[0][0] == v);
    assert(arr2[0][1] == w);

    assert(arr2[1].length == 2);
    assert(arr2[1][0] == v);
    assert(arr2[1][1] == w);

    assert(arr2[2].length == 1);
    assert(arr2[2][0] == w);
  }
}
```

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


test/test.hs
```diff
@@ -354,7 +354,7 @@ tests = testGroup "hevm"
       let simpExpr = mapExprM Expr.decomposeStorage expr
       -- putStrLnM $ T.unpack $ formatExpr (fromJust simpExpr)
       assertEqualM "Decompose did not succeed." (isJust simpExpr) True
-    , test "decompose-6-fail" $ do
+    , expectFail $ test "decompose-6-fail" $ do
```
blishko, line 357: Can you explain again why this test now fails?
Why that would now be expected result?
If the test is still correct, we can temporarily disable it, but not change the expectation.

Same for the other test.

blishko (commented):
I have to say I am confused about this `keccak` to `Lit` and `Lit` to `keccak` transformations.

Can we somewhere document/explain why we need `Lit` -> `keccak` direction?
In some intuitive way?

test/test.hs
msooseth: Thanks, I fixed this and the other one, too, so they are both `ignoreTest` now. Thanks!

msooseth:
Regarding `Lit` -> `Keccak`. It's indeed a bit annoying. When there is an index into some constant value or similar map/array indexing, Solidity will not emit `Keccak (ConcreteBuf ...)`, instead it will compute the `Lit` and emit that. The issue is, in order to be able to know that we can skip that store/load during simplification, we match on `Keccak`. So we temporarily convert the `Lit `to a `Keccak (ConcreteBuf ...)` so the simplification can know that it'll never match.

blishko (approved):
LGTM!

PR #694, opened 2025-03-28
Add --arrays-exp to CVC5 options

zoep:
## Description

This PR adds `--arrays-exp` to CVC5 options. 

Without this option, CVC5 returns the following error on some queries generated by the equivalence: 

```
(error "Term of kind `STORE_ALL` not supported in default mode, try **`--arrays-exp`.")**`
```
Adding the option, prevents the failures. 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [X] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,7 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   determining when (Eq a b) is true when a==b modulo commutativity
 - `hevm test`'s flag ` --verbose` is now `--verb`, which also increases verbosity
   for other elements of the system
-- Add `--arrays-exp` to CVC5 options.
+- Add `--arrays-exp` to cvc5 options.
 
 ## [0.54.2] - 2024-12-12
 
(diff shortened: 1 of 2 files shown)


blishko:
Do you know what term is the problem? Is this related to the constant arrays?

msooseth:
Do you have the query file that gives that error? The one you sent me on chat seems to work without the option you gave. Maybe you are using an older cvc5 version? We have in hevm:

```
This is cvc5 version 1.2.0
compiled with GCC version 13.3.0
on Jan  1 1980 00:00:00
```

Maybe Act has an older cvc5 pinned?

blishko:
I confirm the query works in `cvc5 1.2.0`, but returns the error in `1.2.1`. I am going to ask about it in our channel with `cvc5` people.

blishko:
Just got a confirmation from `cvc5` that this was intentional.
"Hi Martin, this change was intentional since constant arrays are (still) non-standard and we try to hide all non-standard features behind flags that have to be explicitely enabled."

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   determining when (Eq a b) is true when a==b modulo commutativity
 - `hevm test`'s flag ` --verbose` is now `--verb`, which also increases verbosity
   for other elements of the system
+- Add `--arrays-exp` to CVC5 options.
```
blishko, line 79:
```suggestion
- Add `--arrays-exp` to cvc5 options.
```

blishko (approved):
It has been confirmed that we should indeed use this flag.

msooseth:
Thanks! It's great that it will keep working as cvc5 gets upgraded :)

PR #700, opened 2025-04-03
Unifying result type

msooseth:
## Description
`CheckSatResult` vs `ProofResult` has led to a bit of a mess, I think. With this rewrite, it should lead to a cleaner, and easier-to-read code, that is more obviously correct. This also means that in case we have a CopySlice issue, it doesn't get pushed into two different types, which could lead to this kind of thing:

```
      2x -> CopySlice with a symbolically sized region not currently implemented, cannot execute SMT solver on this query
      1x -> SMT result timeout/unknown: CopySlice with a symbolically sized region not currently implemented, cannot execute SMT solver on this query
```

Other changes:
* Another issue fixed is that `reachable` would throw an `internalError` when `reachable` would die due to timeout/symbolic copyslice issue. Now we just say that the leaf may be reachable. We also adjusted the printing so it says "Potentially Reachable" rather than "Reachable".
* Also, `Unknown` has been renamed `UnknownBranch`, since `Unknown` collided with other names. I think `UnknownBranch` is a clean, short-ish name. It actually makes it easier to read some functions, in my opinion.
* Aeson is no longer imported without a qualifier, as it's actually a large namespace and collides with some things. So it's now qualified as `JSON.`, which I think is reasonable.
* I had to use a typeclass to make `Show` work for `Unknown`, but it's actually not that crazy.
* `Qed` for probably historical reason had a `()` parameter at all times -- this was unneeded. I removed it, it's so a lot cleaner now.
* I also had to move some definitions to `Types`. However, I consider this a win, as the cyclical definition issue is now reduced -- `Types` can be imported from anywhere, without cyclical import issues.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Types.hs
+++ b/src/EVM/Types.hs
@@ -4,6 +4,8 @@
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE KindSignatures #-}
 
 {-# OPTIONS_GHC -Wno-inline-rule-shadowing #-}
 
@@ -15,7 +17,6 @@ import Control.Monad (mzero)
 import Control.Monad.ST (ST)
 import Control.Monad.State.Strict (StateT)
 import Crypto.Hash (hash, Keccak_256, Digest)
-import Data.Aeson
 import Data.Aeson qualified as JSON
 import Data.Aeson.Types qualified as JSON
 import Data.Bifunctor (first)
@@ -53,6 +54,7 @@ import Numeric (readHex, showHex)
 import Options.Generic
 import Optics.TH
 import EVM.FeeSchedule (FeeSchedule (..))
+import Data.Kind (Type)
 
 import Text.Regex.TDFA qualified as Regex
 import Text.Read qualified
@@ -616,7 +618,7 @@ data RunAll s where
   PleaseRunAll    :: Expr EWord -> [W256] -> (W256 -> EVM Symbolic s ()) -> RunAll s
 
 -- | The possible return values of a SMT query
-data BranchCondition = Case Bool | Unknown
+data BranchCondition = Case Bool | UnknownBranch
   deriving Show
 
 instance Show (Query t s) where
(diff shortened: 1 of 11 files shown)


src/EVM/Types.hs
```diff
@@ -15,7 +17,6 @@ import Control.Monad (mzero)
 import Control.Monad.ST (ST)
 import Control.Monad.State.Strict (StateT)
 import Crypto.Hash (hash, Keccak_256, Digest)
-import Data.Aeson
```
msooseth, line 18: This was interfering with some things through clashing with the namespace. I think it makes sense to use the `JSON.(whatever)` syntax instead, it's more explicit and it's in not too many places

src/EVM/Types.hs
```diff
@@ -1106,8 +1108,114 @@ data GenericOp a
...
+  x == y = case (x, y) of
+    (Unknown u1, Unknown u2) -> u1 == u2
+    (Error a, Error b)       -> a == b
+    (Cex a, Cex b)           -> a == b
+    (Qed, Qed)               -> True
+    _                        -> False
+
+type VerifyResult = ProofResult (Expr End, SMTCex) (String, Expr End)
+type EquivResult = ProofResult (SMTCex) String
+type SMT2Result = ProofResult (SMTCex) String
```
blishko, line 1191: I think the number two in there could be misleading a bit.  I don't see a need for it.
```suggestion
type SMTResult = ProofResult (SMTCex) String
```

blishko: I would personally prefer to have `Sat` and `Unsat` as the variants of `SMTResult`, though I understand that for `VerifyResult` and `EquivResult` the current terminology makes more sense.

I see SMT and symbolic execution/equivalence checking as working on two different levels, so for me it would be natural to have different representation for them and then some simple translation between the layers.

blishko (commented):
The refactoring is nice!
Personally, I would prefer to keep the SMT result and symbolic execution result separated.
The issue you point out with two different sources of CopySlice error seem to indicate there is not a proper separation between the SMT and the symbolic execution layer and this could be solved in a different way.
But I am also OK with the proposed change if you think this is better.

blishko (approved):
LGMT!

PR #701, opened 2025-04-04
Fix interpretation of buffer model value

blishko:
Previosuly, we would interpret any buffer whose model is a constant-zero array as an empty buffer, disregarding the value of the length of the buffer in the model.
However, the length must be taken into account.

Fun fact: The special case was only triggered if the zero value used hexadecimal format (which is used by Z3). For example, Bitwuzla prints using binary format, so this special handling was not triggered for Bitwuzla.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -4222,6 +4222,27 @@ tests = testGroup "hevm"
       assertEqualM "must not duplicate" simp (nubOrd simp)
       assertEqualM "We must be able to remove all duplicates" (length $ nubOrd simp) (length $ List.nub simp)
   ]
+  , testGroup "calling-solvers"
+  [
+    testCase "correct-model-for-empty-buffer" $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 0}}) $ do
+      withDefaultSolver $ \s -> do
+        let props = [(PEq (BufLength (AbstractBuf "b")) (Lit 0x0))]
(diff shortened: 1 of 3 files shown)


test/test.hs
```diff
@@ -4222,6 +4222,27 @@ tests = testGroup "hevm"
...
+    testCase "correct-model-for-empty-buffer" $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 0}}) $ do
+      withDefaultSolver $ \s -> do
+        let props = [(PEq (BufLength (AbstractBuf "b")) (Lit 0x0))]
+        (res, _) <- checkSatWithProps s props
+        cex :: SMTCex <- case res of
+          Sat c -> pure c
+          _ -> liftIO $ assertFailure "Must be satisfiable!"
+        let value = subModel cex (AbstractBuf "b")
+        assertEqualM "Buffer must be empty" (ConcreteBuf "") value
+    , testCase "correct-model-for-non-empty-buffer-of-all-zeroes" $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 0}}) $ do
```
blishko, line 4236: This test would previously fail, because we would compute the value as empty buffer.

msooseth (approved):

PR #702, opened 2025-04-07
Fix infinite recursion when trying to shrink buffer

blishko:
## Description

When we have a buffer that can only have length somewhere in the range [MAX_VALUE/2, MAX_VALUE] (for max value of W256), then the current computation of trying to find smaller feasible length goes into infinite loop, because the boundary we are trying overflows and starts from 0 again.
We now detect the overflow and stop the recursion to conclude that we are not able to shrink the buffer.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth (approved):
Nice, looking great actually!

PR #703, opened 2025-04-08
Avoid eager translation of buffer models

blishko:
## Description
The goal here is to keep the buffer representation in the compressed form as long as possible, and only lazily translate to the flat representation.
This helps to avoid raising `internal error` when the buffer length is too large to be converted to concrete `ByteString`.
Instead, as the output of the equivalence check we print the compressed representation. This information might still be helpful, definitely more helpful than crashing.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -4344,6 +4344,15 @@ tests = testGroup "hevm"
             Cex _ -> True
             _ -> False
         assertBoolM "Must be satisfiable!" sat
+    , testCase "can-get-value-unrelated-to-large-buffer" $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 0}}) $ do
+      withDefaultSolver $ \s -> do
+        let props = [(PEq (Var "a") (Lit 0x1)), (PGT (BufLength (AbstractBuf "b")) (Lit 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffeb4))]
+        (res, _) <- checkSatWithProps s props
+        cex :: SMTCex <- case res of
+          Cex c -> pure c
+          _ -> liftIO $ assertFailure "Must be satisfiable!"
+        let value = subModel cex (Var "a")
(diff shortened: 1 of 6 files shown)


msooseth (approved):
LGTM. It seems like you are delaying making it concrete to as late as possible, right? Which to me sounds very reasonable. Also, displaying something to the user that's hard(er) to interpret than an internalError is definitely a plus :)

blishko:
Indeed, I am trying to avoid creating `ConcreteBuf` for as long as possible.

msooseth:
Do you wanna add a test case? Perhaps to the `clitest.hs` or something?

test/test.hs
```diff
@@ -4344,6 +4344,15 @@ tests = testGroup "hevm"
...
         assertBoolM "Must be satisfiable!" sat
+    , testCase "can-get-value-unrelated-to-large-buffer" $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 0}}) $ do
+      withDefaultSolver $ \s -> do
+        let props = [(PEq (Var "a") (Lit 0x1)), (PGT (BufLength (AbstractBuf "b")) (Lit 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffeb4))]
+        (res, _) <- checkSatWithProps s props
+        cex :: SMTCex <- case res of
+          Cex c -> pure c
+          _ -> liftIO $ assertFailure "Must be satisfiable!"
+        let value = subModel cex (Var "a")
+        assertEqualM "Can get value out of model in the presence of large buffer!" value (Lit 0x1)
```
blishko, line 4355: This test fails on `main`, because it raises internal error about the buffer being too large.
Here it passes, because we don't inspect the buffer, so it does not need to be flattened.

msooseth: Oh nice! I like it :)

msooseth (approved):
I like this test!

PR #705, opened 2025-04-10
Improve documentation, README, and tutorial

msooseth:
## Description
Goals:
* Make the tutorial easier to follow, and correct. Now it's easier to run and use, as tested by Anja
* Readme is shorter and more to the point
* `architecture.md` is now easier to access, available at `hevm.dev` website rather than essentially hidden in the source code.
* Updated `architecture.md` to better fit what Expr is now, with an example from the research paper
* Reformatted markdown files when I edited them so it's easier to edit them in e.g. vim. Otherwise, editing them is quite painful
* Fixed some small things

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [x] updated the docs
- [ ] updated the changelog


doc/mermaid-init.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    for (const lightTheme of lightThemes) {
+        document.getElementById(lightTheme).addEventListener('click', () => {
+            if (!lastThemeWasLight) {
+                window.location.reload();
+            }
+        });
+    }
+})();
```
msooseth, line 35: Auto-generated by mermaid, which we need because we moved `architecture.md` into the documentation site from basically a hidden space in the hevm repo.

doc/mermaid.min.js
msooseth, line 1: Auto-generated by mermaid, which we need because we moved `architecture.md` into the documentation site from basically a hidden space in the hevm repo.

doc/src/SUMMARY.md
```diff
@@ -18,3 +17,8 @@
 - [Equivalence checking](./equivalence.md)
 - [Concrete execution](./exec.md)
 - [Common options](./common-options.md)
+
+# Development
+
+- [Developers tutorial](./for-hevm-developers.md)
+- [System architecture](./architecture.md)
```
msooseth, line 24: Moved development into its own sub-section, with the `architecture.md` moved here, so it's easier to access.

doc/src/architecture.md
msooseth, line 1: Slightly updated to match what Expr and hevm is now. The picture is new, and is from the paper we presented at CAV'2024. The paper and the presentation are also linked now.

doc/src/expr.png
msooseth, line 1: A screenshot of the paper that we presented, showing Expr, which I think works quite well for the architecture overview.

architecture.md
msooseth, line 1: Moved this file to `hevm.dev` website

msooseth: (and updated it to be more modern)

doc/mermaid.min.js
msooseth, line 1: Auto-generated by mermaid, which we need because we moved architecture.md into the documentation site from basically a hidden space in the hevm repo.

blishko (commented):
Why do we now need the `mermaid` thing? It was not working before?
And do we really need to commit these javascript files into the repository?

doc/src/equivalence.md
```diff
@@ -105,22 +129,21 @@ succeeds.
...
   uint private immutable NUMBER;
   constructor(uint a) {
     NUMBER = 2;
   }
   function stuff(uint b) public returns (uint256) {
-    unchecked{return 2+NUMBER+b;}
+    unchecked {return 2+NUMBER+b;}
   }
 }
 ```
```
blishko, line 144: Isn't there a way to avoid duplicating the code in the file (`create1.sol`) and inlined in the Markdown?
Do we need the separate file?

msooseth: Not sure. I decided to also inline it into Markdown, so it's easy to see and in one place and it's an easy flow for the user.

blishko: I fully agree with that. So, can we drop the solidity files?

msooseth:
Yes!! I would love to drop them! Okay I will do that today :)

doc/src/equivalence.md
msooseth: YES. Let me delete them all :) They are I think useless.

blishko (approved):
LGTM!

PR #707, opened 2025-04-14
Optimize memory representation and operations

elopez:
## Description
An Echidna user reported continuously increasing memory while running mstore/mcopy operations. This PR attempts to resolve that by

* avoiding freezing the full memory on mcopy
* avoiding multiple, repeated small memory expansions, as each expansion has to copy the whole memory contents
  - this is disabled by default, to maintain the concrete memory small (as it might be symbolized at some later point), but a config value on the VM is available by anyone doing pure concrete execution (e.g. Echidna). Limited empirical testing showed 64k is a good value to use.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM.hs
```diff
@@ -672,13 +672,8 @@ exec1 conf = do
             _ -> underrun
             where
             mcopy sz srcOff dstOff = do
-                  m <- gets (.state.memory)
-                  case m of
-                    ConcreteMemory mem -> do
-                      buf <- freezeMemory mem
-                      copyBytesToMemory buf sz srcOff dstOff
-                    SymbolicMemory mem -> do
-                      assign (#state % #memory) (SymbolicMemory $ copySlice srcOff dstOff sz mem mem)
```
elopez, line 681: I think this previous code results in a bit simpler representation for the symbolic case, so we might want to keep it and only modify the concrete case

msooseth:
Thanks! I am looking at it. Only changing the concrete one is definitely OK. I'm trying to see what's the deal with the Symbolic. Feels like it'd be nice to unify the two, like you did.

msooseth:
OK, I had a look. Either way
```
assign (#state % #memory) (SymbolicMemory $ copySlice srcOff dstOff sz mem mem)
```

Will get executed. But with your system, we first execute `buf <- readMemory srcOff sz`, which means:

```
copySlice offset' (Lit 0) size' mem mempty
```

which is basically another copyslice. So I'd rather keep the concrete and the symbolic separate.

Otherwise, it looks cool!

src/EVM.hs
```diff
@@ -2953,7 +2954,12 @@ writeMemory memory offset buf = do
   expandMemory targetSize = do
     let toAlloc = targetSize - VUnboxed.Mutable.length memory
     if toAlloc > 0 then do
-      memory' <- VUnboxed.Mutable.grow memory toAlloc
+      vm <- get
+      -- If you are using pure concrete mode, use a large chunk (e.g. 64k).
+      -- We want to always grow at least a chunk, to avoid the performance impact
+      -- that would happen with repeated small expansion operations, as grow does
+      -- a larger *copy* of the vector on a new place
+      memory' <- VUnboxed.Mutable.grow memory $ max toAlloc vm.config.minMemoryChunk
```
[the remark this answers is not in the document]
msooseth: Nice find! So I had lots of issues in C++ where I was growing by a fixed chunk... (side-note: never use `resize`, use `insert` if it will grow later) not sure how relevant that is here, but it was a painful experience :D My suggestion would be to grow some factor, usually 2x is a good rule of thumb, but 1.5x can also work. But perhaps also make sure we grow at least some value (so we don't allocate by small chunks at the beginning). I'd rather avoid repeated allocations. What do you think?

src/EVM/Exec.hs
```diff
@@ -44,6 +44,7 @@ vmForEthrunCreation creationCode =
     , allowFFI = False
     , freshAddresses = 0
     , beaconRoot = 0
+    , minMemoryChunk = 1
```
msooseth, line 47: Do we need this controllable? We can just set it to a sane default instead? I'd rather not have config options for stuff that a good default works for. It's useful to have this during figuring out what a good default is, but later, it may become a distraction perhaps? What do you think?

elopez: My concern that led me to make this a tunable was symbolic execution. I'm not particularly familiar with how it is represented though so take this with a grain of salt, but what would happen if you end up having a large concrete memory due to the "optimistic growth", which then turns symbolic due to some other reason? How would that get encoded into Z3 et al? Would it hurt performance?

If my concern is unfounded I'm all for picking a good default and not having to have this option 😄

msooseth:
Hey! If you could remove that tunable then I think we could just merge :) Would be nice!

src/EVM.hs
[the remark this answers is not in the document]
msooseth: Thanks, this is nice now, with growth factor being an explicitly named parameter.

msooseth:
Thanks, this is great! Can you please add a CHANGELOG element? Then I think we can merge :)

PR #712, opened 2025-04-22
Limit shrinking to at most 1GB, after that, abandon

msooseth:
## Description
As discussed, there is no point in allowing more than 1GB.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
> Looks good! Is there a way to unify various limits we use for concrete buffers? For example, there is `maxBytes` in `Expr.hs`, which is used in `writeByte` and `copySlice`. And we have some arbitrary limit `120_000_000` in `collapse`.

Yeah, good point. Maybe we should track these in some issue/doc....

PR #713, opened 2025-04-22
Rewrite of equivalence checking, now also printing reasons

msooseth:
## Description
A rewrite of the equivalence checking system. We now also print a String that explains what the actual difference is. The rewrite allows us to explain WHAT is different. It also is MUCH cleaner and actually easier to follow, I think. I finally understand what is going on...

I didn't add any new tests, because the old tests are still valid and working.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Format.hs
+++ b/src/EVM/Format.hs
@@ -6,6 +6,8 @@ module EVM.Format
   , formatPartial
   , formatPartialShort
   , formatProp
+  , formatState
+  , formatError
   , contractNamePart
   , contractPathPart
   , showError
@@ -495,6 +497,13 @@ formatPartialShort = \case
 formatSomeExpr :: SomeExpr -> Text
 formatSomeExpr (SomeExpr e) = formatExpr $ Expr.simplify e
 
+formatState :: Map.Map (Expr EAddr) (Expr EContract) -> Text
+formatState store = indent 2 $ T.unlines (fmap (\(k,v) ->
+              T.unlines
+                [ formatExpr k <> ":"
+                , indent 2 $ formatExpr v
+                ]) (Map.toList store))
+
 formatExpr :: Expr a -> Text
 formatExpr = go
   where
@@ -516,17 +525,10 @@ formatExpr = go
       Success asserts _ buf store -> T.unlines
         [ "(Success"
         , indent 2 $ T.unlines
-          [ "Data:"
-          , indent 2 $ formatExpr buf
+          [ "Data:" , indent 2 $ formatExpr buf
           , ""
-          , "State:"
-          , indent 2 $ T.unlines (fmap (\(k,v) ->
-              T.unlines
-                [ formatExpr k <> ":"
-                , indent 2 $ formatExpr v
-                ]) (Map.toList store))
-          , "Assertions:"
-          , indent 2 . T.unlines $ fmap formatProp asserts
+          , "State:", formatState store
+          , "Assertions:", indent 2 . T.unlines $ fmap formatProp asserts
           ]
         , ")"
         ]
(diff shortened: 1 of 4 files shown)


src/EVM/Format.hs
```diff
@@ -495,6 +497,13 @@ formatPartialShort = \case
 formatSomeExpr :: SomeExpr -> Text
 formatSomeExpr (SomeExpr e) = formatExpr $ Expr.simplify e
 
+formatState :: Map.Map (Expr EAddr) (Expr EContract) -> Text
+formatState store = indent 2 $ T.unlines (fmap (\(k,v) ->
+              T.unlines
+                [ formatExpr k <> ":"
+                , indent 2 $ formatExpr v
+                ]) (Map.toList store))
+
```
msooseth, line 506: I reuse this `formatState` later in SymExec for equivalence checking. Allows me to print the state when it differs between two contracts.

src/EVM/SymExec.hs
```diff
@@ -14,15 +14,16 @@ import Control.Monad.State.Strict (runStateT)
 import Data.Bifunctor (second, first)
 import Data.ByteString (ByteString)
 import Data.ByteString qualified as BS
+import Data.Maybe (mapMaybe)
```
blishko, line 17: There is already `import Data.Maybe` a few lines below. I think this belongs there, no?

src/EVM/SymExec.hs
```diff
@@ -834,19 +853,19 @@ equivalenceCheck' solvers branchesA branchesB create = do
     -- the solver if we can determine unsatisfiability from the cache already
     -- the last element of the returned tuple indicates whether the cache was
     -- used or not
-    check :: App m => UnsatCache -> Set Prop -> m EquivResult
-    check knownUnsat props = do
+    check :: App m => UnsatCache -> (Set Prop, String) -> m (EquivResult, String)
+    check knownUnsat (props, meaning) = do
```
blishko, line 857: I am a bit confused here, why you need to pass the `meaning` parameter here if you don't do anything with it, you are just returning it with the original result.

src/EVM/SymExec.hs
msooseth: Good point, sorry. Fixed.

src/EVM/SymExec.hs
msooseth: As always, you picked a great issue. I have fixed this now. This fix should make `checkAll` nicer too, I think. Thanks for this, it really was silly. Let me know what you think!

src/EVM/SymExec.hs
```diff
@@ -837,84 +855,111 @@ equivalenceCheck' solvers branchesA branchesB create = do
...
+           res <- wrap (env $ check cache props)
+           pure (res, meaning)
 
     -- Takes two branches and returns a set of props that will need to be
     -- satisfied for the two branches to violate the equivalence check. i.e.
     -- for a given pair of branches, equivalence is violated if there exists an
     -- input that satisfies the branch conditions from both sides and produces
     -- a differing result in each branch
-    distinct :: App m => Expr End -> Expr End -> m (Maybe (Set Prop), [EquivResult], [Expr End])
+    distinct :: App m => Expr End -> Expr End -> m ([(Set Prop, String)], EqIssues)
```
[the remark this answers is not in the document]
msooseth: Oh yes. I am sorry. This went through a few refactors and at one point I used this list for something. But with the new refactor, it doesn't need it. Thanks, I fixed it. It's a lot cleaner now.

Sorry for all my mistakes :S

PR #714, opened 2025-04-23
No need for solc to run hevm

msooseth:
There is actually no need for solc to run hevm. Fixing the README.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
LGMT!

PR #716, opened 2025-04-24
More rewrite rules for PEq, PNeg, missing eqByte call, and distributivity for And

msooseth:
More rewrite rules for PEq, PNeg, missing eqByte call, and distributivit for And

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -203,11 +203,12 @@ sar = op2 SAR (\x y ->
 
 peq :: (Typeable a) => Expr a -> Expr a -> Prop
 peq (Lit x) (Lit y) = PBool (x == y)
-peq a@(Lit _) b = PEq a b
-peq a b@(Lit _) = PEq b a -- we always put concrete values on LHS
+peq (LitByte x) (LitByte y) = PBool (x == y)
+peq (ConcreteBuf x) (ConcreteBuf y) = PBool (x == y)
 peq a b
   | a == b = PBool True
-  | otherwise = PEq a b
+  | otherwise = let args = sort [a, b]
+     in PEq (args !! 0) (args !! 1)
 
 -- ** Bufs ** --------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -203,11 +203,12 @@ sar = op2 SAR (\x y ->
...
 peq :: (Typeable a) => Expr a -> Expr a -> Prop
 peq (Lit x) (Lit y) = PBool (x == y)
-peq a@(Lit _) b = PEq a b
-peq a b@(Lit _) = PEq b a -- we always put concrete values on LHS
+peq (LitByte x) (LitByte y) = PBool (x == y)
+peq (ConcreteBuf x) (ConcreteBuf y) = PBool (x == y)
 peq a b
   | a == b = PBool True
-  | otherwise = PEq a b
+  | otherwise = let args = sort [a, b]
```
msooseth, line 210: Sort prefers Lit above all else, so it will put Lit as 1st. But it will also order everything else, which is what we want.

src/EVM/Expr.hs
```diff
@@ -1182,6 +1188,7 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | otherwise = shr a v
 
     -- Bitwise AND & OR. These MUST preserve bitwise equivalence
+    go (And (Lit a) (Lit b)) = Lit (a .&. b)
```
blishko, line 1191: This should not be necessary, no? This simplification should happen in `EVM.Expr.and a b`.

src/EVM/Expr.hs
```diff
@@ -1307,9 +1314,11 @@ simplifyProp prop =
     go (PNeg (PGEq a b)) = PLT a b
     go (PNeg (PLT a b)) = PGEq a b
     go (PNeg (PLEq a b)) = PGT a b
+    go (PNeg (PAnd a b)) = POr (PNeg a) (PNeg b)
```
blishko, line 1317: If we are adding this, shouldn't we add the dual `go (PNeg (POr a b)) = PAnd (PNeg a) (PNeg b)` as well?

blishko (commented):
I think it would be good to add tests for simplifications that would fail before, but pass with the new rules added.

It would show if the rules are really needed.
As I point out, one of the addition should be redundant and not necessary.

src/EVM/Expr.hs
msooseth: Oops, good point. OK, I'll write test cases :)

src/EVM/Expr.hs
msooseth: Yes! Thanks, yeah, I just added this, along with a test case :)

msooseth:
Thanks! I have now added 8 test cases to test each of the new rules. They all pass. And in main they all fail. Thanks, this was actually quite instructive, and it's now guaranteed not to be broken!

blishko (approved):
Nice!

PR #717, opened 2025-04-24
More robust printing

msooseth:
## Description
We no longer `internalError` in case the buffer is too large and cannot be printed. Instead, we handle it as gracefully as possible.

I have now added a test case that exercises this new capability.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
LGTM!

PR #718, opened 2025-04-28
Add test case for robust printing

msooseth:
## Description
Sorry, forgot to add this test case to the robust printing PR. It simply checks that we don't `error` out due to a large buffer.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (commented):
Looks good!
Do you think this test would fit into `calling-solvers` test group, instead of starting a new group?
The query could be build on `Props` level. Or do you see benefit in having the Solidity example?

msooseth:
Thanks! Sorry. Now added a straighforward test, too, which is just a pure `Expr End` that gets printed. I left in the Solidity test, I think it may still be useful.

PR #720, opened 2025-04-29
[DRAFT] Fix unsat cache

msooseth:
## Description
Wow, UNSAT cache was not actually updated when checking equivalence.

TODO:
* Check if we could have an unsat cache also for regular solving? Do we have it already?
* Add a test-case that somehow checks for this working.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -5,7 +5,7 @@ module EVM.SymExec where
 
 import Control.Concurrent.Async (concurrently, mapConcurrently)
 import Control.Concurrent.Spawn (parMapIO, pool)
-import Control.Concurrent.STM (TVar, readTVarIO, newTVarIO)
+import Control.Concurrent.STM (TVar, readTVarIO, newTVarIO, writeTVar, atomically, readTVar)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -79,6 +79,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Pretty printing of buffers is now more robust. Instead of throwing an `internal error`,
   we now try best to print everything we can, and print an appropriate error message
   instead of crashing.
+- During 0xd07e92cd9 we accidentally removed adding already proven Prop sets to
```
blishko, line 82: No need for changelog entry, the cache was working correctly at the last release.

msooseth: Deleted it :) Sorry.

msooseth:
This works locally and is wrong when ran on the CI. My first guess would be some kind of race condition :S

msooseth:
This is outdated, fixed by #747

PR #721, opened 2025-04-29
Cleanup code, use a struct for loop detection configuration

msooseth:
This likely seems minor but I find it so much easier to order these in the right way and not leave empty lines when not necessary. Also, passing around a struct is so much more safe than passing around `Integer`-s, because the order of the integers can be messed up. I find the code easier to read this way.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
LGTM!

PR #723, opened 2025-04-29
Replace actions from DeterminateSystems

blishko:
## Description

This aims to replace the GitHub actions from DeterminateSystems since their Magic Nix Cache action is discontinued.
I believe we can use `cache-nix-action` which simply wraps the `cache` GitHub action and does not store the cache anywhere outside of GitHub.
The proposed action requires a bit of manual configuration but not too much.

Moreover, I believe we only need the cache for CI build (`ci` package of `hevm`,  our `build.yaml` workflow).
The other workflows, `Check Dependencies` and `Release` do not need to use a cache at all.
Specifically, for releases, I think it is better not to use a cache and build the environment from scratch.

There is one more change prosed here: The static build for Linux takes a long time, because it is building the whole GHC from scratch. I believe we do not need to do this on CI, where we want to have relatively quick feedback.
The release workflow will still build static binary (this is the `redistributable` package, as opposed to the `ci` package in `flake.nix`.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,8 +30,18 @@ jobs:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
-      - uses: DeterminateSystems/nix-installer-action@main
-      - uses: DeterminateSystems/magic-nix-cache-action@main
+      - name: Install Nix
+        uses: nixbuild/nix-quick-install-action@v30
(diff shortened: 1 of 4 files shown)


msooseth:
Thank you!!!

PR #729, opened 2025-05-02
Optimize `maybeLit{Byte,Word,Addr}Simp` and `maybeConcStoreSimp`

elopez:
## Description

`concKeccakSimpExpr` can be a bit expensive, and these functions are called frequently (such as on `litSelf`) so provide a fast path for those values that are already concrete.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (commented):
This looks good!
I would merge it, but would like to get confirmation from @msooseth.

PR #735, opened 2025-05-07
Avoid list concatenation when collecting all paths in flattenExpr

blishko:
## Description
Instead of concatenating branches from true and false cases of ITE, we use the accumulator to accumulate the branches.
Since we are prepending, we traverse the false case first, so that in the accumulator, the paths from true case will end up before the branches from the false case.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM/SymExec.hs
```diff
@@ -573,14 +573,17 @@ toEContract c = C c.code c.storage c.tStorage c.balance c.nonce
...
-    go pcs = \case
-      ITE c t f -> go (PNeg ((PEq (Lit 0) c)) : pcs) t <> go (PEq (Lit 0) c : pcs) f
-      Success ps trace msg store -> [Success (nubOrd $ ps <> pcs) trace msg store]
-      Failure ps trace e -> [Failure (nubOrd $ ps <> pcs) trace e]
-      Partial ps trace p -> [Partial (nubOrd $ ps <> pcs) trace p]
+    go :: [Prop] -> [Expr End] -> Expr End -> [Expr End]
+    go pcs acc  = \case
+      ITE c t f ->
+        let
+          accf = go (PEq (Lit 0) c : pcs) acc f
```
msooseth, line 582:
```suggestion
        let accf = go (PEq (Lit 0) c : pcs) acc f
```

msooseth (commented):
Wow this is a bit of a trickery :D I still need to stare at this for longer. Can you please add a changelog entry in the meanwhile?

blishko:
Yeah, it's less readable, so I would like to see a confirmation that it helps with some bottleneck, otherwise I would not merge it.  I'll just convert it to a draft for now to indicate that.

msooseth:
I am closing, because this will be not relevant once we have a queue for sybmolic exec

PR #736, opened 2025-05-09
Increase exploration depth when exploring both side of branches with symbolic conditions

gustavo-grieco:
Please consider limiting the exploration when there are loops with symbolic conditions

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM/SymExec.hs
```diff
@@ -409,7 +409,7 @@ interpret fetcher iterConf vm =
                       -- if we can statically determine unsatisfiability then we skip exploring the jump
                       [PBool False] -> liftIO $ stToIO $ runStateT (continue (Case False)) vm
```
msooseth, line 410:
```suggestion
                      -- if we can statically determine unsatisfiability then we skip exploring the jump
                      [PBool False] -> liftIO $ stToIO $ runStateT (continue (Case False)) vm
                      [] -> liftIO $ stToIO $ runStateT (continue (Case True)) vm
```

We might as well add the constant True as well :)

msooseth (changes requested):
Oh wow, nice catch!! Can you please (1) consider my improvement, and can you please (2) add a changelog entry?

gustavo-grieco:
I´m not sure why the tests are failing 🤔

msooseth:
I have now created a PR for this PR: https://github.com/gustavo-grieco/hevm/pull/3

Once that's merged, we can review this and merge :)

src/EVM/SymExec.hs
```diff
@@ -405,11 +402,12 @@ interpret fetcher iterConf vm =
                     -- ask the smt solver about the loop condition
                     performQuery
                   _ -> do
-                    (r, vm') <- case simpProps of
+                    (r, vm') <- case Expr.concKeccakProps ((cond ./= Lit 0):preconds) of
```
msooseth, line 405:
```suggestion
                    (r, vm') <- case Expr.concKeccakSimpExpr ((cond ./= Lit 0):preconds) of
```

Sorry, I'm an idiot. This is what's needed...

src/EVM/SymExec.hs
msooseth, line 1: Wow this is such a mess! Sorry, I made this mess. Let's undo this change. I need to get into this issue and fix it properly. And it's a different issue altogether anyway. I'm sorry. Please revert this file change, and squash, and then we can merge. I'm sorry, this is a mess I created :(

msooseth (changes requested):
Sorry, that mess I made needs to be removed. Now that I had a deeper look, there are some more serious issues here that I'll fix via a PR.

src/EVM/SymExec.hs
msooseth: This is now a separate PR here: https://github.com/ethereum/hevm/pull/749

gustavo-grieco:
Since I needed to undo some merge commits (and probably doing force push, to avoid polluting the history), I decided it was easier to create a new PR

PR #738, opened 2025-05-10
Fix Windows CI build

elopez:
## Description
Recent MSYS2 updates broke the build again. It turns out libff was being built by referencing the MSYS2-installed headers instead of the GHC-bundled ones. This adjusts the build process to have libff built with the same headers as the rest of hevm.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (commented):
Looks OK, but isn't there a way to use relative path instead of this absolute path `D:/a/_temp/msys64/clang64/include`? Can we at least extract it to a variable instead of repeating it 4 times?

blishko (approved):
OK!

blishko:
Looks like extracting to a variable actually broke the build again?

elopez:
yeah, I defined the variable too late in the process 😅  I went ahead and updated it to do it earlier and also reference the path from the action, hopefully it works this time 🤞

blishko (approved):
Nice!
This is much better, no hardcoded path :)

PR #741, opened 2025-05-12
Avoid internal error when attempting to shrink buffers in model

### LINKED ISSUE #739: In some cases, the solver can return unknown in shrinkBuf
```
echidna: Internal Error: Unexpected solver output: unknown -- CallStack (from HasCallStack):
  internalError, called at src/EVM/Solvers.hs:312:14 in hevm-0.54.2-5ddS7SL5q6i9R3j2SKWke8:EVM.Solvers
  error, called at src/EVM/Types.hs:1521:19 in hevm-0.54.2-5ddS7SL5q6i9R3j2SKWke8:EVM.Types
  internalError, called at src/EVM/Solvers.hs:312:14 in hevm-0.54.2-5ddS7SL5q6i9R3j2SKWke8:EVM.Solvers
```

Unclear what we should do to fix this.

blishko:
## Description
Instead of exiting with internal error, we simply give up on shrinking the current buffer and continue with the original model.

Fixes #739.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Thanks! Perfect fix :)

PR #743, opened 2025-05-14
Move unsat cache to `Solvers.hs`

msooseth:
## Description
This moves the UNSAT cache to a central location, in `Solvers.hs`. This means that UNSAT cache is filled and used even during interpretation. Also, one can simply use `checkSatWithProps` and the UNSAT cache will be there and working.

This also means that the unsat cache is now available to every place that calls `checkSatWithProps`, which also includes `PleaseAskSMT`, i.e. it will now work also when deciding on branches.

We now have a queue for the cache, and it is taken as priority before picking up any new task.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Solvers.hs
+++ b/src/EVM/Solvers.hs
@@ -10,12 +10,14 @@ import GHC.Natural
 import GHC.IO.Handle (Handle, hFlush, hSetBuffering, BufferMode(..))
 import Control.Concurrent.Chan (Chan, newChan, writeChan, readChan)
 import Control.Concurrent (forkIO, killThread)
+import Control.Concurrent.STM (writeTChan, newTChan, TChan, tryReadTChan, atomically)
 import Control.Monad
 import Control.Monad.State.Strict
 import Control.Monad.IO.Unlift
 import Data.Char (isSpace)
 import Data.Map (Map)
 import Data.Map qualified as Map
+import Data.Set (Set, isSubsetOf, fromList)
 import Data.Maybe (fromMaybe, isJust, fromJust)
 import Data.Either (isLeft)
 import Data.Text qualified as TStrict
@@ -69,6 +71,9 @@ data MultiSol = MultiSol
 -- | A script to be executed, a list of models to be extracted in the case of a sat result, and a channel where the result should be written
 data Task = TaskSingle SingleData | TaskMulti MultiData
 
+newtype CacheEntry = CacheEntry [Prop]
+  deriving (Show, Eq)
+
 data MultiData = MultiData
   { smt2 :: SMT2
   , multiSol :: MultiSol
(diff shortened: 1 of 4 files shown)


gustavo-grieco:
Can `checkSatWithProps` be used to in `verify` so we can also benefit from caching?

msooseth:
@gustavo-grieco Basically, this PR is meant to replace all that cache in `verify` and elsewhere, so we have a single, unified cache.

blishko (commented):
Looks good, though I am wondering: How is this going to work? Wouldn't all SMT queries  be put in the queue before the requests to add results to the cache?

Have you checked that this is working as expected?

msooseth:
This is now ready for re-review. @blishko maybe you can take a peek?

src/EVM/SymExec.hs
```diff
@@ -674,34 +673,33 @@ verify solvers opts preState maybepost = do
...
+  do
+    when conf.debug $ liftIO $ putStrLn "   Simplifying expression"
     let expr = if opts.simp then (Expr.simplify exprInter) else exprInter
-    when conf.dumpExprs $ T.writeFile "simplified.expr" (formatExpr expr)
-    when conf.dumpExprs $ T.writeFile "simplified-conc.expr" (formatExpr $ Expr.simplify $ mapExpr Expr.concKeccakOnePass expr)
+    when conf.dumpExprs $ liftIO $ T.writeFile "simplified.expr" (formatExpr expr)
+    when conf.dumpExprs $ liftIO $ T.writeFile "simplified-conc.expr" (formatExpr $ Expr.simplify $ mapExpr Expr.concKeccakOnePass expr)
     let flattened = flattenExpr expr
-    when conf.debug $ do
+    when conf.debug $ liftIO $ do
```
blishko, line 682: This seems unrelated to the main point of the PR.
To me it seems better to write `liftIO` once. What is the problem with that?

blishko (commented):
This looks good!

I would still like to play with it a bit before we merge it.

src/EVM/Solvers.hs
```diff
@@ -69,6 +71,9 @@ data MultiSol = MultiSol
 -- | A script to be executed, a list of models to be extracted in the case of a sat result, and a channel where the result should be written
 data Task = TaskSingle SingleData | TaskMulti MultiData
 
+newtype UnsatCache = UnsatCache [Prop]
```
blishko, line 74: I think this should be renamed, no? `[Prop]` is not the unsat cache. It is more like an `Entry` that should be added to unsat cache.

blishko (commented):
Can you explain why you need `STM` (`TChan`) for the cache? Why the basic `Chan` is not sufficient?

src/EVM/Solvers.hs
msooseth: Good point! Renamed it `CacheEntry`, I hope that works :)

blishko:
Let's add a (CLI) test to check if we are actually using the cache (grep for debug output about founding an entry in the cache)!

msooseth:
OK, added a test, let's hope it passes :)

blishko (approved):
Nice!

PR #744, opened 2025-05-14
Add benchmarking with Solidity examples

elopez:
## Description

This is a refresh of #355 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


bench/bench-perf.hs
```diff
@@ -0,0 +1,410 @@
...
+import GHC.Word
+import System.Directory (setCurrentDirectory)
+import System.FilePath.Find qualified as Find
+import System.IO.Temp
+import System.Process (readProcessWithExitCode)
+import Test.Tasty.Bench
+import Control.Monad.ST
+import EVM.FeeSchedule (feeSchedule)
+
+-- benchmark hevm and EVM tool on the folder called benchmarks
```
msooseth, line 36: I... don't understand. Where is this folder? It's not part of the PR I think? Am I supposed to have it? Sorry I must be missing something about tasty-bench :S

elopez: I saw this implements some functions to run things like geth and evm but I'm not sure how they're supposed to be used. Maybe it's referring to that, or maybe it's dead code 🤔

bench/bench-perf.hs
```diff
@@ -0,0 +1,410 @@
...
+      allowFFI = False,
+      otherContracts = [],
+      freshAddresses = 0,
+      beaconRoot = 0
+    }
+
+benchCode :: ByteString -> Benchmark
+benchCode bs = bench "interpret" $ nfIO $ runApp $ do
+  vm <- vmFromByteString bs
+  isRight <$> Stepper.interpret (Fetch.zero 0 Nothing) vm Stepper.execFully
```
msooseth, line 89: Never called?

bench/bench-perf.hs
```diff
@@ -0,0 +1,410 @@
...
+
+benchFile :: FilePath -> Benchmark
+benchFile path = bench (show path) $ nfIO $ runApp $ do
+  bs <- liftIO $ strip0x <$> ByteString.readFile path
+  vm <- vmFromByteString bs
+  isRight <$> Stepper.interpret (Fetch.zero 0 Nothing) vm Stepper.execFully
+
+-- bench every .bin file in a given folder
+benchFolder :: FilePath -> IO [Benchmark]
+benchFolder path = map benchFile <$> Find.find Find.always (Find.extension Find.==? ".bin") path
```
msooseth, line 99: Never called either?

bench/bench-perf.hs
```diff
@@ -0,0 +1,410 @@
...
+  vm <- liftIO $ vmFromRawByteString bs
+  Stepper.interpret (Fetch.zero 0 Nothing) vm (Stepper.evm (abiCall (vmOptsToTestVMParams (vm0Opts (initialContract (RuntimeCode (ConcreteRuntimeCode bs))))) (Left ("main()", emptyAbi))) >> Stepper.execFully)
+
+benchMain :: (String, ByteString) -> Benchmark
+benchMain (name, bs) = bench name $ nfIO $ runApp $ (\x -> if isRight x then () else error "failed") <$> callMainForBytecode bs
+
+benchBytecodes :: [ByteString] -> [Benchmark]
+benchBytecodes = map (\x -> bench "bytecode" $ nfIO $ runApp (do
+                                                       vm <- liftIO $ vmFromRawByteString x
+                                                       isRight <$> (Stepper.interpret (Fetch.zero 0 Nothing) vm Stepper.execFully)))
```
msooseth, line 140: Also never called?

msooseth (changes requested):
Can you remove these since they are not called?

bench/bench-perf.hs
msooseth: I suspect some dead code related to what's below. Let's remove it all -- they don't seem to serve a purpose for the moment I think?

bench/bench-perf.hs
elopez: Removed

bench/bench-perf.hs
elopez: Removed

bench/bench-perf.hs
elopez: Removed

msooseth (approved):
Thanks, looks good! @blishko what do you think?

bench/bench-perf.hs
```diff
@@ -0,0 +1,304 @@
...
+      chainId = 1,
+      create = False,
+      txAccessList = mempty, -- TODO: support me soon
+      allowFFI = False,
+      otherContracts = [],
+      freshAddresses = 0,
+      beaconRoot = 0
+    }
+
+-- why do we need this?
```
blishko, line 71: Do we have an answer to this question?

doc/src/for-hevm-developers.md
```diff
@@ -55,6 +55,28 @@ Similarly, the `maxSuccess` value can be modified for a single test. The followi
 testProperty <property_name> $ withMaxSuccess 20 $ ...
 ```
 
+## Running benchmarks
+
+You can also measure and compare the performance across hevm versions using the benchmarks.
+`bench-perf` focuses on concrete execution performance, and `bench` is aimed at
```
blishko, line 61: This new code is about `bench-perf`, right? It's about concrete execution.
Do we have documentation on `bench`?

bench/bench-perf.hs
blishko: Well, we need to figure out if we need it :)

If yes, then drop the comment. If no, then drop the whole function :)

blishko: I guess we need it, so you can just drop the comment :)

PR #745, opened 2025-05-14
Allow changing prefix from prove via `--prefix` in `test` mode

msooseth:
## Description
Currently, default prefix is "prove" as per documentation. Actually, we also accept "check". This PR allows us to change this from "prove" to any other value. We also drop "check" as a prefix that's accepted. The "check" was never advertised, so I think it's valid to remove? It makes the code simpler, and I think it's OK this way, as `--prefix check` can always be used to be backwards-compatible. This change is highlighted in the changelog.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Dapp.hs
+++ b/src/EVM/Dapp.hs
@@ -27,7 +27,6 @@ data DappInfo = DappInfo
   , solcByHash :: Map W256 (CodeType, SolcContract)
   , solcByCode :: [(Code, SolcContract)] -- for contracts with `immutable` vars.
   , sources    :: SourceCache
-  , unitTests  :: [(Text, [Sig])]
   , abiMap     :: Map FunctionSelector Method
   , eventMap   :: Map W256 Event
   , errorMap   :: Map W256 SolError
@@ -57,7 +56,6 @@ dappInfo root (BuildOutput (Contracts cs) sources) =
 
   in DappInfo
     { root = root
-    , unitTests = findAllUnitTests solcs
     , sources = sources
     , solcByName = cs
     , solcByHash =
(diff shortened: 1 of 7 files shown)


src/EVM/Dapp.hs
```diff
@@ -27,7 +27,6 @@ data DappInfo = DappInfo
   , solcByHash :: Map W256 (CodeType, SolcContract)
   , solcByCode :: [(Code, SolcContract)] -- for contracts with `immutable` vars.
   , sources    :: SourceCache
-  , unitTests  :: [(Text, [Sig])]
```
msooseth, line 30: This was actually unused

src/EVM/Dapp.hs
```diff
@@ -57,7 +56,6 @@ dappInfo root (BuildOutput (Contracts cs) sources) =
 
   in DappInfo
     { root = root
-    , unitTests = findAllUnitTests solcs
```
msooseth, line 60: This was actually unused, and so `findAllUnitTests` was also unused.

src/EVM/Dapp.hs
```diff
@@ -92,37 +90,33 @@ emptyDapp = dappInfo "" mempty
 unitTestMarkerAbi :: FunctionSelector
 unitTestMarkerAbi = abiKeccak (encodeUtf8 "IS_TEST()")
 
-findAllUnitTests :: [SolcContract] -> [(Text, [Sig])]
-findAllUnitTests = findUnitTests ".*:.*\\.(check|prove).*"
-
```
msooseth, line 97: This function was unused.

blishko:
I don't think we need to pass the `prefix` through all the methods.
For example, `unitTestMethodsFiltered` already takes a filter, `Text -> Bool`. I think we should just include the check for the prefix in this filter.
`mkSig` does not need to do any filtering, it can return `Sig` instead of `MaybeSig`.

Though I am not sure if I understand correctly the `match` option. From the code, I inferred that it is a regex, but does it match against contract name or again method names?

msooseth:
> I don't think we need to pass the `prefix` through all the methods. For example, `unitTestMethodsFiltered` already takes a filter, `Text -> Bool`. I think we should just include the check for the prefix in this filter. `mkSig` does not need to do any filtering, it can return `Sig` instead of `MaybeSig`.
> 
> Though I am not sure if I understand correctly the `match` option. From the code, I inferred that it is a regex, but does it match against contract name or again method names?

It allows to match against both contract and method names via `contractname.methodname`:
```
  let testName (Sig n _) = c.contractName <> "." <> n
```
Which `matcher` is ran against. However, `prefix` is matched only against the method name. What do you think we should do?

blishko (approved):
I was hoping there is nice way to unify matching against contract and method name and against method prefix.

But I don't see one right now. So if you also don't see a nice way to simplify this, then let's merge it and move on!

PR #746, opened 2025-05-16
Fix implementation of Ord for Prop

blishko:
## Description

Previously, the implementation of `<=` for Prop would not obey the antisymmetry law, I believe in the case of `PEq` specifically. This leads to weird behaviour, namely that duplicates are not properly recognized in Set.

The proposed change is to implement `compare` directly, which I think is easier than trying to fix `<=`.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   we now try best to print everything we can, and print an appropriate error message
   instead of crashing.
 - We no longer produce duplicate SMT assertions regarding concrete keccak values.
+- Ord is now correctly implemented for Prop.
 
 ## Changed
 - Warnings now lead printing FAIL. This way, users don't accidentally think that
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   we now try best to print everything we can, and print an appropriate error message
   instead of crashing.
 - We no longer produce duplicate SMT assertions regarding concrete keccak values.
+- Ord is now correctly implemented for Prop.
```
msooseth, line 83: Yes. Sorry :)

msooseth (approved):
Yes. Well done. Thanks, I learned something!

CHANGELOG.md
blishko: This might be optimistic from me :)
Time will tell...

PR #747, opened 2025-05-16
Fix unsat cache in equivalence checking

blishko:
Recently, in a refactoring, the write to the cache disappeared, so the cache was actually always empty.
Here we bring back writing to the cache after unsat result.

Moreover, we make additional changes related to the cache: Given a set of proposition, we check if it is a *superset* of any known unsatisfiable set of propositions. The previous check for subset relation was not correct: a subset of an unsatiafiable set can be satisfiable.
Related to that, we order the queries from the smallest to largest, so that we increase the likelyhood of cache hits.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -5,7 +5,7 @@ module EVM.SymExec where
 
 import Control.Concurrent.Async (concurrently, mapConcurrently)
 import Control.Concurrent.Spawn (parMapIO, pool)
-import Control.Concurrent.STM (TVar, readTVarIO, newTVarIO)
+import Control.Concurrent.STM (TVar, readTVarIO, newTVarIO, atomically, modifyTVar')
 import Control.Monad (when, forM_, forM)
 import Control.Monad.IO.Unlift
 import Control.Monad.Operational qualified as Operational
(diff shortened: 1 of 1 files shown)


msooseth:
Oh wow, yes. Definitely. It's a superset. And the ordering via `compare` is much better. These are actually the right things.... I think it's good, although I would like to integrate them into https://github.com/ethereum/hevm/pull/743/ and merge that way.


I double-checked the superset and indeed it's what we want:

```haskell
ghci> supersetAny a bs = any (`isSubsetOf` a) bs
ghci> supersetAny (fromList [5]) [fromList [5,6]]
True
```

and the new compare is also correct:

```haskell
ghci> sortBySize = sortBy (\(a, _) (b, _) -> compare (Set.size a) (Set.size b))
ghci> sortBySize [(fromList [5,6],"a"), (fromList [4], "a")]
[(fromList [4],"a"),(fromList [5,6],"a")]
```

We should do the small ones first, then the supersets will hit more often.

I am thinking about merging and rebasing my PR on top of this. What do you think?

msooseth (approved):
Yes, thanks!!

PR #749, opened 2025-05-19
More complete simplification of branching during interpretation

msooseth:
## Description
This is a minor change that allows us to do more simplification during interpretation. It also accounts for the case of pure SAT (not only UNSAT) in branching.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
Looks OK!

Issue #751, opened 2025-05-19
Behaviour difference between revert() and revert("message")


shanjiaming:
hevm version
0.54.2 [no git revision present]

It seems that `revert()` behaves like `assume`, while `revert("message")` behaves like ` assert`. This cause a big trouble when I'm using hevm, since I want all revert behaves like `assume`.

Here is a demo to demonstrate the difference:
```
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.17;
import "forge-std/Test.sol";
contract RevertAssertDemoTest is Test {

    function prove_func(uint256 revertControlInput) public {
        if (revertControlInput == 0) {
            revert();
        }
        assert(true);
    }
} 
```

```
rm -rf out/TestRevertAssert.sol && forge build --ast && hevm test --max-iterations 1000000000000000
[⠒] Compiling...
[⠰] Compiling 19 files with 0.8.17
[⠒] Solc 0.8.17 finished in 2.13s
Compiler run successful with warnings:
Warning (2018): Function state mutability can be restricted to pure
 --> src/TestRevertAssert.sol:6:5:
  |
6 |     function prove_func(uint256 revertControlInput) public {
  |     ^ (Relevant source part starts here and spans across multiple lines).

Checking 1 function(s) in contract src/TestRevertAssert.sol:RevertAssertDemoTest
[RUNNING] prove_func(uint256)
   [PASS] prove_func
```

```
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.17;
import "forge-std/Test.sol";
contract RevertAssertDemoTest is Test {

    function prove_func(uint256 revertControlInput) public {
        if (revertControlInput == 0) {
            revert("some error message");
        }
        assert(true);
    }
} 
```

```
rm -rf out/TestRevertAssert.sol && forge build --ast && hevm test --max-iterations 1000000000000000
[⠒] Compiling...
[⠰] Compiling 19 files with 0.8.17
[⠒] Solc 0.8.17 finished in 2.20s
Compiler run successful with warnings:
Warning (2018): Function state mutability can be restricted to pure
 --> src/TestRevertAssert.sol:6:5:
  |
6 |     function prove_func(uint256 revertControlInput) public {
  |     ^ (Relevant source part starts here and spans across multiple lines).

Checking 1 function(s) in contract src/TestRevertAssert.sol:RevertAssertDemoTest
[RUNNING] prove_func(uint256)
   [FAIL] prove_func
   Counterexample:
[...]

PR #753, opened 2025-05-20
Fixing checking assert failure

msooseth:
## Description
As per discussion on the public matrix channel and the issue #751 we have a problem with `revert()`/`revert("message")/`assert` in `test` mode.  Fixes:
- `require(a==b, "str")` is no longer considered a FAIL
- `revert(string)` is no longer considered a FAIL

As before, all `assert`-s are considered a FAIL. Also, as before, `revert()` is not considered a fail.

The `badVault` code seems to PASS now. It used to fail on the `require(stuff, "string")` which should not have been a cause for FAIL. It may actually be a good vault, and so maybe the PASS is correct.

We now also print a yellow WARNING in case it's  FAIL because of Cex or Warnings -- but ALSO everything reverted. So at least we indicate that there's more than just the issue with the FAIL.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
OK, I think this is finally ready for review :) I added a number of test cases, and removed badVault. We can come back to badVault later.

src/EVM/UnitTest.hs
```diff
@@ -210,7 +210,10 @@ symRun opts@UnitTestOptions{..} vm (Sig testName types) = do
             Success _ _ _ store -> if opts.checkFailBit then PNeg (failed store) else PBool True
             Failure _ _ (Revert msg) -> case msg of
               ConcreteBuf b ->
-                if (BS.isPrefixOf (selector "Error(string)") b) || b == panicMsg 0x01 then PBool False
+                -- We need to drop the selector (4B), the offset value (aligned to 32B), and the length of the string (aligned to 32B)
+                -- NOTE: assertTrue/assertFalse does not have the double colon after "assertion failed"
+                let assertFail = selector "Error(string)" `BS.isPrefixOf` b && "assertion failed" `BS.isPrefixOf` (BS.drop (4+32+32) b)
+                in if assertFail || b == panicMsg 0x01 then PBool False
                 else PBool True
               b -> b ./= ConcreteBuf (panicMsg 0x01)
```
d-xo, line 218: we should run the prefix check in the symbolic case too

msooseth:
The commit 34fdaedc18173d4decab7b5d14949c3694a30bd3 now also handles the symbolic case when the prefix is a concrete "assertion failed" of the symbolic string.

We currently don't handle when the string returned doesn't have a concrete prefix matching "assertion failed"

src/EVM/SymExec.hs
```diff
@@ -508,6 +508,7 @@ getExpr solvers c signature' concreteArgs opts = do
     - 0x51: If you call a zero-initialized variable of internal function type.
 
   see: https://docs.soliditylang.org/en/v0.8.6/control-structures.html?highlight=Panic#panic-via-assert-and-error-via-require
+  NOTE: does not deal with e.g. `assertEq()`
 -}
 checkAssertions :: [Word256] -> Postcondition s
```
d-xo, line 510: Is this used anywhere? can we either fix it or delete it?

d-xo: or is there a reason to only check for direct calls to `assert` in solidity?

src/EVM/UnitTest.hs
```diff
@@ -210,9 +212,12 @@ symRun opts@UnitTestOptions{..} vm (Sig testName types) = do
             Success _ _ _ store -> if opts.checkFailBit then PNeg (failed store) else PBool True
             Failure _ _ (Revert msg) -> case msg of
```
d-xo, line 216: For the sake of completeness, we should also handle the invalid opcode (`0xFE`) which was raised by solc < 0.4.

src/EVM/UnitTest.hs
```diff
@@ -257,7 +262,14 @@ symRun opts@UnitTestOptions{..} vm (Sig testName types) = do
     liftIO $ putStr txtResult
     liftIO $ printWarnings ends results t
     pure (not (any isCex results), not (warnings || unexpectedAllRevert))
-
+    where
+      symbolicFail :: Expr Buf -> Prop
+      symbolicFail e =
+        let text = V.fromList $ map (fromIntegral . ord) "assertion failed"
+            panic = e == ConcreteBuf (panicMsg 0x01)
+            assertFail =  V.take (length text) (Expr.concretePrefix e) == text
```
d-xo: I don't think this is sound? What if we have a symbolic prefix that could still be "assertion failed"?

msooseth: Yeah, that would give us a false positive.  But we provide the CEX, so it can be checked. I think it's not a likely problem to come up. But I'm OK with removing it, of course -- I just don't know what to put in its place. Should we just ignore this kind of symbolic fail and simply have

```haskell
      symbolicFail e =
        let panic = e == ConcreteBuf (panicMsg 0x01)
        in PBool (not panic)
```

What do you think?

src/EVM/SymExec.hs
msooseth: This is used in checkAssert which is used everywhere in test.hs (100+ use-cases). It is also used as the primary way that we run `hevm symbolic --code CODE`.

I don't think this code is necessarily wrong? It picks up the standard issues that one likes to pick up without those in `cheatActions`, such as `asssertEq`. Or what do you think we should do instead?

src/EVM/UnitTest.hs
msooseth: Ah, nice catch! I added it with 83d35dc5 -- let me know what you think!

src/EVM.hs
```diff
@@ -1041,6 +1041,8 @@ exec1 conf = do
                 finishFrame (FrameReverted output)
             _ -> underrun
 
+        OpInvalid -> finishFrame (FrameReverted $ errorMsg "Invalid opcode: INVALID")
```
d-xo: I think we want to match the semantics of OpUnknown here? If I'm reading it right this would not consume all gas, which is incorrect.

msooseth: Oh wow, yes, gas is not consumed!

d-xo: I think we can match on branches that finished as failures with unknown opcode 0xfe and treat those as assertion violations?

src/EVM/UnitTest.hs
d-xo: I think that's probably fine? I would prefer to be sound as a default, I'm also fine to keep the existing concrete prefix extraction as a special case optimization and have this fully symbolic case as a fallback?

src/EVM.hs
msooseth: Ah OK, let me try to do that :)

src/EVM/UnitTest.hs
d-xo: and unless I'm misunderstanding, right now a symbolic prefix that could still be "assertion failed" would not be considered as an assertion (since concretePrefix wouldn't find it), and we would be missing potentially failing branches?

d-xo: are we even able to test this case? we probably need to use assembly actually and like pass some symbolic memory region into revert directly?

src/EVM/UnitTest.hs
msooseth: Done now :)

src/EVM/UnitTest.hs
msooseth: I had a go at this :) Can you please check if it's what you were thinking of?

The test case I will write later, good catch!

src/EVM.hs
msooseth: I reverted back, with a special case for `            Failure _ _ (UnrecognizedOpcode 0xFE) -> PBool False`  as discussed. Let me know what you think!

msooseth:
TODO: add tests cases -- especially for symbolic return values.

blishko (approved):

PR #754, opened 2025-05-21
Filter out duplicate read assumptions

blishko:
## Description

When we collect assumptions about read access, we might collect duplicate assumptions when the same buffer is accessed at overlapping intervals.
We can filter out these duplicates before we send them to the SMT solver.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko:
Hmm, I think I can add a test case. Let me try it.

blishko:
Added a test that is failing on main, but passes with this PR.

msooseth (approved):
Nice, let's merge! :)

PR #756, opened 2025-05-26
Masking rewrite rule

msooseth:
## Description
When either the Array offset is low or 0, Maps and Arrays cannot clash. Furthermore, I added one more rewrite rule that allows us to deal with Solidity argument packing.

I added a test for the masking rewrite rule. It also happens to fix an issue we had in one of our tests, so that has been enabled as well (should have had `ignoreTest`, not `expectFail`, now it's enabled)

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1094,6 +1094,10 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | x == 0 = b
       | otherwise = a
 
+    -- Masking as as per Solidity bit-packing of e.g. function parameters
+    go (And (Lit mask1) (Or (And (Lit mask2) _) x)) | (mask1 .&. mask2 == 0)
+         = And (Lit mask1) x
+
     -- address masking
     go (And (Lit 0xffffffffffffffffffffffffffffffffffffffff) a@(WAddr _)) = a
 
--- a/test/test.hs
+++ b/test/test.hs
@@ -2471,6 +2471,29 @@ tests = testGroup "hevm"
       putStrLnM $ "successfully explored: " <> show (Expr.numBranches expr) <> " paths"
       assertBoolM "The expression is NOT error" $ not $ any isError ret
       assertBoolM "The expression is NOT partial" $ not $ Expr.containsNode isPartial expr
+    , test "no-overapprox-when-present" $ do
+      Just c <- solcRuntime "C" [i|
+        contract ERC20 {
+          function f() public {
+          }
+        }
+
+        contract C {
+          address token;
+
+          function no_overapp() public {
+            token = address(new ERC20());
+            token.delegatecall(abi.encodeWithSignature("f()"));
+          }
+        } |]
+      let sig2 = Just (Sig "no_overapp()" [])
+      (expr, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig2 [] defaultVeriOpts
+      -- putStrLnM $ "expr: " <> show expr
+      putStrLnM $ "successfully explored: " <> show (Expr.numBranches expr) <> " paths"
+      assertBoolM "The expression is NOT error" $ not $ any isError ret
+      assertBoolM "The expression is NOT partial" $ not $ Expr.containsNode isPartial expr
+      let numCexes = sum $ map (fromEnum . isCex) ret
+      assertEqualM "number of counterexamples" 0 numCexes
     -- NOTE: below used to be symbolic copyslice copy error before new copyslice
     --       simplifications in Expr.simplify
     , test "overapproximates-undeployed-contract-symbolic" $ do
@@ -3953,8 +3976,7 @@ tests = testGroup "hevm"
                           _ -> False
           assertBoolM "Did not find expected storage cex" testCex
           putStrLnM "expected counterexample found"
-        ,
-        expectFail $ test "calling unique contracts (read from storage)" $ do
+        , test "calling-unique-contracts--read-from-storage" $ do
           Just c <- solcRuntime "C"
             [i|
               contract C {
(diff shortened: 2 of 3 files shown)


test/test.hs
```diff
@@ -3953,8 +3976,7 @@ tests = testGroup "hevm"
                           _ -> False
           assertBoolM "Did not find expected storage cex" testCex
           putStrLnM "expected counterexample found"
-        ,
-        expectFail $ test "calling unique contracts (read from storage)" $ do
+        , test "calling-unique-contracts--read-from-storage" $ do
```
blishko, line 3979: Can you explain what changed here?
What exactly is this test checking?
What was happening before and what is happening now?

src/EVM/Solvers.hs
```diff
@@ -161,6 +161,7 @@ getMultiSol smt2@(SMT2 cmds cexvars _) multiSol r inst availableInstances fileCo
       writeChan r Nothing
     Right _ -> do
       sat <- liftIO $ sendLine inst "(check-sat)"
+      when conf.dumpQueries $ liftIO $ writeSMT2File smt2 (show fileCounter <> "-origquery")
```
blishko: Should this be in a separate PR?

src/EVM/Expr.hs
```diff
@@ -1079,6 +1088,10 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | x == 0 = b
       | otherwise = a
 
+    -- Masking as as per Solidity bit-packing of e.g. function parameters
+    go (And (Lit mask1) (Or (And (Lit mask2) _) x)) | (mask1 .&. mask2 == 0)
+         = And (Lit mask1) x
+
```
blishko, line 1100: Nice!
Do we get this pattern also in optimized bytecode? Or only unoptimized?
Meaning, is this maybe some optimization that solc should be doing but is not?

src/EVM/Expr.hs
```diff
@@ -643,16 +643,25 @@ readStorage w st = go (simplifyNoLitToKeccak w) (simplifyNoLitToKeccak st)
...
+      (ArraySlotZero idA, MappingSlot idB _)          | isArray idA, isMap' idB  -> go slot prev
+      (ArraySlotZero idA, Keccak idB)                 | isArray idA, isMap idB  -> go slot prev
+
+      -- as above, but with low concrete offset
+      -- NOTE: ArraySlotWithOffs2 could only be rewritten if both offsets are Lit (i.e. concrete)
+      --       but then the simplifier rewrites it to ArraySlotWithOffs
+      (MappingSlot idA _, ArraySlotWithOffs idB (Lit offs)) | isMap' idA, isArray idB, offs < 1000  -> go slot prev
+      (Keccak idA, ArraySlotWithOffs idB (Lit offs))        | isMap idA, isArray idB, offs < 1000 -> go slot prev
+      (ArraySlotWithOffs idA (Lit offs), Keccak idB)        | isArray idA, isMap idB, offs < 1000 -> go slot prev
+      (ArraySlotWithOffs idA (Lit offs), MappingSlot idB _) | isArray idA, isMap' idB, offs < 1000 -> go slot prev
```
blishko: Why is the magic number `1000`?

blishko (commented):
Nice!
I am going to trust you with all this map + array business :)
The rewrite rule is nice!

src/EVM/Solvers.hs
msooseth: Good point, removed, added to another PR! It was useful to debug why the multiSol was giving us so many solutions -- it turns out because the value was symbolic.

src/EVM/Expr.hs
msooseth: I moved this away. I'll make the magic number 256, it's been used before as a good magic separation number. if it's allowed to be very large then it's possible to manipulate. I'll explain in the other PR :)

test/test.hs
msooseth: I can partially explain. This is the contract:

```solidity
              contract C {
                uint x;
                A a;

                function call_A() public {
                  a = new A();
                  // should fail since x can be anything
                  assert(a.x() == x);
                }
              }
              contract A {
                uint public x;
              }
```

The `a = new A()` deploys A into a symbolic address `freshSymAddr1`. Then, the `a.x()` tries to unmask the address, but it can't without the rewrite rule. So it runs the `multiSol` but it gets many different solutions for `freshSymAddr1`. So it overapproximates (since it's a staticcall it can overapproximate --- the auto-generated getter function `x()` is `public view` so it becomes a `staticcall`). Which happens to lead to TWO Cex-es, one spurious.

The multiSol query looks like:

```
   1   │ ;(PEq
   2   │ ;  (Var "multiQueryVar")
   3   │ ;  (And
   4   │ ;    1461501637330902918203684832716283019655932542975
   5   │ ;    (Or
   6   │ ;      (And
   7   │ ;        115792089237316195423570985007226406215939081747436879206741300988257197096960
   8   │ ;        (SLoad
   9   │ ;          slot:
  10   │ ;            1
  11   │ ;          storage:
  12   │ ;            (AbstractStore (SymAddr "entrypoint") Nothing)
  13   │ ;        )
  14   │ ;      )
  15   │ ;      (WAddr
  16   │ ;        (SymAddr "freshSymAddr1")
  17   │ ;      )
  18   │ ;    )
  19   │ ;  )
  20   │ ;)
````

Seems like the masking is added on top of the storage load operation. Kinda weird, some Solidity quirk perhaps...

msooseth: The test case "no-overapprox-when-present" that I added also illustrates this issue. There also a new symbolic variable, `freshSymAddr1` is created, and then due to the masking, the `multiSol` is invoked, which chokes, because there are many solutions, so there also it overapproximates.

src/EVM/Expr.hs
msooseth: Yes. It happens for ABI unpacking: https://docs.soliditylang.org/en/latest/abi-spec.html#abi-packed-mode and seems also to be triggered for addresses that are called that are stored in storage/memory -- hence the test enabled & added one now working!

PR #761, opened 2025-06-09
Allow signed integers to be in the expected range

gustavo-grieco:
## Description

The encoding of signed integers from the ABI was incorrect. Added a small function to provide the correct constraints. This PR will need a good number of tests, but you can start taking a look since the code changes is small.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1633,6 +1633,8 @@ containsNode p = getAny . foldExpr go (Any False)
 inRange :: Int -> Expr EWord -> Prop
 inRange sz e = PAnd (PGEq e (Lit 0)) (PLEq e (Lit $ 2 ^ sz - 1))
 
+inRangeSigned :: Int -> Expr EWord -> Prop
+inRangeSigned sz e = PAnd (PLEq e (Lit $ 2 ^ (sz - 1) - 1)) (PLT e (Lit $ 2 ^ (sz - 1)))
 
 -- | images of keccak(bytes32(x)) where 0 <= x < 256
 preImages :: [(W256, Word8)]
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -1646,6 +1646,9 @@ containsNode p = getAny . foldExpr go (Any False)
 inRange :: Int -> Expr EWord -> Prop
 inRange sz e = PAnd (PGEq e (Lit 0)) (PLEq e (Lit $ 2 ^ sz - 1))
```
msooseth: Can you please also fix this? `PGEq e (Lit 0)` is trivially PBool True, so it can be removed.

msooseth (changes requested):
Really nice catch! Can you please fix the other one, too, and then we can merge :)

msooseth:
Oh, and also, please add a line to the Changelog!

src/EVM/Expr.hs
```diff
@@ -1646,6 +1646,9 @@ containsNode p = getAny . foldExpr go (Any False)
...
 
+inRangeSigned :: Int -> Expr EWord -> Prop
+inRangeSigned sz e = PAnd (PGEq e (Lit 0)) (PLEq e (Lit $ 2 ^ (sz - 1) - 1))
```
msooseth:
```suggestion
inRangeSigned sz e = PLEq e (Lit $ 2 ^ (sz - 1) - 1)
```

The other thing is `PBool True`

src/EVM/Expr.hs
gustavo-grieco: Done

msooseth (approved):
Thanks, nice!

PR #762, opened 2025-06-09
Correctly print and return partials path logs when symbolic exploration is done

gustavo-grieco:
## Description

This small PR contains two changes:
1. It makes sure partial path warnings are correctly printed when debug is true (this was incorrectly refactored)
2. It returns the list of logs to the caller can decide what to do. Typically, if you want to perform verification of some code, you are expecting that the partial log is empty when the exploration is done.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -649,9 +649,6 @@ printPartialIssues flattened call =
                 <> T.pack call <> " due to the following issue(s):"
     T.putStr . T.unlines . fmap (indent 5 . ("- " <>)) . fmap formatPartial . getPartials $ flattened
 
-getAllPartials :: [Expr End] -> [Text]
-getAllPartials = fmap formatPartial . getPartials
-
 getPartials :: [Expr End] -> [PartialExec]
 getPartials = mapMaybe go
   where
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -692,7 +692,7 @@ verifyInputs
   -> Fetch.Fetcher Symbolic m RealWorld
   -> VM Symbolic RealWorld
   -> Maybe (Postcondition RealWorld)
-  -> m (Expr End, [(SMTResult, Expr End)], [Text])
+  -> m (Expr End, [(SMTResult, Expr End)], [PartialExec])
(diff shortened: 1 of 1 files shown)


src/EVM/SymExec.hs
```diff
@@ -689,7 +692,7 @@ verifyInputs
   -> Fetch.Fetcher Symbolic m RealWorld
   -> VM Symbolic RealWorld
   -> Maybe (Postcondition RealWorld)
-  -> m (Expr End, [(SMTResult, Expr End)])
+  -> m (Expr End, [(SMTResult, Expr End)], [Text])
```
msooseth: Can we instead return

```
-> m (Expr End, [(SMTResult, Expr End)], [Expr End])
```

where the last `Expr End` is the set of partial Expr End-s? Then:

```
getAllPartials = getPartials
```

And if you need it printed, you can just call `fmap formatPartial` on it. I think it's cleaner than returning a `[Text]`, no?

src/EVM/SymExec.hs
```diff
@@ -702,12 +705,15 @@ verifyInputs solvers opts fetcher preState maybepost = do
   when conf.dumpExprs $ liftIO $ do
     T.writeFile "simplified.expr" (formatExpr expr)
     T.writeFile "simplified-conc.expr" (formatExpr $ Expr.simplify $ mapExpr Expr.concKeccakOnePass expr)
+
+  let partials = getAllPartials flattened
+  when conf.debug $ liftIO $ do
```
msooseth, line 707: Ah, good catch, thanks!

src/EVM/SymExec.hs
gustavo-grieco: `getPartials` returns `[PartialExec]` instead of `[Expr End]`. Do you want to return that instead?

msooseth: Ah, that is even better, great idea!

gustavo-grieco: Done!

gustavo-grieco:
I think this looks ready, but feel free to take another look.

msooseth (approved):
LGTM!

PR #765, opened 2025-06-16
[DRAFT] More rewrites for array/map clash

msooseth:
## Description
More rewrites so we don't give up in case of map vs array in `readStorage`, but know that they cannot clash. Here is how we know they cannot clash:
* Maps and Arrays hash a different SIZE input: 64B vs 32B. Hence, you'd need a collision of Keccak for them to collide
* HOWEVER, when there is an offset, then the offset could  be set by the attacker to be keccak(a)-keccak(b). So when an offset is allowed, it must be limited. We limited the offset to 256 is other places:

```haskell
      -- Finding two Keccaks that are < 256 away from each other should be VERY hard
      -- This simplification allows us to deal with maps of structs
      (Add (Lit a2) (Keccak _), Add (Lit b2) (Keccak _)) | a2 /= b2, abs (a2-b2) < 256 -> go slot prev
      (Add (Lit a2) (Keccak _), (Keccak _)) | a2 > 0, a2 < 256 -> go slot prev
      ((Keccak _), Add (Lit b2) (Keccak _)) | b2 > 0, b2 < 256 -> go slot prev
```

so I'm limiting it here to the same offset. Which means it's impossible to find a collision that are within 256 of each other. This is accepted to be impossible. 

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -639,20 +639,26 @@ readStorage w st = go (simplifyNoLitToKeccak w) (simplifyNoLitToKeccak st)
       (a, b) | surelyEqual a b -> Just val
       (a, b) | surelyNotEqual a b -> go slot prev
 
-      -- slot is for a map + map -> skip write
-      (MappingSlot idA _, MappingSlot idB _)       | isMap' idB, isMap' idA, idsDontMatch idA idB  -> go slot prev
-      (MappingSlot idA keyA, MappingSlot idB keyB) | isMap' idB, isMap' idA, surelyNotEqual keyA keyB -> go slot prev
+      (MappingSlot idA _, MappingSlot idB _)                | isMap' idB, isMap' idA, idsDontMatch idA idB  -> go slot prev
+      (MappingSlot idA keyA, MappingSlot idB keyB)          | isMap' idB, isMap' idA, idA == idB, surelyNotEqual keyA keyB -> go slot prev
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -663,40 +669,27 @@ readStorage w st = go (simplifyNoLitToKeccak w) (simplifyNoLitToKeccak st)
...
-        surelyNotEqual a b -> go slot prev
-      (ArraySlotWithOffs idB offB2, ArraySlotWithOffs2 idA offA1 offA2)        | isArray idA, idA == idB,
-        surelyNotEqual (Add (Lit offA1) offA2) offB2 -> go slot prev
       (ArraySlotWithOffs2 idA offA1 offA2, ArraySlotWithOffs idB offB2)        | isArray idA, idA == idB,
         surelyNotEqual (Add (Lit offA1) offA2) offB2 -> go slot prev
       (ArraySlotWithOffs2 idA offA1 offA2, ArraySlotWithOffs2 idB offB1 offB2) | isArray idA, idA == idB,
         surelyNotEqual (Add (Lit offA1) offA2) (Add (Lit offB1) offB2) -> go slot prev
 
+      -- DANGEROUS: offset could possibly be VERY large and SPECIFIC in order to make them clash
+      -- This would constitute a hack, and would require a specifically badly written code
```
blishko, line 687: Are we sure we want to do these rewrites?

msooseth:
Closing, not so important.

Issue #725, opened 2025-04-30
Allow to dump only queries that returned timeout or unknown


gustavo-grieco:
Dumping the queries as smt files is very useful, but having all of them is usually an overkill. It will be more useful if we could select to dump only the queries that returned either unknown or timeout so we can easily re-run them with another solver to see if we could get better results. 
So instead of `dumpQueries : Bool` we could have a special type: `None | OnlyUnsolved | All`

msooseth:
This has now been done. Closing.

PR #766, opened 2025-06-16
RFC: Dump unsolved queries into a particular path

gustavo-grieco:
## Description

A small change to dump unsolved queries into a particular directory (#725). The idea is to allow users to keep unsolved so they can use other solvers or longer timeouts. If they solve it and the result is SAT, they won´t be able to easily extract the counter example, but at least, they will be able to know that their invariants do not hold.
If you like the idea, I can change the other options to also dump into a particular folder instead in the current directory. This makes the integration with 3rd-party software much easier.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Solvers.hs
+++ b/src/EVM/Solvers.hs
@@ -206,6 +206,12 @@ getMultiSol smt2@(SMT2 cmds cexvars _) multiSol r inst availableInstances fileCo
            writeChan r Nothing
         "unknown" -> liftIO $ do
            when conf.debug $ putStrLn "Unknown result by SMT solver."
+           case conf.dumpUnsolved of
+             Just fp -> do
+              let filename = fp <> "unsolved-" <> show fileCounter <> "-sol" <> show (length vals)
+              liftIO $ putStrLn $ "Dumping unsolved SMT query to: " <> filename  
+              liftIO $ writeSMT2File fullSmt filename
+             Nothing -> pure ()
            writeChan r Nothing
         "sat" -> do
           if length vals >= multiSol.maxSols then liftIO $ do
--- a/src/EVM/Effects.hs
+++ b/src/EVM/Effects.hs
@@ -35,6 +35,7 @@ data Config = Config
   { dumpQueries      :: Bool
   , dumpExprs        :: Bool
   , dumpEndStates    :: Bool
+  , dumpUnsolved     :: Maybe FilePath
   , debug            :: Bool
   , dumpTrace        :: Bool
   , numCexFuzz       :: Integer
@@ -56,6 +57,7 @@ defaultConfig = Config
   { dumpQueries = False
   , dumpExprs = False
   , dumpEndStates = False
+  , dumpUnsolved = Nothing
   , debug = False
   , dumpTrace = False
   , numCexFuzz = 10


src/EVM/Solvers.hs
```diff
@@ -206,6 +206,12 @@ getMultiSol smt2@(SMT2 cmds cexvars _) multiSol r inst availableInstances fileCo
            writeChan r Nothing
         "unknown" -> liftIO $ do
            when conf.debug $ putStrLn "Unknown result by SMT solver."
+           case conf.dumpUnsolved of
+             Just fp -> do
+              let filename = fp <> "unsolved-" <> show fileCounter <> "-sol" <> show (length vals)
+              liftIO $ putStrLn $ "Dumping unsolved SMT query to: " <> filename  
+              liftIO $ writeSMT2File fullSmt filename
+             Nothing -> pure ()
```
msooseth, line 214: This should also go to `getOneSol`. `getMultiSol` is never used to check for a Cex. It is instead used to extract e.g. potential jump destinations. The hardest queries all go to `getOneSol`.

Notice that even in `getOneSol`, there is no guarantee that if it's SAT, there is a Cex that can be generated. We actually dispatch SMT queries also during interpretation (e.g. to check for loop conditions), not only when we have a Cex in mind (i.e. checking if the erroneous endstate is reachable).

Nevertheless, it would be helpful to dump the unsolved queries -- it would be helpful as a debug measure, so we can know what wasn't possible to solve, and perhaps write rewrite rules to make these queries faster. So I like the idea. However, it certainly needs to go (also) into `getOneSol`, and we should be careful not to tell users much about what these *mean*, because we may not be able to guarantee where they come from.

msooseth (changes requested):
I actually like the idea! See details in my comment. Also please let's expose it in cli.hs as a common command line option so it's easily accessible.

gustavo-grieco:
> I actually like the idea! See details in my comment. Also please let's expose it in cli.hs as a common command line option so it's easily accessible.

Do you think I should do the same to dump all the queries to a particular path or only these are fine?

src/EVM/Solvers.hs
gustavo-grieco: Uhm, perhaps it is better to dump expressions associated with End Expr and otherwise allow to dump everything as we already have?

msooseth: We actually don't see what the query is about -- End Expr or otherwise, when in Solver.hs, and I'd prefer to keep it like that -- this way we are agnostic towards the queries that come in, which I think is a good idea. Would you wanna give it a go to fix this up, or should I have a go at it? I like the idea, so I think we should just do it the minor fixes required, and then we could merge! I think it would be very helpful to everyone :)

msooseth:
> > I actually like the idea! See details in my comment. Also please let's expose it in cli.hs as a common command line option so it's easily accessible.
> 
> Do you think I should do the same to dump all the queries to a particular path or only these are fine?

Sure, that's fine, to give a path name. I actually like it.

src/EVM/Solvers.hs
gustavo-grieco: I should be able to handle this small PR, but thanks :)

msooseth:
Closing in favour of https://github.com/ethereum/hevm/pull/779 so both you and I can push into it. I wanna do some small changes, and then merge :) I like the idea too much not to merge soon :)

PR #773, opened 2025-06-19
More Prop rewrite rules for PLT and PEq

msooseth:
## Description
These rewrite rules allow us to rewrite things like `(PLT (Lit 0) (LT a b))` into `(PLT a b)`, in two steps. It's a lot cleaner and more complete than the previous setup we had.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
LGMT!

msooseth:
Thanks for the review @blishko !

PR #777, opened 2025-06-20
Use the pranked address' nonce

### LINKED ISSUE #775: Consecutive pranked deployment in forge setUp revert
Environment: 
```
HEVM 0.54.2, static exec, arch linux x64
forge Version: 1.2.3-stable
Forge Commit SHA: a813a2cee7dd4926e7c56fd8a785b54f32e0d10f
```

Using the following forge test contract based on the official docs:

```solidity
/// SPDX-License-Identifier: GPL-3.0-or-later
pragma solidity 0.8.26;

contract MyVault {
    mapping(address => uint256) public balance;
    function deposit() external payable {
        balance[msg.sender] += msg.value;
    }
}

contract SymbolicTest is Test {
    function setUp() public {
        address caller = address(0xdead);
        vm.startPrank(caller);
        new MyVault();

        // Without this line, setUp and the whole
        // test suite completes without issues.
        new MyVault();
    }

    function prove_correct(uint8 amt) public {
        assert(1 == 1);
      }
}

```

TestERC20 and is a minimal instance of openzeppelin erc20, its content is irrelevant to the bug

If the line `new MyVault();` is removed, the test suite completes without problems. 
Adding it back 

```
# forge  clean
# forge build --ast
# hevm test

Failure: setUp
  src/test/Symbolic.t.sol:SymbolicTest
   ├╴constructor
   └╴setUp()
      ├╴call HEVM::startPrank(0x000000000000000000000000000000000000dEaD) (src/test/Symbolic.t.sol:20)
      │  └╴← 0x
      ├╴create 0x2387b3383E89c164781d173B7Aa14d9c46eD2642 (src/test/Symbolic.t.sol:21)
      │  └╴← 2523 bytes of code
      └╴error Revert 0x (lib/forge-std/src/Base.sol:27)
```

The reported line contains:

```
    // Cheat code address, 0x7109709ECfa91a80626fF3989D68f67F5b1DD12D.
    address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));

    Vm internal constant vm = Vm(VM_ADDRESS); // line 27
```

However this line seems to be a misleading revert location, as it is also inconsistent. In around 12% of cases the revert selects another library contract from `lib/` with `len(max_lines_of_file)+1` as the revert location

msooseth:
## Description
We used the non-pranked' address' nonce, instead of the nonce of the pranked address' nonce when doing OpCreate. This lead to clash of addresses being calculated, because the nonce was never updated -- only the pranked address' nonce was updated, correctly. Now we do the address calculation correctly.

This only fixes the case when the address exists. When it doesn't, and we are running in concrete execution mode (such as in `setUp`, then we return a symbolic address, and that's not supported in concrete execution mode, so the system bails out.

Fixes #775

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - SignExtend is now correctly being constant-folded
 - Some of our property-based testing was ineffective because of inadvertent
   simplification  happening before calling the SMT solver. This has now been fixed.
+- When pranking an address, we used the non-pranked address' nonce
+  to calculate the new address. This was incorrect, and lead to address clash,
+  as the nonce was never incremented.
 
 ## Changed
 - Warnings now lead printing FAIL. This way, users don't accidentally think that
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -864,7 +864,11 @@ exec1 conf = do
                   assign (#state % #overrideCaller) Nothing
                   assign (#state % #resetCaller) False
 
-                newAddr <- createAddress from' this.nonce
+                nonce <- zoom (#env % #contracts) $ do
+                  contr <- use (at from')
+                  pure $ maybe Nothing (\c -> c.nonce) contr
+
+                newAddr <- createAddress from' nonce
                 _ <- accessAccountForGas newAddr
                 burn' cost $ do
                   initCode <- readMemory xOffset xSize
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -864,9 +864,10 @@ exec1 conf = do
                   assign (#state % #overrideCaller) Nothing
                   assign (#state % #resetCaller) False
 
+                touchAddress from'
                 nonce <- zoom (#env % #contracts) $ do
                   contr <- use (at from')
-                  pure $ maybe Nothing (\c -> c.nonce) contr
+                  pure $ (\c -> c.nonce) =<< contr
 
                 newAddr <- createAddress from' nonce
                 _ <- accessAccountForGas newAddr
@@ -979,6 +980,8 @@ exec1 conf = do
                         assign (#state % #overrideCaller) Nothing
                         assign (#state % #resetCaller) False
 
+                      touchAddress from'
+
                       let (cost, gas') = costOfCreate fees availableGas xSize True
                       newAddr <- create2Address self xSalt initCode
                       _ <- accessAccountForGas newAddr
@@ -1053,10 +1056,6 @@ transfer _ _ (Lit 0) = pure ()
 transfer src dst val = do
   sb <- preuse $ #env % #contracts % ix src % #balance
   db <- preuse $ #env % #contracts % ix dst % #balance
-  baseState <- use (#config % #baseState)
-  let mkc = case baseState of
-              AbstractBase -> unknownContract
-              EmptyBase -> const emptyContract
   case (sb, db) of
     -- both sender and recipient in state
     (Just srcBal, Just _) ->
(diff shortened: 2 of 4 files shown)


elopez:
Good catch! I'm seeing crashes in echidna with this PR though, I suppose it is related with this thing you mention
> This only fixes the case when the address exists. When it doesn't, and we are running in concrete execution mode (such as in setUp, then we return a symbolic address, and that's not supported in concrete execution mode, so the system bails out.

could we do something about it?

```plain
Internal Error: concrete vm, shouldn't ever happen -- CallStack (from HasCallStack):
  internalError, called at src/EVM.hs:3215:14 in hevm-0.54.2-inplace:EVM
  error, called at src/EVM/Types.hs:1518:19 in hevm-0.54.2-inplace:EVM.Types
  internalError, called at src/EVM.hs:3215:14 in hevm-0.54.2-inplace:EVM
```

this is the contract I used in Echidna by the way. `echidna_startPrankExist` doesn't crash, but `echidna_startPrank` and `echidna_prank` do.

```solidity
interface VM {
    function prank(address newSender) external;
    function startPrank(address newSender) external;
}

contract C {
}

contract Test1 {
  VM constant vm = VM(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);

  function echidna_startPrankExist() public returns (bool) {
    address caller = address(new C());
    vm.startPrank(caller);
    new C();
    new C();
    return true;
  }

  function echidna_startPrank() public returns (bool) {
    address caller = address(0xdead);
    vm.startPrank(caller);
    new C();
    new C();
    return true;
  }

  function echidna_prank() public returns (bool) {
    address caller = address(0xdead);
    vm.prank(caller);
    new C();
    vm.prank(caller);
    new C();
    return true;
  }
}
```

elopez:
I suppose that for the concrete case where you're pranking an address that doesn't exist, we could initialize it to be an empty contract, similarly to how `transfer` handles it for when you send ether to a non-existent address.

elopez:
something like this appears to work fine (although it can surely be made prettier and deduplicated with `transfer`). An analogue should also be implemented for CREATE2

```diff
diff --git a/src/EVM.hs b/src/EVM.hs
index a0de4d88..a669c2d8 100644
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -864,6 +864,12 @@ exec1 conf = do
                   assign (#state % #overrideCaller) Nothing
                   assign (#state % #resetCaller) False

+                baseState <- use (#config % #baseState)
+                let mkc = case baseState of
+                            AbstractBase -> unknownContract
+                            EmptyBase -> const emptyContract
+                (#env % #contracts) %= (Map.insertWith (\_ e -> e) from' (mkc from'))
+
                 nonce <- zoom (#env % #contracts) $ do
                   contr <- use (at from')
                   pure $ maybe Nothing (\c -> c.nonce) contr
```

msooseth:
> something like this appears to work fine (although it can surely be made prettier and deduplicated with `transfer`). An analogue should also be implemented for CREATE2
> 
> ```diff
> diff --git a/src/EVM.hs b/src/EVM.hs
> index a0de4d88..a669c2d8 100644
> --- a/src/EVM.hs
> +++ b/src/EVM.hs
> @@ -864,6 +864,12 @@ exec1 conf = do
>                    assign (#state % #overrideCaller) Nothing
>                    assign (#state % #resetCaller) False
> 
> +                baseState <- use (#config % #baseState)
> +                let mkc = case baseState of
> +                            AbstractBase -> unknownContract
> +                            EmptyBase -> const emptyContract
> +                (#env % #contracts) %= (Map.insertWith (\_ e -> e) from' (mkc from'))
> +
>                  nonce <- zoom (#env % #contracts) $ do
>                    contr <- use (at from')
>                    pure $ maybe Nothing (\c -> c.nonce) contr
> ```

Ah nice! I like this. Do you wanna push a version of that into this PR, such that it works for create&create2, etc?

PR #779, opened 2025-06-25
[DRAFT] Dump unsolved queries, use "empty-smt-solver"

msooseth:
## Description
As per @gustavo-grieco PR of #766  but now in here, so I can push into it :)

A small change to dump unsolved queries into a particular directory (https://github.com/ethereum/hevm/issues/725). The idea is to allow users to keep unsolved so they can use other solvers or longer timeouts. If they solve it and the result is SAT, they won´t be able to easily extract the counter example, but at least, they will be able to know that their invariants do not hold.

This also introduces the "empty" solver type that simply responds "unknown" to all SMT queries. This can be extremely helpful in debugging.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -6,17 +6,17 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1751304424,
-        "narHash": "sha256-PALfH4kbHq0U2zcD8yfs5W2fgz2+0KtPmn/zMPJVHuo=",
+        "lastModified": 1751304960,
+        "narHash": "sha256-6+GUDEY6aLyq5KLRaIMeBEVlKiYymKi4CB5B8jG6wY0=",
         "owner": "msooseth",
         "repo": "empty-smt-solver",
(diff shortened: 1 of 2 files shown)


msooseth:
@gustavo-grieco let me know what you think. We could also refactor the other options to put the files into a subdirectory. Do you want me to do it, or do you wanna do it?

msooseth:
Hey @blishko what do you think about this changeset?

blishko:
I think this can be split into two PRs, no? The empty solver and the dumping of unsolved queries are two independent things, no?
Also, I would very much prefer if we did not have to the empty solver as an external process.
I wonder how much of a change is needed to allow empty solver implementation directly in our codebase.

msooseth:
Good point. Let me split, close, and open 2 PRs :)

msooseth:
This has now been superseeded by #814 #815

PR #782, opened 2025-06-26
Small update to SMT encoding of MulMod and AddMod

blishko:
We apply the following changes:
- Instead of concatenating vector of 0 bits,  we can use SMT-LIB build-in `zero_extend`. Maybe the solvers has better native support for that.
- Instead of having `ite` nested under the `extract` operation, we can pull it to the top-level and do the extension and shortening only in one branch.
- For AddMod, we do not need to extend all the way to 512 bits, extending just one bit to 257 bits is sufficient.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -721,18 +721,18 @@ exprToSMT = \case
     aExp <- exprToSMT a
     bExp <- exprToSMT b
     cExp <- exprToSMT c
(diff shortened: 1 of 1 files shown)


msooseth (approved):
LGTM, of course! AddMod only needs one more bit!

PR #783, opened 2025-06-27
Adding a number of serious improvements to dispatched SMT2 query

msooseth:
## Description
Major changes:

### concKeccakSimpProps/concKeccakProps
* `concKeccakSimpProps` allows us to do simplification + concretization of Props until fixedpoint.
* `concKeccakProps` is only concretization

We now have a helper function for `assertProps`. We check if we are supposed to simplify. If so, we simplify and decompose at toplevel, and then run the real `assertPropsHelper`, with an indicator whether we are supposed to simplify inside during concretization.

### Simplification Rules
I have added a number of new simplification rules. This should help us improve our simplifications. Especially related to signed operations, but also AddMod and MulMod

### Improved Keccak axioms
When gathering Keccak axioms, we are looking for (Keccak bs). But we can (and should) always simplify `bs` as much as possible. We forgot to do this before.

### SMT2
The final SMT2 is now a single line, `SMT2 (fmap (\p -> "(assert " <> p <> ")") encs) (cexInfo storageReads) ps`, which encodes all the final asserts, injects all we need for the CEX recovery, and adds the comments (`ps`). The comments is now the concretized (and normally simplified) `ps` rather than the `psPreConc` -- which could be extremely large.

### Test cases for SignExtend
Some extra test cases for SignExtend

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
I think this looks good!

I haven't spotted any obvious problem :)
And we have discussed most of these changes offline.

PR #784, opened 2025-06-28
RFC: Futamura projection proof-of-concept

gustavo-grieco:
## Description

This PR was created just to receive comments regarding a potential usage of the futamura projection. It is not working as it needs a few important pieces more:
1. Take the bytecodes, splits it in basic blocks (with external calls considered as the end of a basic block as well), and save each futamura projection. This needs to be done for every contract involved.
2. Write a top level function, similar to `run` (EVM/Exec.hs) which fetches the current futamura projection and runs it. It should find the next basic block and repeat until the transaction is over.

However, you can take a look a the current proof of concept to see if you like the approch.

The final implementation will requiere a large refactoring of the opcode implementation (but it should not be too hard, since all the code is there). In the meantime, some limited benchmarking could be done, once steps 1 and 2 are complete, using a small number of opcodes in very simple contracts. The final implementation will also require the user to install ghc (or stack) to use this feature.

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -104,6 +104,7 @@ library
     EVM.Format,
     EVM.Fetch,
     EVM.FeeSchedule,
+    EVM.Futamura,
     EVM.Op,
     EVM.Precompiled,
     EVM.RLP,
@@ -141,6 +142,10 @@ library
   install-includes:
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
+    ghc                               >= 9.6 && < 10,
+    ghc-paths                         >= 0.1,
+    directory                         >= 1.3,
+    temporary                         >= 1.2.0 && < 1.4, 
     system-cxx-std-lib                >= 1.0 && < 2.0,
     QuickCheck                        >= 2.13.2 && < 2.15,
     Decimal                           >= 0.5.1 && < 0.6,
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -29,7 +29,7 @@ import EVM.Effects (Config (..))
 
 import Control.Monad (unless, when)
 import Control.Monad.ST (ST)
-import Control.Monad.State.Strict (MonadState, State, get, gets, lift, modify', put)
+import Control.Monad.State.Strict (MonadState, State, StateT, get, gets, lift, modify', put)
 import Data.Bits (FiniteBits, countLeadingZeros, finiteBitSize)
 import Data.ByteArray qualified as BA
 import Data.ByteString (ByteString)
@@ -63,6 +63,7 @@ import Data.Vector.Storable.ByteString (vectorToByteString)
 import Data.Word (Word8, Word32, Word64)
 import Text.Read (readMaybe)
 import Witch (into, tryFrom, unsafeInto, tryInto)
+import Witch.From (From)
 
 import Crypto.Hash (Digest, SHA256, RIPEMD160)
 import Crypto.Hash qualified as Crypto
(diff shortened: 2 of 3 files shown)


msooseth:
Oh wow, this is really cool! I am very busy with writing a presentation for a 4h seminar I will be giving on smart contract verification next week, but I will be back full speed on this and other PRs in about 1-1.5 weeks! This actually looks like a great place to start :) Thanks for this, I really needed this initial kick-off, I think I will be able to do some things from here on!

d-xo:
Very exciting! I'm curious if you had considered at all generating llvm directly instead of haskell?

gustavo-grieco:
Current state of this PR:

* Basic blocks are fully produced (but probably the code to split them is not 100% correct)
* Basic blocks should be properly switch when the PC jumps into them
* Code produced is not very efficient.

Missing components:
* Only a few opcodes implemented. Jumps and calls are missing so we don't know if the basic block switching really works.
* No support for handling more than one contract, nor deploying new ones.

Demo of the code produced:
```
$ hevm exec --code 0x5F5F9000005FFD  --use-futamura
Splitting into basic blocks: [((0,4),[OpPush0,OpPush0,OpSwap 1,OpStop]),((4,5),[OpStop]),((5,7),[OpPush0,OpRevert])]
...
```

The code produced contains this one:

```haskell
runBasicBlock_0_4 :: StateT (VM Concrete s) (ST s) ()
runBasicBlock_0_4 = do
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodePush0
     Just r -> return ()
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodePush0
     Just r -> return ()
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodeSwap (1 :: Int)
     Just r -> return ()
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodeStop
     Just r -> return ()
runBasicBlock_4_5 :: StateT (VM Concrete s) (ST s) ()
runBasicBlock_4_5 = do
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodeStop
     Just r -> return ()
runBasicBlock_5_7 :: StateT (VM Concrete s) (ST s) ()
runBasicBlock_5_7 = do
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodePush0
     Just r -> return ()
 get >>= \vm ->
   case vm.result of
     Nothing -> let ?op = 1 in runOpcodeRevert
     Just r -> return ()
```

gustavo-grieco:
I finally implemented enough opcodes to be able to benchmark a simple loop, with some operations and different revert conditions. The contract I'm using for benchmark looks like this:

```solidity
contract C {
	fallback() payable external {
        uint256 z = 1;
        for (int i = 0; i < (10 ** 6) * 8; i++) {
        	z = 2 * z + 1;
		z = z + 1;
                z = z / 5;
                z = z + 1;
                z = z * 6;
                z = (z * 5) / 10;
        }
        if ((z % 1) == 0) 
	        revert("one");
        if ((z % 2) == 0) 
	        revert("two");
        if ((z % 3) == 0) 
	        revert("three");
        if ((z % 4) == 0) 
	        revert("four");
        }
}
```

Currently, the basic blocks are stored in a map, and then called when need it, using the `dispatcherLoop` function. 
The prototype contains everything needed to run the contract and compare it to the non-futamura execution (e.g. using the interpreter). Unfortunately, **this iteration is around 5% to 10% slower slower than the interpreter** for this contract. I believe there are a few reasons for it:

1. The use of the stack is hard to optimize: it is immutable and requires to be copied every time it is used. There are different options here, either we do stack unboxing where the top of the stack is moved into another data structure and updated when a basic block has finished executing, or we remove the stack completely. Removing the stack means to have a way to compute pure expressions from every basic block, which what they did [the authors of EFCF](https://github.com/uni-due-syssec/efcf-framework).

2. The dispatcher loop calls the compiled basic block code as pointers. This means that GHC cannot see the complete program during the compilation process, and therefore, cannot apply optimizations beyond the basic block level. It seems that basic block optimizations are not enough, and if it is able to reasoning about the complete program, including loops, it will be able to know which parts of the VM state are changed. In order to workaround this, we will need to generate the complete program as Haskell code, not only basic block. 
To get that, we will need to know how basic block are connected so we can chain then with if conditions like this: 
```solidity
runBasicBlock_0_13 :: StateT (VM Concrete s) (ST s) ()
runBasicBlock_0_13 = do
 runOpcodePush 1
 ...
 runOpcodeJumpi
  if cond /= 0
    then runBasicBlock_25_48
    else runBasicBlock_48_49 
```

gustavo-grieco:
I need to discuss with the rest of the team regarding this prototype, but my impression on what needs to be done to unlock full speed here:

1. Process gas and PC at the basic block level, so it will only get updated once. This will mean that, if the gas is not enough to execute the complete basic block, then we can abort the execute before it is finishes. For the PC is easier, since we know the value of it at each point of the basic block.
2. Compute stack items (e.g. last three elements) at the basic block level, this could allow to only execute pure expressions at the end of each basic block instead of the intermediate steps.

gustavo-grieco:
I will come back to this once @blishko have some stable code for the concrete execution.

PR #794, opened 2025-07-15
Rewrite exponentiation to bitshift

blishko:
The legacy codegen of Solidity uses exponentiation to simulate bitshift. This was needed back when EVM did not have left bitshift instruction. See also https://github.com/ethereum/solidity/issues/16117 This pattern occurs when storing a value smaller that a storage slot size, i.e., when Solidity packs multiple values into one storage slot.

We can detect this pattern and replace it with actual bitshift. This should remove one possible source of exponentiation and allow us to verify more contracts.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1186,17 +1186,11 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       where l = sort [a, b, c]
             an = EVM.Expr.and
 
-    -- A pattern used by Solidity's legacy codegen when storing value of size < 32 bytes
-    -- into a storage slot (which has size 32 bytes)
-    -- It uses exponentiation to simulate bit shift.
-    -- This is needed on EVMs that do not support left shift (before Constantinopole), but is still part of legacy codegen in Solidity 0.8.30 (regardless of EVM target)
-    --
-    -- We can rewrite the exponentiation into a bit-shift, but we need to be conservative.
-    -- For `256 ^ offset`, which is equal to `2 ^ (8 * offset)` we can rewrite only if `8 * offset < 256`, so `offset < 32`.
-    -- Solidity uses specific pattern for the exponent, where we can verify that this condition is satisfied.
+    -- A special pattern sometimes generated from Solidity that uses exponentiation to simulate bit shift.
+    -- We can rewrite the exponentiation into a bit-shift under certain conditions.
     go (Exp (Lit 0x100) offset@(Mul (Lit a) (Mod _ (Lit b))))
-      | a * b <= 32 && (maxWord256 `Prelude.div` a) > b = (shl (Lit 1) (mul (Lit 0x8) offset))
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -1186,6 +1186,18 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       where l = sort [a, b, c]
             an = EVM.Expr.and
 
+    -- A pattern used by Solidity's legacy codegen when storing value of size < 32 bytes
+    -- into a storage slot (which has size 32 bytes)
+    -- It uses exponentiation to simulate bit shift.
+    -- This is needed on EVMs that do not support left shift (before Constantinopole), but is still part of legacy codegen in Solidity 0.8.30 (regardless of EVM target)
+    --
```
msooseth: I don't think such a long description is necessary. Perhaps just a single line about this being generated by Solidity sometimes instead of bit-shift?

src/EVM/Expr.hs
```diff
@@ -1186,6 +1186,18 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
...
+    -- We can rewrite the exponentiation into a bit-shift, but we need to be conservative.
+    -- For `256 ^ offset`, which is equal to `2 ^ (8 * offset)` we can rewrite only if `8 * offset < 256`, so `offset < 32`.
+    -- Solidity uses specific pattern for the exponent, where we can verify that this condition is satisfied.
+    go (Exp (Lit 0x100) offset@(Mul (Lit a) (Mod _ (Lit b))))
+      | a * b <= 32 && (maxWord256 `Prelude.div` a) > b = (shl (Lit 1) (mul (Lit 0x8) offset))
```
msooseth: I don't understand the second guard think. If a*b <=32, then `maxWord256` divided by the max that `a` can be, which is 32, will always be larger than 32, which is the max that  b could be. In other words, maxWord256 is larger than 32*32. So maybe this check is not needed? Or was am I misunderstanding this? Sorry, it's kinda not clear to me :(

msooseth (commented):
Just some confusion, sorry. Plus a minor ask to make the comment a bit shorter regarding Solidity.

src/EVM/Expr.hs
blishko: The second part of the guard checks that `a * b` does not overflow. I think we need that in order for the rewrite to be sound. You can get `a * b <= 32` with `a` and `b` much larger than `32`, if the product overflows.

src/EVM/Expr.hs
blishko: I wanted to explain the situation because it is very non-obvious rewrite.
I'll try to shorten it.

msooseth: Thanks!

src/EVM/Expr.hs
msooseth: Aaaaaaaa!!! YESSS, I was wondering about the overflow. OK, perfect, thanks!

PR #795, opened 2025-07-15
Make `checkSat` a public function, allow cached props to be Nothing

msooseth:
## Description
This function is needed for Act. The `props` for UNSAT cache now allow the `props` to be Nothing indicating that the cache should not be saved.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
LGMT!

PR #796, opened 2025-07-16
Fix argument order of shift operation in rewrite rule

blishko:
## Description

Apparently, for the SHL expression, the first argument is the number of bits we are shifting and the second argument is the bit-vector we are shifting.
I messed up the order because for me having the bit-vector first seems more natural.

This is a follow up on #794.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
Ooops, thanks!

PR #799, opened 2025-07-16
Tests: Small fixes in equivalence tests

blishko:
To target the tests, we provide the signature of the function we want to check.
Moreover, the type of value in the uint128 test was incorrect. It should have been uint128, not uint64.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -5195,7 +5195,7 @@ tests = testGroup "hevm"
             [i|
               contract C {
                 uint128[5] arr;
-                function set(uint i, uint64 v) external returns (uint) {
+                function set(uint i, uint128 v) external returns (uint) {
                   arr[i] = v;
                   arr[i] = v;
                   return 0;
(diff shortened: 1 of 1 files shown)


msooseth:
Sure, I will merge. I general, I prefer the `Nothing`, because I have made so many typos before, and then things `PASS` but they actually were never tested... have has very bad days of me chasing my tail why something PASS-es, but should FAIL, only to figure out I mistyped the signature :rofl:

blishko:
> Sure, I will merge. I general, I prefer the `Nothing`, because I have made so many typos before, and then things `PASS` but they actually were never tested... have has very bad days of me chasing my tail why something PASS-es, but should FAIL, only to figure out I mistyped the signature 🤣

That's a valid point :) I spent some time figuring out why it is not checking what I wanted it to check, until I found out I cannot have a space between the parameters in the signature.
I made damn sure these tests are really checking what they are supposed to.

I thought about using `functionAbi` from `Solidity.hs`. Or maybe we should have some other strongly typed helper, instead of this error-prone way currently used in tests.

PR #803, opened 2025-07-21
Add cost centers to opcodes, optimize

elopez:
## Description

This PR does the following:
* Adds cost center annotations to each EVM opcode, so that they can be told apart during profiling.
* Optimizes `writeMemory` by taking advantage of `byteStringToVector`
* Switches operations on the stack and program counter to be strict

I'm seeing double digit % improvements on our concrete benchmarks with these changes.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -274,7 +274,7 @@ isCreation = \case
 
 -- | Update program counter
 next :: (?op :: Word8) => EVM t s ()
-next = modifying (#state % #pc) (+ (opSize ?op))
+next = modifying' (#state % #pc) (+ (opSize ?op))
 
 getOpW8 :: forall (t :: VMType) s . FrameState t s -> Word8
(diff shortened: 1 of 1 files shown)


blishko (commented):
Looks good!

I'll test our concrete performance tests today/tomorrow.

What is the benefit of using strict operations on stack and program counter?

blishko (approved):
I am also seeing nice improvements on some of the concrete performance benchmarks.

elopez:
Thanks for testing!

> 
> What is the benefit of using strict operations on stack and program counter?

I'm not fully sure tbh, but the numbers make it seem that possibly suspending the computation of the stack pushes and PC changes is more expensive than just doing them directly.

PR #812, opened 2025-07-29
More complete keccak computation

msooseth:
## Description
This extracts all keccak computations -- previously some were missed. The new test case passes with the new setup, but failed in the old one. Only one of the two keccaks were extracted from:

```haskell
      let buf1 = (Keccak (ConcreteBuf "abc"))
          eq = (Eq buf1 (Lit 0x12))
          buf2 = WriteWord eq (Lit 0x0) mempty
          props = [PEq (Keccak buf2) (Lit 0x123)]
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (approved):
LGTM!

PR #814, opened 2025-08-04
Allow dumping unsolved SMT files

msooseth:
## Description
Allow dumping unsolved SMT files, as requested by @gustavo-grieco 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


src/EVM/Solvers.hs
```diff
@@ -273,6 +278,15 @@ getOneSol smt2@(SMT2 cmds cexvars ps) props r cacheq inst availableInstances fil
     -- put the instance back in the list of available instances
     writeChan availableInstances inst
 
+dumpUnsolved :: SMT2 -> Int -> Maybe FilePath -> IO ()
+dumpUnsolved fullSmt fileCounter dump = do
+   case dump of
+     Just path -> do
+      let filename = "unsolved-" <> show fileCounter
+      putStrLn $ "Dumping unsolved SMT query to: " <> path <> filename
```
blishko: Do we want to alway print this message?
Would "storing" be better than "dumping"?

blishko (commented):
Looks good!
Just a few comments about how to message user about this, but I am also fine with as it is now.

src/EVM/Solvers.hs
msooseth: Ah yeah, should only print this in verbose mode. And yeah, storing is better.

msooseth: OK, removed the printing, thanks!

PR #818, opened 2025-08-05
With `--only-deployed`, only resolve addresses to deployed contracts 

msooseth:
## Description
Note that NOT all addresses deployed are concrete addresses! e.g. `origin` is a symbolic deployed contract.

This code now works when passed `--only-deployed`:
```solidity

contract Stuff {
  function myfun(uint a ) public pure returns (uint) {
    unchecked {
      return a * 2;
    }
  }
}

contract MyOnlyDeployed is Test {
  Stuff stuff;
  function setUp() public {
    stuff = new Stuff();
  }

  // cast keccak "myfun(uint256)"
  // 0xb2d71d6870ba4b4e58f932901330aa9aaf5e31a65a07d4c73632529167626820
  function prove_only_deployed(uint a, address addr) public {
    (, bytes memory ret) = addr.call(abi.encodeWithSelector(0xb2d71d68, a));
    uint x = abi.decode(ret, (uint));
    assert(x == a * 2);
  }
}
```

Without `--only-deployed`:
```
[RUNNING] prove_only_deployed(uint256,address)
   [FAIL] prove_only_deployed
   [WARNING] prove_only_deployed all branches reverted
   [WARNING] hevm was only able to partially explore the test prove_only_deployed due to:
      1x -> Unexpected symbolic arguments to opcode: CALL
```

With `--only-deployed`:
```
cabal run -f devel exe:hevm -- test --root tmp --match ".*only_deployed.*" --only-deployed
Checking 1 function(s) in contract src/contract-only-deployed-contracts.sol:MyOnlyDeployed
[RUNNING] prove_only_deployed(uint256,address)
   [PASS] prove_only_deployed
```

Needs much more love:
- [x] test cases
- [ ] review
- [x] testing
- [x] changelog

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -252,7 +252,7 @@ getSolutions solvers symExprPreSimp numBytes pathconditions = do
   conf <- readConfig
   liftIO $ do
     let symExpr = Expr.concKeccakSimpExpr symExprPreSimp
-    -- when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
+    when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
     ret <- collectSolutions symExpr pathconditions conf
     case ret of
       Nothing -> pure Nothing
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -252,7 +252,7 @@ getSolutions solvers symExprPreSimp numBytes pathconditions = do
   conf <- readConfig
   liftIO $ do
     let symExpr = Expr.concKeccakSimpExpr symExprPreSimp
-    when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
+    -- when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
     ret <- collectSolutions symExpr pathconditions conf
     case ret of
       Nothing -> pure Nothing
--- a/src/EVM/Types.hs
+++ b/src/EVM/Types.hs
@@ -616,7 +616,7 @@ data RunBoth s where
 
 -- | Execution could proceed down one of several branches
 data RunAll s where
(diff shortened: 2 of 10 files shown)


src/EVM/Fetch.hs
```diff
@@ -252,7 +252,7 @@ getSolutions solvers symExprPreSimp numBytes pathconditions = do
   conf <- readConfig
   liftIO $ do
     let symExpr = Expr.concKeccakSimpExpr symExprPreSimp
-    -- when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
+    when conf.debug $ putStrLn $ "Collecting solutions to symbolic query: " <> show symExpr
```
blishko: Do we want to have this enabled now?

src/EVM/Types.hs
```diff
@@ -1605,6 +1605,19 @@ untilFixpoint f a =
...
+forceEAddrToEWord = \case
+  LitAddr a -> Lit (into a)
+  SymAddr a ->  WAddr (SymAddr a)
+  _ -> internalError "Unexpected forceAddr"
+
+forceEWordToEAddr :: Expr 'EWord -> Expr 'EAddr
+forceEWordToEAddr = \case
+  Lit a -> LitAddr (truncateToAddr a)
+  WAddr (SymAddr a) -> SymAddr a
+  _ -> internalError "delegateCall: expected concrete address"
```
blishko: Should there be a different error message?
Now that this is a standalone function, are we sure this will always be called in the context of `delegateCall`?

blishko (commented):
Looks good!
Just a few minor questions.

src/EVM/Fetch.hs
msooseth: Yes, sorry, fixed!

src/EVM/Types.hs
msooseth: Thanks, yeah, I have now updated the error messages for both `forceEWordToEAddr` and `orceEAddrToEWord`.

PR #821, opened 2025-08-06
Better error propagation in reading bytecode from JSON

blishko:
## Description

Previously, some errors encountered when trying to read bytecode from JSON would crash hevm.
Now we propagate the error to the caller.

As part of this change I introduce strong types for possible errors and defer how exactly to report these errors to the caller.
The advantage is that different callers can decide on how to process these errors.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Solidity.hs
+++ b/src/EVM/Solidity.hs
@@ -87,7 +87,7 @@ import System.FilePath.Posix
 import System.Process
 import Text.Read (readMaybe)
 import Witch (unsafeInto)
-import Data.Either.Extra (maybeToEither)
+import Data.Either.Extra (maybeToEither, fromRight)
 
 data StorageItem = StorageItem
   { slotType :: SlotType
@@ -371,23 +371,24 @@ readSolc pt root fp = do
       _ -> "missing " <> (T.unpack field) <> " field"
     explain (MissingOrInvalidJsonField field) = "missing or invalid " <> (T.unpack field) <> " field"
     explain (InvalidSourceMap) = "invalid sourceMap format"
-    explain _ = internalError "TODO"
+    explain (UnlinkedLibrary contract) = "Unlinked libraries detected in bytecode, in " <> (T.unpack contract)
+    explain (OtherError contract errMsg) = "Error when reading bytecode of " <> (T.unpack contract) <> ": " <> T.unpack errMsg
 
-yul :: Text -> Text -> IO (Maybe ByteString)
+yul :: Text -> Text -> IO (Either BytecodeReadingError ByteString)
 yul contractName src = do
   json <- solc Yul src False
   let f = (json ^?! key "contracts") ^?! key (Key.fromText "hevm.sol")
       c = f ^?! key (Key.fromText $ if T.null contractName then "object" else contractName)
       bytecode = c ^?! key "evm" ^?! key "bytecode" ^?! key "object" % _String
-  pure $ (toCode contractName) <$> (Just bytecode)
+  pure $ toCode contractName bytecode
 
-yulRuntime :: Text -> Text -> IO (Maybe ByteString)
+yulRuntime :: Text -> Text -> IO (Either BytecodeReadingError ByteString)
 yulRuntime contractName src = do
   json <- solc Yul src False
   let f = (json ^?! key "contracts") ^?! key (Key.fromText "hevm.sol")
       c = f ^?! key (Key.fromText $ if T.null contractName then "object" else contractName)
       bytecode = c ^?! key "evm" ^?! key "deployedBytecode" ^?! key "object" % _String
-  pure $ (toCode contractName) <$> (Just bytecode)
+  pure $ toCode contractName bytecode
 
 solidity
   :: (MonadUnliftIO m)
@@ -432,8 +433,8 @@ data BytecodeReadingError
   = MissingJsonField Text
   | MissingOrInvalidJsonField Text
   | InvalidSourceMap
-  | UnlinkedLibrary
-  | OtherError
+  | UnlinkedLibrary Text
+  | OtherError Text Text
 
 type BuildArtifacts = (Contracts, Asts, Sources)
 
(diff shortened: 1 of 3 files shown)


blishko:
I would like to update `readStdJSON` similarly next, but I think that can be done separately.

src/EVM/Solidity.hs
```diff
@@ -358,28 +358,37 @@ readSolc pt root fp = do
...
-yul :: Text -> Text -> IO (Maybe ByteString)
+  where
+    explain :: BytecodeReadingError -> String
+    explain (MissingJsonField field) = case field of
+      ast | ast == "ast" || ast == "sources" -> "missing field " <> (T.unpack ast) <> ". Recompile with `forge clean && forge build --ast"
+      _ -> "missing " <> (T.unpack field) <> " field"
+    explain (MissingOrInvalidJsonField field) = "missing or invalid " <> (T.unpack field) <> " field"
+    explain (InvalidSourceMap) = "invalid sourceMap format"
+    explain (UnlinkedLibrary contract) = "Unlinked libraries detected in bytecode, in " <> (T.unpack contract)
+    explain (OtherError contract errMsg) = "Error when reading bytecode of " <> (T.unpack contract) <> ": " <> T.unpack errMsg
```
elopez: could this be extracted (maybe to a Show instance?) so it's usable by other tools?

src/EVM/Solidity.hs
```diff
@@ -358,28 +358,37 @@ readSolc pt root fp = do
   let contractName = T.pack $ takeBaseName fp
   case readJSON pt contractName fileContents of
       Left err -> pure . Left $ "unable to parse " <> show pt <> " project JSON: " <> fp
-        <> " Contract: " <> show contractName <> "\nError: " <> err
+        <> " Contract: " <> show contractName <> "\nError: " <> explain err
       Right (contracts, asts, sources) -> do
         conf <- readConfig
         when (conf.debug) $ liftIO $ putStrLn $ "Parsed contract: " <> show contractName <> " file: " <> fp
         sourceCache <- liftIO $ makeSourceCache root sources asts
         pure (Right (BuildOutput contracts sourceCache))
```
msooseth, line 366:
```suggestion
        pure $ Right $ BuildOutput contracts sourceCache
```

src/EVM/Solidity.hs
```diff
@@ -358,28 +358,37 @@ readSolc pt root fp = do
...
-
-yul :: Text -> Text -> IO (Maybe ByteString)
+  where
+    explain :: BytecodeReadingError -> String
+    explain (MissingJsonField field) = case field of
+      ast | ast == "ast" || ast == "sources" -> "missing field " <> (T.unpack ast) <> ". Recompile with `forge clean && forge build --ast"
+      _ -> "missing " <> (T.unpack field) <> " field"
+    explain (MissingOrInvalidJsonField field) = "missing or invalid " <> (T.unpack field) <> " field"
+    explain (InvalidSourceMap) = "invalid sourceMap format"
+    explain (UnlinkedLibrary contract) = "Unlinked libraries detected in bytecode, in " <> (T.unpack contract)
```
msooseth:
```suggestion
    explain (UnlinkedLibrary contract) = "Unlinked libraries detected in bytecode of contract " <> (T.unpack contract)
```

src/EVM/Solidity.hs
```diff
@@ -358,28 +358,37 @@ readSolc pt root fp = do
...
 
-yulRuntime :: Text -> Text -> IO (Maybe ByteString)
+yulRuntime :: Text -> Text -> IO (Either BytecodeReadingError ByteString)
 yulRuntime contractName src = do
   json <- solc Yul src False
   let f = (json ^?! key "contracts") ^?! key (Key.fromText "hevm.sol")
       c = f ^?! key (Key.fromText $ if T.null contractName then "object" else contractName)
       bytecode = c ^?! key "evm" ^?! key "deployedBytecode" ^?! key "object" % _String
-  pure $ (toCode contractName) <$> (Just bytecode)
+  pure $ toCode contractName bytecode
```
msooseth: Can we unify these with a common function, and then make `yul` into `yulReader "bytecode"` and `yulRuntime` into `yulReader "deployedBytecode"`? (where `yulReader` is probably not a great name, sorry). Or am I trying to generalize too much?

By the way... I don' t understand the hardcoded `hevm.sol` :laughing: Maybe it's a standard?

src/EVM/Solidity.hs
```diff
@@ -420,31 +429,40 @@ functionAbi f = do
 force :: String -> Maybe a -> a
 force s = fromMaybe (internalError s)
 
-readJSON :: ProjectType -> Text -> Text -> Err (Contracts, Asts, Sources)
+data BytecodeReadingError
+  = MissingJsonField Text
+  | MissingOrInvalidJsonField Text
+  | InvalidSourceMap
+  | UnlinkedLibrary Text
+  | OtherError Text Text
```
msooseth, line 428: Super nice!

src/EVM/Solidity.hs
```diff
@@ -420,31 +429,40 @@ functionAbi f = do
...
 readFoundryJSON contractName json = do
-  runtime <- maybeToEither "missing 'deployedBytecode' field" $ json ^? key "deployedBytecode"
-  runtimeCode <- maybeToEither "missing 'deployedBytecode.object' field" $
-    (toCode contractName) . strip0x'' <$> runtime ^? key "object" % _String
+  runtime <- maybeToEither (MissingJsonField "deployedBytecode") $ json ^? key "deployedBytecode"
+  runtimeCode <- (maybeToEither (MissingJsonField "deployedBytecode.object") $ runtime ^? key "object" % _String) >>= (toCode contractName) . strip0x''
   runtimeSrcMap <- case runtime ^? key "sourceMap" % _String of
-    Nothing -> Right $ force "Source map creation error" $ makeSrcMaps ""  -- sourceMap is optional
-    Just smap -> maybeToEither "invalid sourceMap format" $ makeSrcMaps smap
+    Nothing -> pure mempty  -- sourceMap is optional
```
msooseth, line 453: Ah nice!

msooseth (approved):
LGTM

src/EVM/Solidity.hs
blishko: Yes, unifying these two is also on my agenda :)

I actually did it, but wanted to keep it separate from the changes in this PR.

I also don't understand the hardcoded value, I'll dig into that as well. I think that a temporary name was needed at some point :) I don't even know if we create a temporary file or not.

src/EVM/Solidity.hs
blishko: OK, I can do that, provide a default to conversion to error message.
I felt that maybe different users of this functionality may want to customize error messaging.

src/EVM/Solidity.hs
msooseth: OK, let's keep it separate :) As for the `hevm.sol`, let's forget about it, it's not important :)

src/EVM/Solidity.hs
blishko: @elopez , I have added `Show` instance. Can you check if that is sufficient for you?

PR #824, opened 2025-08-09
Added missing assert condition to catch failures in solc < 0.8

gustavo-grieco:
## Description
This condition is needed to catch assertions failures when using solc strictly below 0.8. It is even part of the documentation of checkAssertions but it was missing. 

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
LGTM

PR #825, opened 2025-08-11
Solidity: Refactor reading standard JSON from solc

blishko:
## Description

We refactor the method `readStdJSON` so that it can better propagate errors, similar to other JSON reading functions.

To enable that we introduce a datatype to represent the contract data in the JSON representation. This simplifies JSON parsing significantly, because we can rely on Aeson to do the heavy lifting, for the most part.

Additionally, we change the return type to only `Contracts`, since this is the only part that callers were interested in. We do not try to construct Asts and Sources anymore.

We should probably not export the function at all and keep it internal.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Solidity.hs
+++ b/src/EVM/Solidity.hs
@@ -66,8 +66,7 @@ import Data.Char (isDigit)
 import Data.Foldable
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
-import Data.HashMap.Strict qualified as HMap
-import Data.List (sort, isPrefixOf, isInfixOf, isSuffixOf, elemIndex, tails, findIndex)
+import Data.List (isPrefixOf, isInfixOf, isSuffixOf, elemIndex, tails, findIndex)
 import Data.List.NonEmpty (NonEmpty)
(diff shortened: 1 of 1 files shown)


src/EVM/Solidity.hs
```diff
@@ -482,60 +485,80 @@ readFoundryJSON contractName json = do
...
+            abiMap = mkAbiMap abis,
+            eventMap = mkEventMap abis,
+            errorMap = mkErrorMap abis,
+            storageLayout = mkStorageLayout contractData.storageLayout,
+            immutableReferences = fromMaybe mempty $ do
+              x' <- contractData.evm.deployedBytecode.immutableReferences
+              case fromJSON x' of
+                Success a -> pure a
+                _ -> Nothing
+          }
```
msooseth: Can we make the indentation like:
```
    where
      blah
      blah
```

instead of:

```
    where  blah
           blah
```

blishko: Should be fixed now!

msooseth: Thhanks!

PR #827, opened 2025-08-11
[DRAFT] initial concrete check -- currently wrong

msooseth:
## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM/Stepper.hs
```diff
@@ -22,10 +23,12 @@ where
...
 
 import Control.Monad.IO.Class
-import Control.Monad.Operational (Program, ProgramViewT(..), ProgramView, singleton, view)
+import Control.Monad.Operational (Program, ProgramViewT(..), ProgramView, singleton)
 import Control.Monad.ST (stToIO, RealWorld)
-import Control.Monad.State.Strict (execStateT, runStateT, get)
+import Control.Monad.State.Strict (execStateT, get)
 import Data.Text (Text)
+import Control.Monad.State.Strict (StateT(..))
+import Control.Monad.Operational qualified as Operational
```
blishko: Can we merge these two with the same imports above?

src/EVM/SymExec.hs
```diff
@@ -1117,6 +1119,23 @@ prettyBuf (ConcreteBuf "") = "Empty"
 prettyBuf (ConcreteBuf bs) = formatBinary bs
 prettyBuf b = internalError $ "Unexpected symbolic buffer:\n" <> T.unpack (formatExpr b)
 
+calldataFromCex :: App m => SMTCex -> Expr Buf -> Text -> [AbiType] -> m (Err ByteString)
+calldataFromCex cex buf funName types = do
+  let fullSig = (funName <> "(" <> T.intercalate "," (map abiTypeSolidity types) <> ")")
+      sigKeccak = BS.take 4 $ keccakSig $ encodeUtf8 fullSig
```
blishko: Isn't one of the `take 4` redundant? Aren't you already taking the first 4 bytes already in `keccakSig`?

src/EVM/UnitTest.hs
```diff
@@ -13,11 +13,12 @@ import EVM.FeeSchedule (feeSchedule)
 import EVM.Fetch qualified as Fetch
 import EVM.Format
 import EVM.Solidity
-import EVM.SymExec (defaultVeriOpts, symCalldata, verify, extractCex, prettyCalldata, panicMsg, VeriOpts(..), flattenExpr, groupIssues, groupPartials, IterConfig(..), defaultIterConf)
+import EVM.SymExec (defaultVeriOpts, symCalldata, verify, extractCex, prettyCalldata, calldataFromCex, panicMsg, VeriOpts(..), flattenExpr, groupIssues, groupPartials, IterConfig(..), defaultIterConf, calldataFromCex)
```
blishko: `calldataFromCex` is listed twice.

src/EVM/SymExec.hs
```diff
@@ -1117,6 +1119,23 @@ prettyBuf (ConcreteBuf "") = "Empty"
 prettyBuf (ConcreteBuf bs) = formatBinary bs
 prettyBuf b = internalError $ "Unexpected symbolic buffer:\n" <> T.unpack (formatExpr b)
 
+calldataFromCex :: App m => SMTCex -> Expr Buf -> Text -> [AbiType] -> m (Err ByteString)
+calldataFromCex cex buf funName types = do
+  let fullSig = (funName <> "(" <> T.intercalate "," (map abiTypeSolidity types) <> ")")
```
blishko: We use the function name and types only to construct this signature, Wouldn't it be better to take signature directly?
I was looking at the users of this function, and basically the only user is `symRun`, which constructs the signature anyway.

src/EVM/Stepper.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DataKinds #-}
```
blishko: Why is this needed?

src/EVM/Stepper.hs
msooseth: Good point! Fixed :)

src/EVM/SymExec.hs
msooseth: Haha oh yes. Thanks, fixed!

src/EVM/UnitTest.hs
msooseth: Oops. Yes, thanks, fixed!

src/EVM/SymExec.hs
msooseth: I am not 100% sure I understand what you are saying, I'm sorry. 

We actually don't really get a calldata from the SMT solver, because we try to minimize each part of the calldata. So we get chunks of the calldata (i.e. the values for the types), and that's why I am using this function to re-assemble it. Also, we may want to, in the future, rewrite the Cex in other ways, not just minimize it, so I thought it'd best to re-assemble it from the Cex, rather than try to pass it from the SMT solver directly. I find this somewhat cleaner, plus it helps to essentially fuzz our de- and re-construction of the calldata from the types + values.

Is this what you were referring to? Or is there some place where `symRun` has the calldata that I don't see? I think it's only reconstructing a pretty calldata here:

```haskell
        txtFails <- symFailure opts testName (fst cd) types toPrintData
```

which calls:

```haskell
      mkMsg conf (leaf, cex, repro) = intercalate "\n" $
        ["Counterexample:   " <> reproToText repro
        ,"  calldata: " <> let ?context = dappContext (traceContext leaf)
                           in prettyCalldata cex cd testName types
        ,"  result:   " <> showRes leaf
        ] <> verbText conf leaf
```

But I think that doesn't actually give us a raw calldata? Or am I missing something? Sorry, I am a bit confused.

src/EVM/Stepper.hs
msooseth: Oops, I thought it'd warn me if it's not needed. Sorry, it was needed at one point during the creation of this PR. Removed! Thanks :)

src/EVM/SymExec.hs
gustavo-grieco: My comment related to this code:

It will be great to have two functions:
- A function to transform a CEX into a high-level solidity call (e.g. f(1,2,3)). 
- A function to transform a high-level solidity call into calldata to run the transaction and verify the result.

So the user will be able to see the values in the call, and then debug an invalid CEX if necessary. Echidna already have the code for these two can that can be easily adapted or even re-used here if necessary.

msooseth: Hi,

The `A function to transform a CEX into a high-level solidity call (e.g. f(1,2,3)).` is already available, it's called `prettyCallData`, it's called in `symFailure` like this: `prettyCalldata cex buf funName types`. It returns a text that does what you want, I think!

The `A function to transform a high-level solidity call into calldata to run the transaction and verify the result.` that's what I wrote, I think, i.e. `calldataFromCex cex buf funName types`.

At least as far as I can understand, that's what they do. However, it may be that I am misunderstanding what you mean, or that you need something more specific/etc.

src/EVM/UnitTest.hs
```diff
@@ -47,6 +48,7 @@ import System.IO (hFlush, stdout)
 import Witch (unsafeInto, into)
 import Data.Vector qualified as V
 import Data.Char (ord)
+import Control.Monad.State.Strict (runStateT)
```
blishko: Can you move merge this import with the other `Control.Monad.State.Strict` 24 lines above? If it is really needed...

src/EVM/UnitTest.hs
```diff
@@ -159,6 +161,43 @@ initializeUnitTest opts theContract = do
...
+
+  -- Note, we should not need solvers to execute this code
+  vm2 <- Stepper.interpret (Fetch.oracle utoConc.solvers utoConc.rpcInfo) vm $ do
+    Stepper.evm $ do
+      pushTrace (EntryTrace $ "checking cex for function " <> repCex.testName <> " with calldata: " <> (Text.pack $ bsToHex repCex.callData))
+      makeTxCall utoConc.testParams (ConcreteBuf repCex.callData, [])
+    Stepper.evm get
+
+  (res, (vm3, vmtrace)) <- runStateT (Tracing.interpretWithTrace (Fetch.oracle utoConc.solvers utoConc.rpcInfo) Stepper.execFully) (vm2, [])
+  -- res <- Stepper.interpret (Fetch.oracle utoConc.solvers utoConc.rpcInfo) vm2 Stepper.execFully
```
blishko: Can we remove this comment?

src/EVM/SymExec.hs
blishko: I had a second look and instead of taking `Text` and `[AbyType]`, I think this method should be taking `Sig`.
And we should have a method to convert this to the text representation of a signature. Not sure if `Show` does that for `Sig` or not.
We should not have to repeat code like `(funName <> "(" <> T.intercalate "," (map abiTypeSolidity types) <> ")")`

src/EVM/UnitTest.hs
msooseth: Ah yes, doing that now!

src/EVM/UnitTest.hs
msooseth: Ah yes, sorry, fixed.

msooseth:
@blishko Thanks for the review! I fixed everything, and I am now using `Sig`. I have also created a new helper function, `callSig` in `ABI.hs` that takes a `Sig` and returns a `Text`. This now makes thing cleaner. I hope this is nicer now :)

src/EVM/UnitTest.hs
```diff
@@ -160,6 +161,42 @@ initializeUnitTest opts theContract = do
     Left e -> pushTrace (ErrorTrace e)
     _ -> popTrace
 
+validateCex :: App m
+  => UnitTestOptions RealWorld
+  -> VM Concrete RealWorld
+  -> ReproducibleCex
+  -> m Bool
+validateCex uTestOpts vm repCex = do
+  let utoConc = uTestOpts { testParams = uTestOpts.testParams { caller = LitAddr 0xacab}}
```
blishko: Is `0xacab` some special value? Can we give it a name?

blishko (commented):
I have one final question, then I think we can go ahead with this one!

src/EVM/UnitTest.hs
msooseth: OK, changed to `0xdeadbeef`. It just needs to be a constant.

PR #833, opened 2025-08-14
SymExec: Pre-filter safe branches

blishko:
## Description

Not long ago, we were pre-filtering branches that could not lead to postcondition violation.  This relied on calling simplification and the constraints being simplified to [PBool False].  However, recently (in c34d4d69c), the code changed and it no longer simplified the constraints before sending them further down the pipeline, and thus the statically safe branches were no longer filtered before we started asking about their feasibility.

I believe we should bring back the filtering, if only to have an information about how many branches lead to Failure that we are interested in.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -734,8 +734,12 @@ verifyInputs solvers opts fetcher preState maybepost = do
     getCallPrefix :: Expr Buf -> String
     getCallPrefix (WriteByte (Lit 0) (LitByte a) (WriteByte (Lit 1) (LitByte b) (WriteByte (Lit 2) (LitByte c) (WriteByte (Lit 3) (LitByte d) _)))) = mconcat $ map (printf "%02x") [a,b,c,d]
     getCallPrefix _ = "unknown"
-    toProps leaf vm post = let keccakConstraints = map (\(bs, k)-> PEq (Keccak (ConcreteBuf bs)) (Lit k)) (Set.toList vm.keccakPreImgs)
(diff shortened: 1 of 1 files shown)


blishko:
> This is good!
> 
> I just want to note that the simplification and check before running the SMT solver will take place in `checkSatWithProps` -- it doesn't dispatch it to the SMT solver in case the concKeccak simplification returns a `PBool False`. That also means that the simplifications take place in a thread, concurrently so it's kinda nice. However, we do construct some Prop array and that's unnecessary. Note that in the old code, in [c34d4d6](https://github.com/ethereum/hevm/commit/c34d4d69c837eedcf76c39ae20b617a1508180ff) we actually simplified, too, and checked that for `PBool False`, so it was better than this version. However, this is much faster in many cases, I'm sure. So this is definitely an improvement, however, it's kinda different than the original code, and ultimately the `PBool False` will be filtered out before it hitting the SMT solver.

Indeed, the constraints are simplified in `checkSatWithProps`, so we have never been sending unnecessary queries to the solver.
Prefiltering in advance has the benefit that we have a rough estimate how many of the branches can actually lead to postcondition violation. For example, there is no need to look at the branches that end in `Success`.

> 
> BTW, this could also be:
> 
> ```
> canBeSat (a, _) = not (PBool False `elem` a)
> ```
> 
> Which should be more complete I think?

This is true. But I don't really think our encoding produces `PBool False` anywhere else, so I don't believe this would really cover more case, but it would be more expensive.

Issue #807, opened 2025-07-24
Add yices and bitwuzla with abstraction flag to the solvers available


gustavo-grieco:
Reference: https://github.com/a16z/halmos/blob/dac5018c4043ef392e7c0a1c4f646e6c0be909d1/src/halmos/solvers.py

blishko:
I have a branch for Yices, but I ran into an issue with our use of constant arrays in the SMT encoding: https://github.com/SRI-CSL/yices2/issues/597

For Bitwuzla, the abstraction is turned on by default since `0.8.0` (see https://github.com/bitwuzla/bitwuzla/releases/tag/0.8.0),
so we just have to wait until that becomes available on Nix packages.

gustavo-grieco:
Was bitwuzla 0.8.x tested in any way?

blishko:
Tested in what sense? Performance of hevm with this version bitwuzla? I don't know that we have a good benchmark set for that.
Or do you have something else in mind?

gustavo-grieco:
I saw some cases where bitwuzla 0.8.x was not reporting timeouts or perhaps it was taking more time than expected. I will try to reproduce it..

gustavo-grieco:
Btw, [yices is the default for halmos 0.3.0](https://a16zcrypto.com/posts/article/halmos-v0-3-0-release-highlights/#:~:text=Yices%20is%20the%20new%20default%20solver). Any idea how they deal with the lack of constant arrays?

blishko:
> Btw, [yices is the default for halmos 0.3.0](https://a16zcrypto.com/posts/article/halmos-v0-3-0-release-highlights/#:~:text=Yices%20is%20the%20new%20default%20solver). Any idea how they deal with the lack of constant arrays?

No, I would like to investigate that. Maybe constant arrays do not appear in their encoding. Would be very good to know.

gustavo-grieco:
Hey, any updates on this one?

blishko:
Not yet, not sure how much is this important.
Bitwuzla seems pretty good to me.

gustavo-grieco:
I think there are two main reason to consider Yices:
1. It is the solver that halmos is using and I'm sure there is a good reason for that.
2. Sometimes, what we need is just an alternative solver, even if it not the best all the time. This is to workaround timeouts in bitwuzla.

mpreiner:
[...]

PR #835, opened 2025-08-15
Nix: Update nixpkgs

blishko:
## Description

This updates the solvers to the following versions:
* Bitwuzla 0.8.2
* cvc5 1.3.0
* z3 4.15.2

Bitwuzla 0.8.2 now uses abstraction by default.
On my simplified vat contract, solving was order of magnitude faster
than with Bitwuzla 0.7.0.

Additionally, the new Bitwuzla comes with it's own Windows release, so we can remove our own Bitwuzla build from our CI pipeline.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/contracts/fail/dsProveFail.sol
+++ b/test/contracts/fail/dsProveFail.sol
@@ -26,13 +26,15 @@ contract SolidityTest is Test {
         }
     }
 
-    // This seems to be fast now on SMT solvers, due to built-in heuristics
-    // hence no timeout
-    function prove_smtTimeout(uint x, uint y, uint z) public {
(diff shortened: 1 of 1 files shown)


msooseth:
Oh wow, this looks nice! Should we merge?

blishko:
I would very much like to merge :)

I have updated the description. I think that actually all the solvers get updated.
@gustavo-grieco mentioned some problems with updated Bitwuzla in #807. So far, I have not seen any problem.

blishko:
I actually had to update the example where we test SMT solver timeout, because the new bitwuzla solved the previous one under the 1 second timeout on my laptop.

PR #836, opened 2025-08-25
Expr: Avoid unnecessary allocations in `Expr` simplifications

blishko:
## Description

In this batch we make two modifications:

-  Add a fail-fast cases when reading from abstract buffer. This avoids a more involved case which allocates when it tries to determine if we are reading concrete bytes. We know we are not reading concrete bytes when we are reading from abstract buffer.
- Remove `stripWrites` method. This method was called on `simplifyReads`, but even when it manages to strip writes, it just creates a temporary simplified expression which is passed to `readWord` or `readByte`. Instead, we can apply read-over-write simplifications directly in these methods, without creating temporary expressions.

Overall, this decreases the number of allocations we do in simplification and speeds up analysis of `vat` contract.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -292,6 +292,8 @@ readWord idx b@(WriteWord idx' val buf)
     -- we do not have enough information to statically determine whether or not
     -- the region we want to read overlaps the WriteWord
     _ -> readWordFromBytes idx b
+readWord i@(Lit idx) (WriteByte (Lit idx') _ buf)
+  | idx' < idx || (idx' >= idx + 32 && idx <= (maxBound :: W256) - 32) = readWord i buf
 -- reading a Word that is lower than the dstOffset-32 of a CopySlice, so it's just reading from dst
 readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst) | dstOffset >= x+32 =
   readWord i dst
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -27,7 +27,7 @@ import Data.Vector.Mutable (MVector)
 import Data.Vector.Storable qualified as VS
 import Data.Vector.Storable.ByteString
 import Data.Word (Word8, Word32)
-import Witch (unsafeInto, into, tryFrom, tryInto)
+import Witch (unsafeInto, into, tryInto)
 import Data.Containers.ListUtils (nubOrd)
 
 import Optics.Core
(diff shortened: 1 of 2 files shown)


gustavo-grieco:
How large is the impact of this PR in the `vat` exploration?

msooseth (approved):
LGTM!

We should do more performance evaluation. I have been working with a cluster in my research projects lately more and more and it's great to have each major PR checked to see if we have any perf degradation/improvement.

This should go through -- but we should aim for regular performance checks in the future :)

msooseth:
> How large is the impact of this PR in the `vat` exploration?

Good question, but in general, I think we need to start having some kind of collection of contracts we want to check the performance against, and run large(r) scale experiments, so we have some way of seeing if we are making progress. Also, I want to make sure we sacrifice performance for completeness, if possible. It's a bit annoying when we give up. Also, we need to have some kind of queue for the paths, because the current way of launching a new thread is unsustainable in the long run. With that queue, we could also do path merging more easily.

blishko:
I would love to have regular performance checks. The question is, on what benchmarks? I am currently using the `vat` contract as my benchmark.
Maybe we could try repeating the experiments from the hevm CAV paper?

msooseth:
> I would love to have regular performance checks. The question is, on what benchmarks? I am currently using the `vat` contract as my benchmark. Maybe we could try repeating the experiments from the hevm CAV paper?

We can have a look at them, however, if I remember correctly, they are mostly like the SAT competition "crafted hard benchmarks" -- they are cool, fun, but real-world contracts tend to be quite different. I'll check if any of them look interesting from a real-world perspective. I suspect not many. We should make sure we are _complete_ on these crafted instances, but performance on them is not what I want to optimize for right now. I'd prefer to optimize for real-world contracts. I'll look around for some, and I'll  ask the halmos team, too :)

blishko:
> How large is the impact of this PR in the `vat` exploration?

On my simplified version:

main
```
Using assertion code(s): 0,1,17,18,33,34,49,50,65,81
   Flattening expression
   Exploration finished, 3624 branch(es) to check in call prefix 0xunknown
   Keccak preimages in state: 0
   Checking for reachability of 1237 potential property violation(s) in call prefix 0xunknown

real	0m37.016s
user	1m21.796s
sys	0m3.675s
```

This PR:
```
Using assertion code(s): 0,1,17,18,33,34,49,50,65,81
   Flattening expression
   Exploration finished, 3624 branch(es) to check in call prefix 0xunknown
   Keccak preimages in state: 0
   Checking for reachability of 1237 potential property violation(s) in call prefix 0xunknown

real	0m32.442s
user	1m18.057s
sys	0m2.750s
```

On the full VAT:
main:
```
Using assertion code(s): 0,1,17,18,33,34,49,50,65,81
   Flattening expression
   Exploration finished, 12818 branch(es) to check in call prefix 0xunknown
   Keccak preimages in state: 0
   Checking for reachability of 3029 potential property violation(s) in call prefix 0xunknown

real	9m51.183s
user	26m30.436s
sys	1m29.114s
```

This PR:
```
Using assertion code(s): 0,1,17,18,33,34,49,50,65,81
   Flattening expression
   Exploration finished, 12818 branch(es) to check in call prefix 0xunknown
   Keccak preimages in state: 0
   Checking for reachability of 3029 potential property violation(s) in call prefix 0xunknown

real	8m32.060s
user	23m24.382s
sys	1m6.745s
```

This is only the exploration. It stops right after we flatten the program expression and collect the paths, but before we dispatch them to be checked by SMT.

PR #840, opened 2025-08-29
Revert instead of crash when a missing cheat code is used in concrete mode

gustavo-grieco:
## Description

This small change force a revert in concrete mode instead of crashing (after calling `partial`) if an unknown cheat code is used. It preserves the same behavior if the VM is symbolic

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1794,7 +1794,7 @@ cheat gas (inOffset, inSize) (outOffset, outSize) xs = do
       case Map.lookup abi' cheatActions of
         Nothing -> do
           vm <- get
-          partial $ CheatCodeMissing vm.state.pc abi'
+          caseVMType (partial $ CheatCodeMissing vm.state.pc abi') (finishFrame $ FrameReverted $ ConcreteBuf $ "Unknown cheat code") 
         Just action -> action input
 
 type CheatAction t s = Expr Buf -> EVM t s ()
@@ -3116,6 +3116,8 @@ instance VMOps Symbolic where
           Lit v -> continue $ Just v
           _ -> internalError "runAllPaths can only get concrete values here"
 
+  caseVMType s _ = s
+
 instance VMOps Concrete where
   burn' n continue = do
     available <- use (#state % #gas)
@@ -3244,6 +3246,7 @@ instance VMOps Concrete where
   partial _ = internalError "won't happen during concrete exec"
   branch _ (forceLit -> cond) continue = continue (cond > 0)
   manySolutions _ _ _ = internalError "SMT solver should never be needed in concrete mode"
+  caseVMType _ c = c
 
 -- Create symbolic VM from concrete VM
 symbolify :: VM Concrete s -> VM Symbolic s
--- a/src/EVM/Types.hs
+++ b/src/EVM/Types.hs
@@ -888,6 +888,7 @@ class VMOps (t :: VMType) where
   partial :: PartialExec -> EVM t s ()
   branch :: Maybe Int -> Expr EWord -> (Bool -> EVM t s ()) -> EVM t s ()
   manySolutions :: Maybe Int -> Expr EWord -> Int -> (Maybe W256 -> EVM t s ()) -> EVM t s ()
+  caseVMType :: EVM Symbolic s () -> EVM Concrete s () -> EVM t s ()
 
 -- Bytecode Representations ------------------------------------------------------------------------
 


src/EVM.hs
```diff
@@ -1794,7 +1794,7 @@ cheat gas (inOffset, inSize) (outOffset, outSize) xs = do
       case Map.lookup abi' cheatActions of
         Nothing -> do
           vm <- get
-          partial $ CheatCodeMissing vm.state.pc abi'
+          whenSymbolicElse (partial $ CheatCodeMissing vm.state.pc abi') (vmError $ BadCheatCode "Cannot understand cheatcode." abi')
```
msooseth: Why do we have `CheatCodeMissing` when symbolic but `BadCheatCode` when concrete, even though they seem to be the same issue?

src/EVM.hs
```diff
@@ -1819,7 +1819,7 @@ cheatActions = Map.fromList
                 in query (PleaseDoFFI cmd vm.osEnv cont)
               _ -> vmError (BadCheatCode "ffi(string[]) decoding of string failed" sig)
             _ -> vmError (BadCheatCode "ffi(string[]) parameter decoding failed" sig)
-        else unexpectedSymArg "ffi disabled: run again with --ffi if you want to allow tests to call external scripts" ([] :: [Expr EWord])
+        else vmError $ BadCheatCode "ffi disabled: run again with --ffi if you want to allow tests to call external scripts" sig
```
msooseth, line 1823: Nice!

msooseth (changes requested):
I have a question, marked below regarding `CheatCodeMissing/BadCheatCode`. Also, can you please add a ChangeLog, and also a test case maybe? If adding a testcase is complicated, I'm happy to add a test case, though -- `test.hs` can be a bit annoying.

msooseth:
This is a great fix, BTW! We should definitely not do any kind of "let's die" thing just because a cheatcode is missing

src/EVM.hs
```diff
@@ -1191,7 +1191,7 @@ executePrecompile preCompileAddr gasCap inOffset inSize outOffset outSize xs  =
   input <- readMemory inOffset inSize
   let fees = vm.block.schedule
       cost = costOfPrecompile fees preCompileAddr input
-      notImplemented = internalError $ "precompile at address " <> show preCompileAddr <> " not yet implemented"
+      notImplemented = vmError $ NonexistentPrecompile preCompileAddr
```
msooseth: Hmmm can we do a `whenSymbolicElse (Partial $ CheatCodeMissing ...) (vmError $ NonexistentPrecompile preCompileAddr)` instead?

msooseth:
If we do the `whenSymbolicElse`, I am happy with the symbolic execution, but I will be concerned that we don't set any flag when we do a `vmError` on bad ffi/cheatcode in concrete mode. In a multi-frame setting, the `vmError` may be "eaten" by a frame and the user may think that their cheatcode/ffi got executed, and all went well. So we need a way to be able to indicate to the user that not everything went well, to be able to warn them.

gustavo-grieco:
I'm coming back to this

src/EVM.hs
gustavo-grieco: This is for precompiles so using `ChatCodeMissing` looks odd. Should we add a partial such as `PrecompileMissing`? I don't think we have more than one case for that 🤔

src/EVM.hs
gustavo-grieco: `InvalidOrMissingCheatCode` looks better to you for both cases? (with slightly different names to avoid collisions)

gustavo-grieco:
> In a multi-frame setting, the vmError may be "eaten" by a frame and the user may think that their cheatcode/ffi got executed, and all went well. So we need a way to be able to indicate to the user that not everything went well, to be able to warn them

One option is to append an special log event coming from the cheat precompile itself (e.g. `InvalidCheatcode(uint8)`, so we can print this to the user after the execution is over.

src/EVM.hs
gustavo-grieco: Actually, added just that as a new `Partial`.

src/EVM.hs
msooseth: OK, I see. One of them is a Partial, the other is a vmerror, they cannot be the same.

msooseth:
> > 
> 
> One option is to append an special log event coming from the cheat precompile itself (e.g. `InvalidCheatcode(uint8)`, so we can print this to the user after the execution is over.

Yeah, this sounds reasonable. Do you wanna do that maybe? Otherwise, I'm fine with this.

d-xo:
Probably missing a lot of context here but what's the issue with a partial in a concrete context? Also curious how foundry behaves in this situation?

gustavo-grieco:
> Yeah, this sounds reasonable. Do you wanna do that maybe? Otherwise, I'm fine with this.

I think it will prefer to include that in a future PR: we will also need to show events for reverted transactions and I think the cli is not doing that right now.

> Probably missing a lot of context here but what's the issue with a partial in a concrete context? 

In concrete mode, trying "execute" partials will trigger an `error` crashing the caller.

> Also curious how foundry behaves in this situation?

I think for invalid cheatcodes it will revert without any additional information. In theory, users should be able to quickly debug this using a trace, as it will show that it reverts in that point.

msooseth:
> I think for invalid cheatcodes it will revert without any additional information. In theory, users should be able to quickly debug this using a trace, as it will show that it reverts in that point.

What makes me a bit nervous is that depending on the contract, this may be "eaten" as a silent return value, if the user doesn't look in the trace? But I don't understand how users deal with runs and traces.

d-xo:
> In concrete mode, trying "execute" partials will trigger an error crashing the caller.

This seems worth fixing regardless of what we do here.

> I think for invalid cheatcodes it will revert without any additional information.

Can we please test and confirm this. I think if foundry reverts here then it's probably fine for us to as well. The question of how cheat codes should be treated semantically is a bit vague in general since they are a mix of a normal contract that gets called and a special feature of the vm.

I do share mates concerns around users not noticing a missing cheatcode, especially given that we support a subset of what people might be used to in foundry.

elopez:
> > In concrete mode, trying "execute" partials will trigger an error crashing the caller.
> 
> This seems worth fixing regardless of what we do here.

I think it's intentional; there should be no partials in fully concrete mode. It's hardcoded to be an internalError now: https://github.com/argotorg/hevm/blob/release/0.54.2/src/EVM.hs#L3086

> > I think for invalid cheatcodes it will revert without any additional information.
> 
> Can we please test and confirm this. I think if foundry reverts here then it's probably fine for us to as well. The question of how cheat codes should be treated semantically is a bit vague in general since they are a mix of a normal contract that gets called and a special feature of the vm.
> 
> I do share mates concerns around users not noticing a missing cheatcode, especially given that we support a subset of what people might be used to in foundry.

We used to have a vmError, which would crash echidna but you'd get (vague) feedback about what went wrong ("bad cheat code"). But using partials as it is now results in even more vague errors ("won't happen during concrete exec") when you're doing concrete execution. I personally think something like a vmError that can show more metadata (which signature / address / code location if possible) would strike a good balance about silently failing and going unnoticed vs full crashes that are hard to diagnose.

https://github.com/argotorg/hevm/blob/release/0.54.2/src/EVM.hs#L1705

d-xo:
> I personally think something like a vmError that can show more metadata (which signature / address / code location if possible) would strike a good balance about silently failing and going unnoticed vs full crashes that are hard to diagnose.

Wouldn't we achieve exactly this by replacing the `internalError` in concrete mode `partial` with something like: `partial e = assign #result $ Just (Unfinished e)`? Anyway, I agree with the general sentiment here, and think some kind of specific error with decent metadata is probably a better way to go than just `revert` regardless of the specific implementation.

gustavo-grieco:
For reference, Foundry [adds a revert reason](https://github.com/foundry-rs/foundry/blob/1c4d334e95bc1cad3a390cc735d0f3ec59eea07f/crates/cheatcodes/src/inspector.rs#L593-L610) when no cheatcode is found. This is how it looks like in a trace:
image

I will add a similar message in this PR.

gustavo-grieco:
This is how a hevm trace looks like when there is an invalid cheatcode used:
image
I think it very clear for the user what's going on in that case. There is still the chance that user will do something like this:
```solidity
try hevm.pran(address(0x42424242)) {} catch {}
```
In that case, they will miss the failure, while before, it crashed. In general, they should not do that and they will have the same issue with Foundry.

gustavo-grieco:
All the tests are passing, please take a look to see if this can be merged.

msooseth (approved):
I think this is OK? Do you have any concerns @d-xo ?

src/EVM/Types.hs
```diff
@@ -591,6 +592,7 @@ data PartialExec
   = UnexpectedSymbolicArg { pc :: Int, addr :: Expr EAddr, opcode :: String, msg  :: String, args  :: [SomeExpr] }
   | MaxIterationsReached  { pc :: Int, addr :: Expr EAddr }
   | JumpIntoSymbolicCode  { pc :: Int, addr :: Expr EAddr, jumpDst :: Int }
+  | PrecompileMissing     { preAddr :: Addr }
```
blishko: Should we add `pc` here as well?
Might be good to know where exactly the missing precompile occurred in the code.

src/EVM.hs
```diff
@@ -1793,7 +1793,7 @@ cheat gas (inOffset, inSize) (outOffset, outSize) xs = do
       case Map.lookup abi' cheatActions of
         Nothing -> do
           vm <- get
-          partial $ CheatCodeMissing vm.state.pc vm.state.contract abi'
+          whenSymbolicElse (partial $ CheatCodeMissing vm.state.pc vm.state.contract abi') (vmError $ BadCheatCode "Cannot understand cheatcode." abi')
```
blishko: What was happing here before for concrete execution?

gustavo-grieco: A crash: partial cannot happen in concrete mode

src/EVM/Types.hs
gustavo-grieco: Normally traces will already show that but if you want to add this here we will a way save the PC from the last call

blishko: I am OK with it as is. Just wondering why we have `pc` for all other cases and not here.

msooseth: Wait... Mapping back line of code is not done for this Partial. Shouldn't we do that via `toTxt addr pc`? Just so the user knows where it was called. I'm talking about the `formatPartialDetailed` function

blishko: > Wait... Mapping back line of code is not done for this Partial. Shouldn't we do that via `toTxt addr pc`? Just so the user knows where it was called. I'm talking about the `formatPartialDetailed` function

This was my concern exactly.

msooseth:
Created a PR for this PR: https://github.com/gustavo-grieco/hevm/pull/5

gustavo-grieco:
> Created a PR for this PR: [gustavo-grieco#5](https://github.com/gustavo-grieco/hevm/pull/5)

Already merged!

PR #843, opened 2025-09-01
SymEx: Do not simplify whole program representation

blishko:
Instead of running simplification on the whole program tree, we postpone simplifications and run it when checking feasibility of potential error paths. This avoid unnecessary simplification of a large amount of safe paths.

Note that this changes the return value of `verifyInputs`. It now returns unsimplified program representation instead of the simplified one.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


gustavo-grieco:
I think this approach is closer to halmos, right? Do you see a notable impact in DAI exploration?

blishko:
I don't know what halmos is doing. I am putting reading through halmos code on my todo list :)

What do you mean by DAI?

gustavo-grieco:
I mean the DAI code benchmark you were testing

msooseth (approved):
LGTM, although I am not sure this makes a huge difference, but it probably also doesn't hurt, so it's OK. By the way, does this mean that we won't ever have e.g. this simplification trigger?

```
    go (ITE (Or (Lit x) a) t f)
      | x == 0 = ITE a t f
      | otherwise = t
    go (ITE (Or a b@(Lit _)) t f) = ITE (Or b a) t f
```

Because we don't run the simplification on the un-flattened Expr? Or do we run it on the unflattened Expr at one point?

blishko:
> I mean the DAI code benchmark you were testing

Oh, I see. I only think about it as the "vat contract".

It's a bit difficult to evaluate. Depends what you consider the exploration phase.
Exploring the contract and building its `Expr` representation finishes much faster, but that is because we postpone the bulk of the work to later stage.
Simplification of the expression was taking the larger part of the "exploration".
Now simplification is now triggered later, and it happens concurrently, on each branch separately.
But the benefit is that we never try to simplify obviously safe branches. Which saves significant amount of time.

blishko:
> By the way, does this mean that we won't ever have e.g. this simplification trigger?
> 
> ```
>     go (ITE (Or (Lit x) a) t f)
>       | x == 0 = ITE a t f
>       | otherwise = t
>     go (ITE (Or a b@(Lit _)) t f) = ITE (Or b a) t f
> ```
> 
> Because we don't run the simplification on the un-flattened Expr? Or do we run it on the unflattened Expr at one point?

Hmm, not sure. Probably not on the code path I am using. But I am not sure about others. It could be triggered in the equivalence checking, I guess.

blishko:
On the simplified version of the vat contract I got this measurements for the full run, including SMT solving with 6 bitwuzla instances.

Before:
```
real	1m48.713s
user	2m54.871s
sys	0m28.192s
```

After:
```
real	1m27.715s
user	2m40.599s
sys	0m18.060s
```

I have not measured the full `vat` contract because the memory consumption there seems to be still very high in the SMT phase.

blishko:
I am seeing very high memory usage in the next stage, going into SMT translation, in the full vat contract.
I need to investigate if this is connected to this PR or not.

blishko:
Doesn't seem it is connected to this PR specifically, but there is certainly another memory issue down the line.
I am going to merge this one and investigate further.

PR #845, opened 2025-09-03
SMT: Avoid unnecessary allocations during Expr translation

blishko:
## Description

Previously, `exprToSMT` and `writeBytes` were creating unnecessary temporary structures for processing numbers and bytestrings. However, Haskell libraries offer a way to process these directly.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
Oh, this is really nice!

PR #852, opened 2025-09-05
Fixing double depth incrementing 

msooseth:
## Description
The depth should not have been incremented here. It gets incremented inside runBoth when we actually have to branch. If we can determine that the branch can only go one way, this is not supposed to be incremented. Furthermore, with this, we increment it _twice_ when we branch, and we actually have to go both ways. Notice that `UnknownBranch` does this:

```haskell
      runBothPaths loc exploreDepth UnknownBranch =
        (runBoth depthLimit exploreDepth ) . PleaseRunBoth condSimp $ (runBothPaths loc exploreDepth) . Case
```

where:

```haskell
      Stepper.Fork (PleaseRunBoth cond continue) -> do
        frozen <- liftIO $ stToIO $ freezeVM vm
        let newDepth = vm.exploreDepth+1
        evalLeft <- toIO $ do
          (ra, vma) <- liftIO $ stToIO $ runStateT (continue True) frozen { result = Nothing, exploreDepth = newDepth }
          interpret fetcher iterConf vma (k ra)
        evalRight <- toIO $ do
          (rb, vmb) <- liftIO $ stToIO $ runStateT (continue False) frozen { result = Nothing, exploreDepth = newDepth }
          interpret fetcher iterConf vmb (k rb)
        (a, b) <- liftIO $ concurrently evalLeft evalRight
        pure $ ITE cond a b
```
Hence, it's incremented already.

I have also renamed a parameter name to be more consistent.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
LGTM!

PR #855, opened 2025-09-06
Explicit last case in ForkMany

gustavo-grieco:
## Description
Not sure if this is the right fix, but I was getting nonsensical reports since the `val` in the last case is never used. This is a problem when `onlyDeployed` is true, as it seems that the last address from the list of possible ones is never included in any constraint and therefore, the solver has no way to know it. 🤔 

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -346,7 +346,7 @@ interpret fetcher iterConf vm =
         where
           goITE :: [(Expr EWord, Expr End)] -> Expr End
           goITE [] = internalError "goITE: empty list"
-          goITE [(_, end)] = end
+          goITE [(val, end)] = ITE (Eq expr val) end end
           goITE ((val,end):ps) = ITE (Eq expr val) end (goITE ps)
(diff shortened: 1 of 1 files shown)


msooseth (approved):
Oooooo, this is a REALLY nice bug! Thanks, yesss, this needs to go in. 

Basically my view was: there are KNOWN to be only K solutions to equation E. So once we have used up all the solutions but one (i.e. K-1 solutions), there is only ONE solution that remains. But! I was an idiot, and didn't add the EQ to force that one solutions. Really dumb of me -- that last thread actually does not force the value and you can get nonsensical output. OOops!

msooseth:
I'll add a changelog entry, and merge. This is a serious bug actually :S

msooseth:
OK, so my thought process was:

If the SMT solver says that there are only K solutions, then there are only K solutions, period... but... that's not true. In the SMT solver, we don't add constraints on e.g. address sizes for SymAddr. So the solver in the final SMT query may give us MORE solutions that we thought were possible. I'll think about what we can do here.

msooseth:
I think that #856 may be the right fix here... we'll see...

PR #856, opened 2025-09-08
Adding forcing of value for runAll

msooseth:
## Description
As pointed out by Gustavo, we have actually missed forcing the value of the `runAll`. This lead to incorrect address being generated for e.g. the small contract with `--only-deployed`:
```solidity
pragma solidity ^0.8.19;

contract C {
  function f() public pure returns (uint256) {
    return 42;
  }
}
contract T is Test {
  C c = new C();
  function prove_only_deployed_const(address x) pure public {
    uint256 y = C(x).f();
    assert(y != 42);
  }
}
```
The address should have been that of `c`, but instead, it was a random address. With this fix, the CEX validation system properly validates the CEX, i.e. the address is not random, but that of `c`.

This PR can only be merged on top of https://github.com/argotorg/hevm/pull/827 -- currently failing because this PR is not yet merged

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


test/contracts/pass/only-deployed-force-addr.sol
blishko, line 1: Is this really using `PleaseRunAll`? Isn't there just one possible value?
I just want to make sure we are testing what we want to test.

gustavo-grieco: Yes, there are at least two known address here: the deployed C and T contracts. The origin addresses are also valid ones.

msooseth: Fair point.

With current `main`, running: `cabal run exe:hevm -- test --root tmp --match "only_deployed_force_addr" --only-deployed`  gives us:

```
[RUNNING] prove_only_deployed_force_addr(address)
   [FAIL] prove_only_deployed_force_addr
   Counterexample:   [not reproducible]
     calldata: prove_only_deployed_force_addr(0x0470C202ca30050800087234Ffee81fb7FCEaFc4)
```

This current branch gives:
```
[RUNNING] prove_only_deployed_force_addr(address)
   [FAIL] prove_only_deployed_force_addr
   Counterexample:   [validated]
     calldata: prove_only_deployed_force_addr(0xD95322745865822719164b1fC167930754c248DE)
```

The difference between the two is *only* this PR. The reason for this is that this `runAll`:

```haskell
onlyDeployed addrExpr fallback continue = do
    vm <- get
    if not (?conf.onlyDeployed) then fallback addrExpr
    else case eqT @t @Symbolic of
      Just Refl -> do
        let deployedAddrs = map forceEAddrToEWord $ mapMaybe (codeMustExist vm) $ Map.keys vm.env.contracts
        runAll (?conf.maxDepth) vm.exploreDepth $ PleaseRunAll addrExpr deployedAddrs runAllPaths
      _ -> internalError "Unknown address in Concrete mode"
```

Will run for all `--only-deployed` as long as there are too many solutions. So it will run for this case, too, since the address can be anything here:
```solidity
  function prove_only_deployed_force_addr(address x) pure public {
    uint256 y = C(x).f();
    assert(y != 42);
  }
```

There is no restriction on `x`. So the SMT solver gets 100 different soltuions and then it gives up, and falls back to `onlyDeployed`:
```haskell
forceAddr addrExpr fallback continue = case wordToAddr addrExpr of
  Nothing -> manySolutions (?conf).maxDepth addrExpr 20 $ \case
    Just sol -> continue $ LitAddr (truncateToAddr sol)
    Nothing -> onlyDeployed addrExpr fallback continue
  Just c -> continue c
```

I hope this is a good explanation :) Took me a while to think through this but I hope it's a good description of what's going on!

gustavo-grieco:
Is there any blocker to merge? There is a test in red, but I don't know it is related 🤔

blishko:
The failing test is unrelated. We'll investigate later.

Issue #528, opened 2024-08-17
Confusion about "Latest" BlockNumber when Symbolic executing a contract


ethever:
Confusing about BlockNumber of Latest when Symbolic execution a contract. What happens when i symbolic execute a contract that consuming a lot of time (span lots of blocks), the interpreter fetching the latest state from rpc. But from the point of contract, those state may not be consistent?
https://github.com/ethereum/hevm/blob/eee947de40c62ec7c57a8b7288d7844b82d2415a/src/EVM/Fetch.hs#L70

msooseth:
Sorry, this is a bit unclear to me. Can you maybe clarify a bit? I may be missing something to fully grasp what you mean. Can you maybe give an example?

ethever:
I found in https://docs.alchemy.com/reference/eth-getstorageat that the third params of `eth_getStorageAt` stands for getting latest proposed block when being set to `latest`.
As to blockchain, the block being chain together in remote rpc node: `Block0 --> Block1 --> Block2 --> ... --> BlockN -->...`.
And then the state changing becomes: `State0 --> State1 --> State2 --> ... --> StateN --> ...`.

Now supprots I run `hevm symbolic --address some_live_address_onchain` on BlockN. The hevm will fetche the latest storage at the contract `some_live_address_onchain` at `BlockN` to perform some execution. Unfortunately, however, the contract `some_live_address_onchain` contains some complex logic that will take hevm a significant amount of time to execute, this can take up to 100 blocks' worth of time. Finally, hevm have to perform some `eth_getStorageAt` action to end the execution supports. But this time hevm fetch storage at `BlockN+100`, `BlockN+100` and `BlockN` will probably change the storage of contract `some_live_address_onchain`, causing a non-consistent at the view of execution `some_live_address_onchain` contract.

Yet, I think the semantic of `latest` param at rpc call should be `latest block`, but with block a specified blocknumber being fetched at the start of the execution.

Or what mistake I made?

msooseth:
Oh, I see. Yeah, this is a "time of check time of use" issue kinda, see TOCTOU: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

OK, I'll see what I can do :)

msooseth:
OK, I now 100% understand what's going on. We should update the `Latest` the moment we perform `fetchBlockWithSession`, which is the first thing we do when there is a `latest`. From there on, `Latest` should mean whatever number we fetched then.

[...]

PR #861, opened 2025-09-10
Adding session use for RPC, along with shared cache and fixed Latest Block

msooseth:
## Description
This PR changes the way we do RPC. Basically, we have a mutable cache that is inside Session. We also have a mutable latestBlockNum. The latestBlockNum is updated only ONCE, from Nothing to Just, the first time something with `Latest` is fetched. This way, `Latest` will become a fixed block, and from then on, we use that instead of `Latest`. The cache is such that we keep everything there that we ever fetched.

Cache: due to threads asking for the same data, it is possible that two threads both fetch the same data. However, since the blockchain is immutable, they'd get the same data back, so they'd store the same thing in the cache. So there is no hazard of concurrency, worst case we fetch the same data twice.

The only issue that can arise is that two threads could potentially ask for the Latest, and they get different values back. However, for `test`, the first thing we do normally is to: `rpcDat = Fetch.mkRpcInfo blockUrlInfo mockData` before launching any symbolic execution. So at that point, the `Latest` will be replaced with a value.  For `run` there is no issue, as it's not multi-threaded. For `symbolic`, first thing we do is `    Just url -> liftIO $ Fetch.fetchBlockWithSession conf sess block url` so that's also OK. For equivalence, block number does not matter.

## Old cache in EVM.hs
This as the old cache in EVM.hs:
```haskell
data Cache = Cache
  { fetched :: Map Addr Contract
  , path    :: Map (CodeLocation, Int) Bool
  } deriving (Show, Generic)
```
But that cache was PER VM:
```
data VM (t :: VMType) s = VM
  { result         :: Maybe (VMResult t s)
  , state          :: FrameState t s
  , frames         :: [Frame t s]
[...]
  , cache          :: Cache
```
So each thread would inherit their parent's cache:
* `A -> (B,C)` fork A into B and C. B and C inherit A's cache. Now:
* `B->(D,F)` Another fork. B into D, F. Now, if C fetched some data, D would have no idea, and would re-fetch it. And F would re-fetch it. Actually, also B would re-fetch it. etc. So it's a huge mess, and tons of re-fetches.

Another annoyance was that this "cache" was abused to store the paths visited, which were then used to detect loops:
```haskell
maxIterationsReached vm (Just maxIter) =
  let codelocation = getCodeLocation vm
      (iters, _) = view (at codelocation % non (0, [])) vm.iterations
  in if unsafeInto maxIter <= iters
     then Map.lookup (codelocation, iters - 1) vm.cache.path
```
This had nothing to do with a  cache. It was simply tracking where we have been. This `cache.path` has now been renamed `pathsVisited`.

## Small Changes
* Fetch is also now a module, and not everything is exported out.
* We now also only open one HTTP channel with the RPC host, and use that throughout, as part of `sess :: NetSession.Session` inside `Session`. This reduces latency and pressure on the RPC host. 
* Contracts less than equal to address 0xdeadbeef are treated as if they didn't exist, i.e. the system automatically treats them as if the RPC call finished and returned nonexistent. This should be fine.
* The warning for fetching zero address has been removed. It doesn't make sense, zero address will get fetched when we try to run e.g. a precompile. So instead, we deal with it as per above, i.e. return as if the RPC happened and returned that nothing is deployed there.

## Overall effect
Overall, this PR should make working with RPC a lot faster. It should also make sure we don't overwhelm the RPC host and get data faster. 

Fixes https://github.com/argotorg/hevm/issues/528

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -121,6 +121,8 @@ library
     EVM.Effects,
   other-modules:
     EVM.CheatsTH,
+    EVM.SMT.Types,
+    EVM.SMT.SMTLIB,
     Paths_hevm
   autogen-modules:
     Paths_hevm
@@ -175,7 +177,6 @@ library
     wreq                              >= 0.5.3 && < 0.6,
     regex-tdfa                        >= 1.2.3 && < 1.4,
     base                              >= 4.9 && < 5,
-    here                              >= 1.2.13 && < 1.3,
     smt2-parser                       >= 0.1.0 && < 0.2,
     spool                             >= 0.1 && < 0.2,
     stm                               >= 2.5.0 && < 2.6.0,
@@ -324,6 +325,7 @@ test-suite cli-test
     clitest.hs
   build-depends:
     base,
+    here,
     hevm,
     hspec,
     process,
@@ -331,7 +333,6 @@ test-suite cli-test
     bytestring,
     filepath,
     split,
-    here,
 
 --- Benchmarks ---
 
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -1610,7 +1610,7 @@ onlyDeployed addrExpr fallback continue = do
     else case eqT @t @Symbolic of
       Just Refl -> do
         let deployedAddrs = map forceEAddrToEWord $ mapMaybe (codeMustExist vm) $ Map.keys vm.env.contracts
-        runAll (?conf.maxDepth) vm.exploreDepth $ PleaseRunAll addrExpr deployedAddrs (continue . forceEWordToEAddr)
+        runAll (?conf.maxDepth) vm.exploreDepth $ PleaseRunAll addrExpr deployedAddrs runAllPaths
       _ -> internalError "Unknown address in Concrete mode"
   where
     codeMustExist :: (VM t s) -> Expr EAddr -> Maybe (Expr EAddr)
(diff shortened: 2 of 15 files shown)


hevm.cabal
```diff
@@ -303,6 +304,7 @@ test-suite rpc-tests
     exitcode-stdio-1.0
   main-is:
     rpc.hs
+  build-depends:
```
blishko: ?

src/EVM.hs
```diff
@@ -151,21 +151,21 @@ makeVm o = do
...
       }
-    , forks = Seq.singleton (ForkState env block cache "")
+    , forks = Seq.singleton (ForkState env block mempty "")
     , currentFork = 0
     , labels = mempty
     , osEnv = mempty
     , freshVar = 0
     , exploreDepth = 0
     , keccakPreImgs = fromList []
+    , pathsVisited = mempty
```
blishko, line 168: These are urls, right? Can we try a different name? `paths` is an overloaded term in our setting.

hevm.cabal
msooseth: OOps, thanks, fixed.

src/EVM.hs
msooseth: Actually, no! They are:
```
-- | A unique id for a given pc
type CodeLocation = (Expr EAddr, Int)
type PathsVisited = Map (CodeLocation, Int) Bool
```

So it's a we have been to this code location (addr,+pc number of iterations) -> went to the True/False side.

It's for figuring out if we are in a loop. I have no clue why it was in the cache. I think it was temporarily there? And then it stayed there.

blishko: Oh, OK! Funny.

blishko (approved):
LGTM!

PR #864, opened 2025-09-15
SMT: Do not return unnecessary information from checkSatWithProps

blishko:
It turns out no one was actually using this information anyway and there is no reason to keep the SMT-LIB script around.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Solvers.hs
+++ b/src/EVM/Solvers.hs
@@ -102,18 +102,15 @@ checkMulti (SolverGroup taskq) smt2 multiSol = do
     -- collect result
     readChan resChan
 
-checkSatWithProps :: App m => SolverGroup -> [Prop] -> m (SMTResult, Err SMT2)
+checkSatWithProps :: App m => SolverGroup -> [Prop] -> m (SMTResult)
 checkSatWithProps sg props = do
   conf <- readConfig
   let psSimp = if conf.simp then simplifyProps props else props
-  if psSimp == [PBool False] then pure (Qed, Right mempty)
+  if psSimp == [PBool False] then pure Qed
   else do
     let smt2 = assertProps conf psSimp
-    if isLeft smt2 then
-      let err = getError smt2 in pure (Error err, Left err)
-    else do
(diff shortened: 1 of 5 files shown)


src/EVM/Solvers.hs
```diff
@@ -102,18 +102,17 @@ checkMulti (SolverGroup taskq) smt2 multiSol = do
...
 checkSatWithProps sg props = do
   conf <- readConfig
   let psSimp = if conf.simp then simplifyProps props else props
-  if psSimp == [PBool False] then pure (Qed, Right mempty)
+  if psSimp == [PBool False] then pure Qed
   else do
     let smt2 = assertProps conf psSimp
     if isLeft smt2 then
-      let err = getError smt2 in pure (Error err, Left err)
+      let err = getError smt2 in pure $ Error err
```
msooseth: This could be cleaner, also maybe can be on a single line with the if?

src/EVM/Solvers.hs
```diff
@@ -102,18 +102,17 @@ checkMulti (SolverGroup taskq) smt2 multiSol = do
...
     else do
-      res <- liftIO $ checkSat sg (Just props) smt2
-      pure (res, Right (getNonError smt2))
+      liftIO $ checkSat sg (Just props) smt2
```
msooseth: Maybe on a single line with the else? Also we can then remove the `do`

msooseth (commented):
Wait, wasn't that return value used somewhere? How do we know if something is Partial? I think we used to return some Partial when the SMT couldn't be emitted?

src/EVM/Solvers.hs
blishko: Fixed!

src/EVM/Solvers.hs
blishko: Fixed!

msooseth (approved):
LGTM

PR #865, opened 2025-09-15
SMT: Do not parse term in get-val response

blishko:
We do not need the term, so there is no reason to parse it. The SMT parser we are using (`smt2-parser`) seems to be quite inefficient, so we should limit the amount of work we require it to do. We are always asking for a single value, and we expect it to be a number. We can easily extract the value part of the response and parse only that.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/SMT.hs
+++ b/src/EVM/SMT.hs
@@ -49,7 +49,7 @@ import Data.Text qualified as TS
 import Data.Text.Lazy qualified as T
 import Data.Text.Lazy.Builder
 import qualified Data.Text.Lazy.Builder.Int (decimal)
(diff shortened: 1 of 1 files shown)


msooseth:
Hmmm isn't this more fragile? Also, I don't expect there to be too many Cex-es in the real world, so maybe this is not the most important to improve the speed of? Or maybe I'm wrong or I'm misunderstanding it...

blishko:
It was showing up in the performance profile. Not sure why, but the library was allocating a lot.
The terms we are asking value for are quite large.

It's not fragile unless SMT-LIB standard changes. Which would break the parser from the library as well.
We just need to ignore the term. We only need to parse the value.

PR #867, opened 2025-09-16
SMT: Fix Keccak constraints to use definitions

blishko:
## Description

Previously, the axioms we pass to the solver ("keccak assumptions") would use fully expressed buffers, even though we have applied CSE before and substituted the buffers with GVars in the rest of SMT.

The problem was that we would collect keccak expressions from the unsimplified Props.

The solution here is to collect the Keccak expressions from the simplified Props. However, we have to take care not lose information about concrete Keccak expressions, which are simplified in the first step.
Therefore we change the flow slightly:
1. Collect all concrete Keccak expressions
2. Simplify the props, including concretizing keccaks
3. Introduce definitions for buffers and stores
4. Collect all keccak expressions from simplified props, substituted buffers and stores, and add concrete keccaks from step 1
5. Add necessary axioms for all keccak expressions from step 4

This ensures that the axioms added in step 5 use definitions introduced in step 3.
The effect is that the SMT-LIB script is much smaller because we do not repeat large buffer expressions in every axiom we add (as we are adding quite a lot).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
BTW, I like this re-architecture idea very much. Let me check...

src/EVM/Keccak.hs
```diff
@@ -32,20 +29,18 @@ findKeccakExpr e = mapExprM_ keccakFinder e
...
-  mapM_ findKeccakProp ps
-  mapM_ findKeccakExpr bufs
-  mapM_ findKeccakExpr stores
+findKeccakPropsExprs :: [Prop] -> [Expr Buf] -> [Expr Storage] -> Set (Expr EWord)
+findKeccakPropsExprs ps bufs stores = (execState action (KeccakStore mempty)).keccakExprs
+  where
+    action = do
+      mapM_ findKeccakProp ps
+      mapM_ findKeccakExpr bufs
+      mapM_ findKeccakExpr stores
```
msooseth, line 38: Oh nice!

src/EVM/SMT.hs
```diff
@@ -174,8 +174,11 @@ assertPropsHelper simp psPreConc = do
 
     -- Keccak assertions: concrete values, distance between pairs, injectivity, etc.
     --      This will make sure concrete values of Keccak are asserted, if they can be computed (i.e. can be concretized)
-    keccAssump = keccakAssumptions psPreConc bufVals storeVals
-    keccComp = keccakCompute psPreConc bufVals storeVals
+    concreteKecc = concreteKeccaks psPreConc
+    keccAssump = keccakAssumptions $ Set.toList allKeccaks
+      where
+        allKeccaks = (findKeccakPropsExprs psElim bufVals storeVals) <> Set.map (Keccak . fst) concreteKecc
```
msooseth:
```suggestion
    allKeccaks = (findKeccakPropsExprs psElim bufVals storeVals) <> Set.map (Keccak . fst) concreteKecc
```

msooseth (approved):
LGTM, except for that small change :)

src/EVM/SMT.hs
blishko: OK, I removed the nested `where` clause.

msooseth:
Sorry, I broke the `main` build due to a variable declared but not used, so the tests will fail. #869 will fix main.

PR #875, opened 2025-09-17
ABI: Do not add extra constraints for symbolic arguments

blishko:
## Description

The extra constraints are not necessary because the proper range should be enforced directly in the bytecode.
Moreover, the constraints added by `inRangeSigned` were equivalent to `True` anyway.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


gustavo-grieco:
Are these constraints always enforced in the bytcode or just in the code produced by solc?

blishko:
They should be enforced by solc, because these types are solc constructs.
EVM has no concept of, e.g.,  `int8`.

PR #876, opened 2025-09-17
Expr: Fix maxLitSigned

blishko:
The maximum signed value needs to start with 8 in hex, so that the most significant byte is `1000`. If it starts with `1`, the byte would be `0001`. This would be a positive number.
Moreover, the smallest value should be the largest value plus one. And the largest signed value is
`0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
I'm... sorry. Yes.

PR #877, opened 2025-09-19
Expr: Add missing simplification for read over copyslice

blishko:
## Description

When we have concrete size and destination index, we can determine that reading past the slice is just reading from the destination directly.

Source and source offset can still be symbolic.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -295,8 +295,8 @@ readWord idx b@(WriteWord idx' val buf)
 readWord i@(Lit idx) (WriteByte (Lit idx') _ buf)
   | idx' < idx || (idx' >= idx + 32 && idx <= (maxBound :: W256) - 32) = readWord i buf
 -- reading a Word that is lower than the dstOffset-32 of a CopySlice, so it's just reading from dst
-readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst) | dstOffset >= x+32 =
-  readWord i dst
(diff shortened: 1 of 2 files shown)


src/EVM/Expr.hs
```diff
@@ -295,8 +295,8 @@ readWord idx b@(WriteWord idx' val buf)
 readWord i@(Lit idx) (WriteByte (Lit idx') _ buf)
   | idx' < idx || (idx' >= idx + 32 && idx <= (maxBound :: W256) - 32) = readWord i buf
 -- reading a Word that is lower than the dstOffset-32 of a CopySlice, so it's just reading from dst
-readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst) | dstOffset >= x+32 =
-  readWord i dst
+readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst) | dstOffset >= x+32 = readWord i dst
+readWord i@(Lit x) (CopySlice _ (Lit dstOffset) (Lit size) _ dst) | x >= dstOffset + size = readWord i dst
```
msooseth, line 299: Let's say we read Word (i.e. 32 bytes) starting at byte 10. So `x=10`. There's a CopySlice with dstoffset 8 and size 2. So it will write to bytes 8 and 9. Therefore from byte 10 we can read.

So this is correct :)

msooseth (approved):
Nice catch, actually! Thanks :)

Issue #819, opened 2025-08-06
Another invalid CEX


gustavo-grieco:
> I'm still seeing some invalid CEX. I found an example in a similar code:
> ```solidity
> pragma solidity 0.8.20;
> 
> contract TickTreeSearchEchidnaTest {
>   bool public IS_TEST = true;
>   int24 internal constant MIN_TICK = -887272;
>   int24 internal constant MAX_TICK = -MIN_TICK;
> 
> 
>   using TickTree for mapping(int16 => uint256);
>   uint32 private tickTreeRoot;
> 
>   mapping(int16 => uint256) private tickSecondLayerBitmap;
>   mapping(int16 => uint256) private tickBitmap;
> 
>   int24[] initedTicks;
>   mapping(int24 => uint256) initedTicksIndexes;
> 
>   constructor() public {
>     prove_NextInitializedTickInvariants(421);
>     prove_NextInitializedTickInvariants(5627980);
>     toggleTick(298506);
>   }
> 
>   // returns whether the given tick is initialized
>   function _isInitialized(int24 tick) private view returns (bool) {
>     int16 rowNumber;
>     uint8 bitNumber;
> 
>     assembly {
>       bitNumber := and(tick, 0xFF)
>       rowNumber := sar(8, tick)
>     }
>     uint256 word0 = tickBitmap[rowNumber];
>     unchecked {
>       return (word0 & (1 << bitNumber)) > 0;
>     }
>   }
> 
>   function toggleTick(int24 tick) public {
>     unchecked {
>       if (tick < MIN_TICK) tick = MIN_TICK;
>       if (tick > MAX_TICK) tick = MAX_TICK;
> 
>       assert(tick >= MIN_TICK);
>       assert(tick <= MAX_TICK);
>       bool before = _isInitialized(tick);
>       tickTreeRoot = tickBitmap.toggleTick(tickSecondLayerBitmap, tickTreeRoot, tick);
>       assert(_isInitialized(tick) == !before);
> 
>       if (!before) {
>         initedTicks.push(tick);
>         initedTicksIndexes[tick] = initedTicks.length - 1;
>       } else {
>         uint256 index = initedTicksIndexes[tick];
>         if (index != initedTicks.length - 1) {
>           int24 last = initedTicks[initedTicks.length - 1];
>           initedTicks[index] = last;
>           initedTicksIndexes[last] = index;
>         }
>         initedTicks.pop();
>       }
>     }
>   }
> 
[...]

PR #879, opened 2025-09-22
Removing Cex fuzzer

msooseth:
## Description
This removes the fuzzer of Expr to find a Cex. It does not remove the fuzzing we do in our test suite etc. It only removes the part where we try to find a Cex given an Expr. This is not a great way of doing this work -- Echidna is a much better fuzzer. While this brought something extra to hevm, it also made it harder to maintain, and it was not very well done. Removing.

Related to https://github.com/argotorg/hevm/issues/819 and should fix that issue.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


test/test.hs
```diff
@@ -109,13 +108,13 @@ testNoSimplify a b = let testEnvNoSimp = Env { config = testEnv.config { simp =
   in testCase a $ runEnv testEnvNoSimp b
 
 testFuzz :: TestName -> ReaderT Env IO () -> TestTree
-testFuzz a b = testCase a $ runEnv (testEnv {config = testEnv.config {numCexFuzz = 100, onlyCexFuzz = True}}) b
+testFuzz a b = testCase a $ runEnv (testEnv {config = testEnv.config}) b
```
blishko: Isn't `testEnv {config = testEnv.config}` the same as just `testEnv`?
Also, if these tests are no longer doing fuzzing, should we change the name `testFuzz`?

msooseth: Yes, sorry. Fixed.

msooseth: I renamed it to `test`, and it's correct as-is then. Sorry about this mess.

test/test.hs
```diff
@@ -4340,11 +4337,10 @@ tests = testGroup "hevm"
                         case Map.lookup addr cex.store of
                           Just s -> case (Map.lookup 0 s, Map.lookup (10 + a) s) of
                                       (Just x, Just y) -> x >= y
-                                      (Just x, Nothing) -> x > 0 -- arr1 can be Nothing, it'll then be zero
+                                      (Nothing, Just y) -> y > 0 -- arr1 can be Nothing, it'll then be zero
```
blishko: I actually don't understand this test.

We are looking at the storage of the contract, right?
So, `s` represents the storage which is a map from slots to values, right?
The layout will be that the first 10 slots represent `arr1` and the second 10 slots `arr2`?
So, `Map.lookup 0` should correspond to value of `arr1[0]` and `Map.lookup (10 + a)` corresponds to `arr2[a]`?
Shouldn't this then say `(Nothing, Just y) -> y <= 0`? (which is equivalent to `y == 0`)
Because if `y > 0` then the assert actually holds, no? The counterexample must be such that the  assertion is violated.

msooseth: Yes, exactly, thank you, and sorry for being late. I have fixed it to that. Also,this shows that our cex-fuzzer was indeed wrong so I'm glad it's gone :)

test/test.hs
```diff
@@ -4342,11 +4337,10 @@ tests = testGroup "hevm"
...
```
blishko:
```suggestion
```

This line is actually incorrect.
The first case should always happen.
Currently, it's not the case because we have a [bug](https://github.com/argotorg/hevm/issues/880) in collecting storage reads.


I will fix that tomorrow and then this line should go.

test/test.hs
```diff
@@ -137,8 +133,7 @@ runSubSet p = defaultMain . applyPattern p $ tests
 
 tests :: TestTree
 tests = testGroup "hevm"
-  [ FuzzSymExec.tests
```
blishko, line 140: I think we want to keep these.
If I understand correctly, these are testing the concrete execution of hevm.
It runs on random(?) programs and compares the result to `evmtool`.

test/test.hs
```diff
@@ -4342,11 +4337,9 @@ tests = testGroup "hevm"
                         case Map.lookup addr cex.store of
                           Just s -> case (Map.lookup 0 s, Map.lookup (10 + a) s) of
                                       (Just x, Just y) -> x >= y
-                                      (Just x, Nothing) -> x > 0 -- arr1 can be Nothing, it'll then be zero
                                       _ -> False
                           Nothing -> False -- arr2 must contain an element, or it'll be 0
-          assertBoolM "Did not find expected storage cex" testCex
-          putStrLnM "Expected counterexample found"
+          assertBoolM "Found expected storage cex" testCex
```
blishko:
```suggestion
          assertBoolM "Did not find expected storage cex" testCex
```
This is the error message in case of failure. I think we should preserve what was there before.
Also, I think we can keep the positive message, to be consistent with `storage-cex-1`.
Although, in general, I think we can spam the output less, maybe this PR is not the place to change it.

test/test.hs
msooseth: Yes. OK, I really need to stop working.

msooseth:
OK, I think I finally got it right. If not, please fix it up -- I seem to be making more mistakes than anything else, likely due to being way too tired, haven't slept much :S  I'm sorry.

PR #883, opened 2025-09-24
Fix GHC 9.10 warnings

elopez:
## Description

This fixes warnings that appear when building with GHC 9.10, due to `foldl'` now being available on prelude as well.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -218,7 +218,6 @@ executable hevm
     witch,
     unliftio-core,
     split,
-    wreq,
 
 --- Test Helpers ---
 
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -3,7 +3,7 @@
 
 module EVM where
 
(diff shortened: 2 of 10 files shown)


elopez:
@blishko see https://discourse.haskell.org/t/prelude-foldl-migration-guide/6950

We cannot use just the Prelude one right now as build with older base/GHC combos would break. Once we're comfortable supporting GHC >=9.10 only we can clean up the imports a bit.

msooseth:
LGTM. Do you wanna merge?

elopez:
If you're all ok with it, let's merge these changes so we're ready to make the switch to 9.10 sometime in the future.

msooseth:
Sure, merged! Thanks a lot :)

Issue #886, opened 2025-09-30
Inconsistent decoding of negative values when using solc 0.6.12


gustavo-grieco:
When using solc version 0.6.12 (and posibly others solc 0.6.x versions), the decoding of certain abi value is inconsistent can confuse the users. To reproduce:
```
forge init test
cd test
rm src/* scripts/* test/*
```

Then create a new contract in `src`:

```solidity
// SPDX-License-Identifier: UNLICENSED

contract Bug {
   bool public IS_TEST = true;
   function prove_negative(int24 x) public {
    assert(x != -42);
   }
}
```
Finally add the following line in `foundry.toml`:
```toml
solc_version = "0.6.12"
```

Now we can reproduce this bug. Let's first run using bitwuzla (0.8.2), which shows the correct result:
```
$ forge clean ; forge build --ast ; hevm test --debug --solver bitwuzla
..
Checking 1 function(s) in contract src/Counter.sol:Bug
[RUNNING] prove_negative(int24)
   Exploring call prefix 0xa4e36d39
   Flattening expression
   Exploration finished, 3 branch(es) to check in call prefix 0xa4e36d39
   Keccak preimages in state: 0
   Checking for reachability of 1 potential property violation(s) in call prefix 0xa4e36d39
   SMT result: Cex SMTCex {vars = fromList [(Var "arg1",0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd6)], addrs = fromList [(SymAddr "miner",0x81f777e94fBe791e0d52386E544c24b2d7c9fd01),(SymAddr "origin",0x604c87E034Bb14FDB20370FFd89D7E13f40a5f19)], buffers = fromList [(AbstractBuf "txdata",Comp Base {byte = 0, length = 0x0})], store = fromList [], blockContext = fromList [], txContext = fromList []}
   Found 1 potential counterexample(s) in call prefix 0xa4e36d39
   -> debug of func: prove_negative Failure at the end of expr: UnrecognizedOpcode 254
   -> debug of func: prove_negative Failure at the end of expr: Revert (ConcreteBuf "")
symRun -- (cex,warnings,unexpectedAllRevert): (True,False,False)
Repro running function: LitAddr 0x000000000000000000000000000000000000ACAb with caller: LitAddr 0x00000000000000000000000000000000DeaDBeef, gas: 281474976710655, and calldata: "a4e36d39ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd6"
[...]

PR #887, opened 2025-09-30
More robust calldata decoding, more shrinking

msooseth:
## Description
This PR does two things.

Firstly, it reduces the shrink limit to 32B from 1024B. This means that we now try shrinking even small buffers. This should make sure that trailing useless bytes are not shown to the user, even if the calldata is small. 

Secondly, this PR makes the decoding more robust, by introducing `decodeBufFuzzy` that is a bit more relaxed, allowing trailing bytes at the end of the buffer, and returning more meaningful errors/warnings to the user, which are displayed at the end of the mostly decoded calldata, such as: `     calldata: prove_negative_886(-42) -- with trailing bytes: [0,0,0]`

Fixes https://github.com/argotorg/hevm/issues/886

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


src/EVM/ABI.hs
```diff
@@ -525,6 +526,27 @@ decodeBuf tps buf =
   where
     isDynamic t = abiKind t == Dynamic
 
+decudeBufRobust :: [AbiType] -> Expr Buf -> (AbiVals, String)
```
blishko: Can we fix the existing method instead of introducing new one?
We can adjust the types to better convey the necessary information.

`decodeBuff` can return an `Either`, where `Left` can pass along the error string. `Right` can be `AbiVals` where we remove `NoVals`.
We can extend it with `Maybe ByteString` to return the unprocessed bytes back, if there are any.

I would prefer to leave it up to the caller to decide what to do in such case, and not decide on the message here in the callee.

I was thinking that in our use case, where we want to pretty print the Solidity function call, the extra bytes do not pose any problem and we can safely ignore them, no?

msooseth: `the extra bytes do not pose any problem` -> in this case, they don't seem to. But we also use this method to decode things for cheatcodes, for example, where we cannot ignore it. Either would work there, actually. Let's say that I don't want to warn the user about the trailing bytes (not my preference, but OK). I'd still need to return a Left, because some functions rely on this being exact. Should I return a `Left AbiVals`?

blishko: What do you mean by "relying on this being exact"?
My idea is that the return type would be something like `Either String (AbiVals, Maybe ByteString)`.
The callers then implement their way of handling each situation.
Which could be an error, or success with all bytes consumed, or success with some leftover bytes.

msooseth: Aaaaaah I see! OK, that make sense!

msooseth: I think it should now be how we want it to be? Let me know what you think :)_

msooseth:
Sorry, I would need to think about this more to do it in a different way. I'll think about it, or you can maybe explain how you mean it? Sorry, I  am not up to par for this.

blishko:
Let's merge this then now. We can try something different later and see what we think about it.

PR #891, opened 2025-10-06
Rename confusing function name

msooseth:
## Description
This function name has confused me endlessly. Let's rename it to what it does.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko (commented):
This is nice!
I agree that this better captures what the function is doing.

I think we can go even a step further.
Since `compileWithForge` does not support `CombinedJSON` as `ProjectType` I think we can remove the `ProjectType` parameter altogether.

Then the ``ProjectType` parameter can also be removed from `runForgeTestCustom`!

msooseth:
> I think we can go even a step further. Since `compileWithForge` does not support `CombinedJSON` as `ProjectType` I think we can remove the `ProjectType` parameter altogether.
> 
> Then the ``ProjectType`parameter can also be removed from`runForgeTestCustom`!

Yesss, I did just that I think! Let me know what you think :)

PR #893, opened 2025-10-08
Added rpc cache serialization code

gustavo-grieco:
## Description

This PR contains some little changes in the RPC api and some code to serialize and deserialize the cache. This is essential for tools like Echidna, as it allows to quickly re-run a fuzzing campaign without the need to re-fetching the slots (which gives native speed execution). While there was some code to save individual slots for debugging that approach is too slow to use in a cache, which can be called hundreds of times per second.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -11,10 +11,10 @@ module EVM.Fetch
   , RpcInfo (..)
   , RpcQuery (..)
   , EVM.Fetch.zero
- -- , readMockData
   , BlockNumber (..)
   , mkRpcInfo
   , mkSession
+  , mkSessionWithoutCache
   , Session (..)
   , FetchCache (..)
   , addFetchCache
@@ -350,7 +350,7 @@ fetchSlotWithCache conf sess nPre url addr slot = do
   cache <- readMVar sess.sharedCache
   case Map.lookup (addr, slot) cache.slotCache of
     Just s -> do
-            when (conf.debug) $ putStrLn $ "-> Using cached slot value for slot " <> show slot <> " at " <> show addr
+      when (conf.debug) $ putStrLn $ "-> Using cached slot value for slot " <> show slot <> " at " <> show addr
       pure $ Just s
     Nothing -> do
       when (conf.debug) $ putStrLn $ "-> Fetching slot " <> show slot <> " at " <> show addr
@@ -448,10 +448,13 @@ mkSession cacheDir mblock = do
   latestBlockNum <- liftIO $ newMVar Nothing
   pure $ Session sess latestBlockNum cache cacheDir
 
+mkSessionWithoutCache :: App m => m Session
+mkSessionWithoutCache = mkSession Nothing Nothing
+
 -- Only used for testing (test.hs, BlockchainTests.hs)
 zero :: Natural -> Maybe Natural -> Fetcher t m s
 zero smtjobs smttimeout q = do
-  sess <- mkSession Nothing Nothing
+  sess <- mkSessionWithoutCache
   withSolvers Z3 smtjobs 1 smttimeout $ \s ->
     oracle s (Just sess) mempty q
 
--- a/test/EVM/Test/Utils.hs
+++ b/test/EVM/Test/Utils.hs
@@ -52,7 +52,7 @@ runForgeTest testFile match = runForgeTestCustom testFile match Nothing Nothing 
 testOpts :: forall m . App m => SolverGroup -> FilePath -> Maybe BuildOutput -> Text -> Maybe Integer -> Bool -> RpcInfo -> m (UnitTestOptions RealWorld)
 testOpts solvers root buildOutput match maxIter allowFFI rpcinfo = do
   let srcInfo = maybe emptyDapp (dappInfo root) buildOutput
-  sess <- Fetch.mkSession Nothing Nothing
+  sess <- Fetch.mkSessionWithoutCache
   params <- paramsFromRpc rpcinfo sess
 
   pure UnitTestOptions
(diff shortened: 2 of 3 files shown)


msooseth:
Ah, this looks really interesting actually, thanks! This can replace my very simple JSON writing/reading, right? If so, can we replace it entirely with this? This seems more systematic and more serious :)

gustavo-grieco:
In theory, this new code can replace the old serialization, however, there is a catch: this serialization only happens at the startup (to restore into the cache) and when the exploration and solving already finished (to save the cache). This means that in order to debug a case, you need to wait until the end. With the current code, just enabling the `debug` flag you will dump the RPC responses as soon as they are requested.

msooseth:
> In theory, this new code can replace the old serialization, however, there is a catch: this serialization only happens at the startup (to restore into the cache) and when the exploration and solving already finished (to save the cache). This means that in order to debug a case, you need to wait until the end. With the current code, just enabling the `debug` flag you will dump the RPC responses as soon as they are requested.

That's actually okay! and then we could remove the RPCContract and other stuff, so we could be lean and mean :) I only wrote that code so we can replay a full, running, example. So it would be a perfect, in fact, better, replacement! Could you maybe adjust the PR to do that? It'd be awesome, because it'd be less code and actually more powerful. Would be really nice. Otherwise, I can also do this change if you like, as a PR to your PR :) Let me know what you think!

gustavo-grieco:
@msooseth the old code was replaced by the cache save/load. I also changed the old cli RPC test with the new one, using a cache file for the particular block used and it seems to be working.

hevm.cabal
```diff
@@ -143,6 +143,7 @@ library
   install-includes:
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
+    directory,
```
blishko: I think we need to add version bounds, no?

src/EVM/Fetch.hs
```diff
@@ -374,26 +401,58 @@ fetchSlotFrom sess nPre url addr slot = do
...
+      pure emptyCache
+
+saveCache :: FilePath -> W256 -> FetchCache -> IO ()
+saveCache dir n cache = do
+  createDirectoryIfMissing True dir
+  let fp = dir </> cacheFileName n
+  putStrLn $ "Saving RPC cache to " ++ fp
+  BSL.writeFile fp (encodePretty cache)
+
+mkSession :: App m => Maybe FilePath -> Maybe W256 -> m Session
```
blishko, line 441: Should we add a helper like this?
```
mkFreshSession :: App m => m Session
mkFreshSession = mkSession Nothing Nothing
```

Seems to me we are repeating `mkSession Nothing Nothing` quite a few times, so it seems a helper is warranted.

hevm.cabal
msooseth:
```suggestion
    directory                         >=1.3.8.0 && <= 1.3.9.0,
```

src/EVM/Fetch.hs
```diff
@@ -338,14 +344,38 @@ makeContractFromRPC (RPCContract code nonce balance) =
...
+fetchSlotWithCache :: Config -> Session -> BlockNumber -> Text -> Addr -> W256 -> IO (Maybe W256)
+fetchSlotWithCache conf sess nPre url addr slot = do
+  n <- getLatestBlockNum conf sess nPre url
+  cache <- readMVar sess.sharedCache
+  case Map.lookup (addr, slot) cache.slotCache of
+    Just s -> do
+      putStrLn $ "-> Using cached slot value for slot " <> show slot <> " at " <> show addr
+      pure $ Just s
+    Nothing -> do
+      putStrLn $ "-> Fetching slot " <> show slot <> " at " <> show addr
```
msooseth:
```suggestion
      when (conf.debug) $ putStrLn $ "-> Fetching slot " <> show slot <> " at " <> show addr
```

src/EVM/Fetch.hs
```diff
@@ -338,14 +344,38 @@ makeContractFromRPC (RPCContract code nonce balance) =
       & set #balance  (Lit balance)
       & set #external True
 
+fetchSlotWithCache :: Config -> Session -> BlockNumber -> Text -> Addr -> W256 -> IO (Maybe W256)
+fetchSlotWithCache conf sess nPre url addr slot = do
+  n <- getLatestBlockNum conf sess nPre url
+  cache <- readMVar sess.sharedCache
+  case Map.lookup (addr, slot) cache.slotCache of
+    Just s -> do
+      putStrLn $ "-> Using cached slot value for slot " <> show slot <> " at " <> show addr
```
msooseth:
```suggestion
            when (conf.debug) $ putStrLn $ "-> Using cached slot value for slot " <> show slot <> " at " <> show addr
```

src/EVM/Fetch.hs
```diff
@@ -374,26 +401,58 @@ fetchSlotFrom sess nPre url addr slot = do
...
+  putStrLn $ "Saving RPC cache to " ++ fp
+  BSL.writeFile fp (encodePretty cache)
+
+mkSession :: App m => Maybe FilePath -> Maybe W256 -> m Session
+mkSession cacheDir mblock = do
   sess <- liftIO NetSession.newAPISession
-  let emptyCache = FetchCache Map.empty Map.empty Map.empty
-  cache <- liftIO $ newMVar emptyCache
+  initialCache <-
+    case (cacheDir, mblock) of
```
msooseth:
```suggestion
  initialCache <- case (cacheDir, mblock) of
```

src/EVM/Fetch.hs
gustavo-grieco: Done!

gustavo-grieco:
@msooseth Already added the line in the changelog and addressed all the other comments.

PR #894, opened 2025-10-11
Fix int decoding when upper bits are not set

gustavo-grieco:
## Description

Fix for the missing case of #886 

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


blishko:
I believe this fix is correct.
I took the liberty to push small changes directly to this branch.
I named the function to make it clear we are sign-extending the number (in case the size of the type is <256 bits)
For completeness I truncate the value to the correct number of bits first.
I also added a simple test case which would previously fail, but it works with this fix.

blishko:
@gustavo-grieco , can you test if this still fixes the issue you described in #886?

gustavo-grieco:
Just tested the recent changes and all the cases are correctly decoded.

msooseth:
Let me push a changelog line :)

PR #900, opened 2025-10-15
SymEx: Always require postcondition in verify* methods

blishko:
## Description

To make the core method simpler we always require postcondition to be given.
If there is no postcondition, we are not really "verifying". However, there is a use case where we don't want to verify anything, we just want to explore a contract.
For such use case, we provide a new helper method to do exactly that. `exploreContract` executes the contract symbolically, and returns its Expr representation.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -459,21 +459,7 @@ checkAssert
   -> VeriOpts
   -> m (Expr End, [VerifyResult])
 checkAssert solvers errs c signature' concreteArgs opts = do
-  checkAssertWithSession solvers Nothing errs c signature' concreteArgs opts
-
--- Used only in testing
-checkAssertWithSession
-  :: App m
-  => SolverGroup
-  -> Maybe Fetch.Session
-  -> [Word256]
-  -> ByteString
-  -> Maybe Sig
-  -> [String]
-  -> VeriOpts
-  -> m (Expr End, [VerifyResult])
-checkAssertWithSession solvers sess errs c signature' concreteArgs opts = do
-  verifyContractWithSession solvers sess c signature' concreteArgs opts Nothing (Just $ checkAssertions errs)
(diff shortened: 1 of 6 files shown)


src/EVM/SymExec.hs
```diff
@@ -568,27 +554,29 @@ verifyContract :: forall m . App m
...
-verifyContractWithSession solvers sess theCode signature' concreteArgs opts maybepre maybepost = do
+  -> m (Expr End)
+exploreContract solvers theCode signature' concreteArgs opts maybepre = do
   calldata <- mkCalldata signature' concreteArgs
   preState <- liftIO $ stToIO $ abstractVM calldata theCode maybepre False
-  let fetcher = Fetch.oracle solvers sess opts.rpcInfo
-  verify solvers fetcher opts preState maybepost
+  let fetcher = Fetch.oracle solvers Nothing opts.rpcInfo
+  executeVM fetcher opts.iterConf preState
+
```
msooseth, line 579: this is only used in one test. It will also need to be removed when we return an `[Expr  End]`, rather than an `Expr End` with a bunch of ITE's in it. So we can also remove it now, if you like, along with the test.

msooseth (approved):
LGTM! I added a comment, more as a reminder than anything else :)

src/EVM/SymExec.hs
blishko: We can keep it for now and remove/adjust later.

blishko:
I guess I should add a changelog entry, since we are changing the signature of a function exposed to the outside world.

msooseth:
> I guess I should add a changelog entry, since we are changing the signature of a function exposed to the outside world.

I think we should _always_ add a changelog entry unless it's a trivial PR :) But yeah. And then we can merge :)

PR #901, opened 2025-10-15
Test: Ensure simplifications are tested properly

blishko:
When checking equivalence of original and simplified expressions, we do not want to apply simplifications during translation to SMT. This would make the check trivial.

Changing the test configuration "test -> testNoSimplify" is necessary only when the test is making an equivalence check, but we change all the tests in the SimplifierUnitTests group for consistency.

**** UPDATE ****
It turns out it is much better to ensure directly in the equivalence check function that the simplifications are turned off. Thanks @msooseth!


blishko:
Yes, that would actually be much better!
I'll try to do that instead!

PR #903, opened 2025-10-16
EVM: Split symbolic and concrete sstore handling

blishko:
## Description

Previously, there was a single piece of code handling both variants. However, this solution unnecessary complicates the code and comes with performance hits.

In symbolic case, we do not model gas, so all the code related to computation and refunding of gas is not needed.
This code was trying to perform expression simplifications, which we now avoid altogether.

In concrete case, we know we have to have concrete values everywhere, so we can directly force expresseions to concrete values and avoid unnecessary case splitting.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

PR #904, opened 2025-10-20
EVM: Separate symbolic and concrete handling of SLOAD

blishko:
## Description

This is a follow-up for #903, but this time for SLOAD opcode.
We separate the handling of symbolic and concrete case.
Symbolic case does not case about gas and concrete case can be simpler because it can focus on handling only concrete values.
Moreover, we can specialize `accessStorageForGas` for concrete case, because now it is used only in concrete mode.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -736,7 +736,7 @@ exec1 conf = do
                   else finalizeLoad $ Expr.readStorage' (Expr.concKeccakOnePass x) this.storage
 
                 concreteRead :: EVM t s () = do
-                  acc <- accessStorageForGas self x
+                  acc <- accessStorageForGas self (forceLit x)
                   let cost = if acc then g_warm_storage_read else g_cold_sload
                   burn cost $ if this.external
                     then accessStorage self x finalizeLoad
@@ -767,7 +767,7 @@ exec1 conf = do
                           | (currentVal == originalVal) = g_sreset
                           | otherwise = g_sload
 
-                    acc <- accessStorageForGas self x
+                    acc <- accessStorageForGas self slot
                     let cold_storage_cost = if acc then 0 else g_cold_sload
                     burn (storage_cost + cold_storage_cost) $ do
                       updateVMState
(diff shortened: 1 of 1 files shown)


src/EVM.hs
```diff
@@ -727,13 +727,22 @@ exec1 conf = do
...
+
+                concreteRead :: EVM t s () = do
+                  acc <- accessStorageForGas self (forceLit x)
+                  let cost = if acc then g_warm_storage_read else g_cold_sload
+                  burn cost $ do
+                    let val = accessConcreteStorage this.storage (forceLit x)
+                    next
+                    assign' (#state % #stack) ((Lit val):xs)
+              in
+                whenSymbolicElse symbolicRead concreteRead
```
msooseth:
```suggestion
              in whenSymbolicElse symbolicRead concreteRead
```

I prefer to have as few lines as possible nowadays. I hope that's OK?

blishko:
I had to keep the old behaviour in case the current contract is `external`.
It means it was fetched by RPC and we might not have its storage represented accurately.
When loading a slot, we need to check if we don't need to fetch it with another RPC call.

src/EVM.hs
```diff
@@ -727,13 +727,23 @@ exec1 conf = do
...
-                  next
-                  assign' (#state % #stack) (y:xs)
+            x:xs ->
+              let
+                finalizeLoad readValue = do next; assign' (#state % #stack) (readValue:xs)
+
+                symbolicRead :: EVM t s () =
+                  case this.external of
+                    True -> accessStorage self x finalizeLoad
+                    False -> finalizeLoad $ Expr.readStorage' (Expr.concKeccakOnePass x) this.storage
```
msooseth: Can we do 
```haskell
symbolicRead :: EVM t s () = if this.external
  then ...
  else ...
```

It'd be compact and clean, no?
here? It'd be one line less

src/EVM.hs
```diff
@@ -727,13 +727,23 @@ exec1 conf = do
...
+
+                concreteRead :: EVM t s () = do
+                  acc <- accessStorageForGas self (forceLit x)
+                  let cost = if acc then g_warm_storage_read else g_cold_sload
+                  burn cost $ do
+                    case this.external of
+                      True -> accessStorage self x finalizeLoad
+                      False -> finalizeLoad $ Lit $ accessConcreteStorage this.storage (forceLit x)
```
msooseth: Same here as above maybe?

msooseth (approved):
Nice! Maybe you could compact it a bit as per my comment?

src/EVM.hs
blishko: Fixed!

src/EVM.hs
blishko: Fixed!

PR #906, opened 2025-10-21
Symbolic execution queue

msooseth:
## Description
A queue for symbolic execution threads. Currently the number of processors, but should be easy to adjust. Returns `[Expr End]`, and now the `ITE` is gone. When asking for reachability, we return a full set of leafs (i.e. `Expr End`-s). All tests pass. 

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -3074,8 +3074,9 @@ instance VMOps Symbolic where
     where
       runBothPaths loc _ (Case v) = do
         assign #result Nothing
-        pushTo #constraints $ if v then PNeg (PEq (Lit 0) cond)
-                                   else PEq (Lit 0) cond
+        let condSimp = Expr.simplify cond
+        pushTo #constraints $ if v then Expr.simplifyProp (Lit 0 ./= condSimp)
+                                   else Expr.simplifyProp (Lit 0 .== condSimp)
         (iteration, _) <- use (#iterations % at loc % non (0,[]))
         stack <- use (#state % #stack)
         assign (#pathsVisited % at (loc, iteration)) (Just v)


src/EVM.hs
```diff
@@ -3092,7 +3090,7 @@ instance VMOps Symbolic where
         assign (#iterations % at loc) (Just (iteration + 1, stack))
         continue v
       -- Both paths are possible; we ask for more input
-      runBothPaths loc exploreDepth UnknownBranch =
+      runBothPaths loc exploreDepth UnknownBranch = do
```
blishko: Why is `do` needed here?

src/EVM/SymExec.hs
```diff
@@ -317,17 +325,113 @@ freezeVM vm = do
...
+  -- Decrement and check if all done
+  liftIO $ atomically $ do
+    n <- readTVar numTasks
+    let n' = n - 1
+    writeTVar numTasks n'
+    when (n' == 0) $ putTMVar allDone ()
+
+-- | Symbolic interpreter that explores all paths. Returns an
+-- 'Expr End' representing the possible executions. This Expr End is NOT flattened,
+--  i.e. it (likely) contains ITE-s. The only End-s possible are: Partial, Failure, Success, ITE
```
blishko: This comment is not accurate, no? There are no ITEs anymore.

src/EVM.hs
msooseth: Oops, not needed, removed. Sorry.

src/EVM/SymExec.hs
msooseth: Good point! I have just pushed a commit that addresses this and another one in the documentation.

doc/src/architecture.md
```diff
@@ -68,16 +68,13 @@ Expressions in this language can have the following types:
...
-where each leaf is an end state. Some end states (e.g. `Return`) contain copies
-of any externally observable data (i.e. returndata and post call storage).
-
-As an example the following Expr encodes a program that branches based on the
-equality of two symbolic words ("a" and "b"), and returns if they are equal and
-reverts if they are not:
+An EVM program is represented by an `[Expr End]`, which is a list of all
+possible end states for a program without branches. As an example the following
+two `Expr End`-s encode a program that branches based on the equality of two symbolic
+words ("a" and "b"), and returns if they are equal and reverts if they are not:
```
blishko, line 74:
```suggestion
An EVM program is represented by an `[Expr End]`, which is a list of all
possible end states for a program. As an example the following
two `Expr End`-s encode a program that branches based on the equality of two symbolic
words ("a" and "b"), and returns if they are equal and reverts if they are not:
```

test/EVM/Test/FuzzSymExec.hs
```diff
@@ -760,38 +761,39 @@ checkTraceAndOutputs contract gasLimit txData = do
...
+             if null simplConcrExprRetval then putStrLn ", but it was a Nothing, so not strong equivalence"
+                                          else putStrLn ""
            else do
              putStrLn $ "original expr                    : " <> (show expr)
              putStrLn $ "concretized expr                 : " <> (show concretizedExpr)
              putStrLn $ "simplified concretized expr      : " <> (show simplConcExpr)
              putStrLn $ "evmtoolTraceOutput.output.output : " <> (show (evmtoolTraceOutput.output.output))
              putStrLn $ "HEVM trace result output         : " <> (bsToHex (bssToBs hevmTraceResult.out))
-             putStrLn $ "ret value computed via symb+conc : " <> (bsToHex (fromJust simplConcrExprRetval))
+             putStrLn $ "ret value computed via symb+conc : " <> (intercalate "," $ map bsToHex simplConcrExprRetval)
```
blishko: Seems unnecessary to use `intercalate` here because you know `simplConcrExprRetval` has exactly one element.

I think something like this should work?
```suggestion
             putStrLn $ "ret value computed via symb+conc : " <> (bsToHex (simplConcrExprRetval !! 0))
```

test/test.hs
```diff
@@ -206,7 +206,7 @@ tests = testGroup "hevm"
           }
         }
         |]
-       let veriOpts = defaultVeriOpts { iterConf = defaultIterConf { maxIter = Just 5 }}
+       let veriOpts = (defaultVeriOpts :: VeriOpts) { iterConf = defaultIterConf { maxIter = Just 5 }}
```
blishko, line 209: Why do you need to specify the type explicitly now?

test/test.hs
```diff
@@ -1884,8 +1860,9 @@ tests = testGroup "hevm"
           let calldata = (WriteWord (Lit 0x0) (Var "u") (ConcreteBuf ""), [])
           initVM <- liftIO $ stToIO $ abstractVM calldata initCode Nothing True
           let iterConf = IterConfig {maxIter=Nothing, askSmtIters=1, loopHeuristic=StackBased }
-          expr <- Expr.simplify <$> interpret (Fetch.oracle s Nothing mempty) iterConf initVM runExpr
-          assertBoolM "unexptected partial execution" (not $ Expr.containsNode isPartial expr)
+          expr <- interpret (Fetch.oracle s Nothing mempty) iterConf initVM runExpr
```
blishko: I think the name of the variable should be updated.
The result of interpret is no longer an `expr`.
It is a list of paths/endpoints.

It should be at least `exprs` to indicate it is a list not a single value.
Though I would prefer `paths`.
This comment applies to multiple places in this file.

src/EVM/SymExec.hs
```diff
@@ -1,26 +1,27 @@
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
 
 module EVM.SymExec where
 
 import Prelude hiding (Foldable(..))
 
+import GHC.Natural
```
blishko: This should be moved to the end of external dependencies,  to keep alphabetical order.
Same for `GHC.Conc`

src/EVM/SymExec.hs
```diff
@@ -34,6 +35,13 @@ import Data.Tuple (swap)
 import Data.Vector qualified as V
 import Data.Vector.Storable qualified as VS
 import Data.Vector.Storable.ByteString (vectorToByteString)
+import Control.Monad.State.Strict
+
+import Control.Concurrent.Chan (Chan, newChan, writeChan, readChan)
+import Control.Concurrent.STM (writeTChan, newTChan, TChan, readTChan, atomically, isEmptyTChan)
+import Control.Concurrent (forkIO, killThread)
+import Control.Concurrent.STM.TVar (TVar, newTVarIO, modifyTVar, readTVar, writeTVar, readTVarIO)
+import Control.Concurrent.STM.TMVar (putTMVar, takeTMVar, TMVar, putTMVar, takeTMVar, newEmptyTMVarIO)
```
blishko: This should be moved before `Data`, to preserve alphabetical order.

src/EVM/SymExec.hs
```diff
@@ -336,41 +434,59 @@ interpret fetcher iterConf vm =
...
-          goITE [] = internalError "goITE: empty list"
-          goITE [(_, end)] = end
-          goITE ((val,end):ps) = ITE (Eq expr val) end (goITE ps)
-          runOne :: App m => VM 'Symbolic RealWorld -> Int -> Expr EWord -> m (Expr 'End)
-          runOne frozen newDepth v = do
+          runOne :: App m => VM 'Symbolic RealWorld -> Int -> [Expr EWord] -> m (Expr End)
+          runOne frozen newDepth [v] = do
+            conf <- readConfig
+            (ra, vma) <- liftIO $ stToIO $ runStateT (continue v) frozen { result = Nothing, exploreDepth = newDepth }
+            let vmConst  = vma { constraints = (PEq v expr) : vma.constraints }
```
blishko: The first three lines are the same in both cases. Maybe we can split the cases only after the common point?

Also,  we now add a constraint to the `constraints` of the VM, but we did were not doing that before.
I actually think this is correct!
Seems like previously we might have been missing some information?

src/EVM/SymExec.hs
```diff
@@ -336,41 +434,59 @@ interpret fetcher iterConf vm =
...
+
+        (ra, vma) <- liftIO $ stToIO $ runStateT (continue True) frozen { result = Nothing, exploreDepth = newDepth }
+        let vmaConst  = vma { constraints = PNeg (PEq (Lit 0) cond) : vma.constraints }
+        let newT = (t :: InterpTask m) { vm = vmaConst, stepper = (k ra) }
+        liftIO $ atomically $ modifyTVar numTasks (+1)
+        liftIO $ writeChan taskq newT
+        when (conf.debug) $ liftIO $ putStrLn $ "Queued new task for Fork at depth " <> show newDepth
+
+        (rb, vmb) <- liftIO $ stToIO $ runStateT (continue False) frozen { result = Nothing, exploreDepth = newDepth }
+        let vmbConst  = vmb { constraints = PEq (Lit 0) cond : vmb.constraints }
```
blishko: Same question about the constraints here.

src/EVM/SymExec.hs
```diff
@@ -336,41 +434,59 @@ interpret fetcher iterConf vm =
...
+        let newT2 = (t :: InterpTask m) { vm = vmbConst, stepper = (k rb) }
+        when (conf.debug) $ liftIO $ putStrLn $ "Continuing task for Fork at depth " <> show newDepth
+        interpretInternal newT2 res
       Stepper.Wait q -> do
         let performQuery = do
               m <- fetcher q
               (r, vm') <- liftIO$ stToIO $ runStateT m vm
-              interpret fetcher iterConf vm' (k r)
+              let newT = (t :: InterpTask m) { vm = vm', stepper = (k r) }
+              interpretInternal newT res
```
blishko: Can't `newT` be inlined?

src/EVM/SymExec.hs
```diff
@@ -695,16 +770,17 @@ verify :: App m
...
 verify solvers fetcher opts preState post = do
-  (expr, res, _) <- verifyInputs solvers opts fetcher preState post
-  pure $ verifyResults preState expr res
+  (ends1, partials) <- verifyInputs solvers opts fetcher preState post
+  (ends2, results) <- verifyResults preState ends1
+  pure (ends2 <> fmap snd partials, filter (not . isQed) results)
 
-verifyResults :: VM Symbolic RealWorld -> Expr End -> [(SMTResult, Expr End)] -> (Expr End, [VerifyResult])
-verifyResults preState expr cexs = if null cexs then (expr, [Qed]) else (expr, fmap toVRes cexs)
+verifyResults :: App m => VM Symbolic RealWorld -> [(SMTResult, Expr End)] -> m ([Expr End], [VerifyResult])
```
blishko: You don't need monadic context here, no?
This function is pure.

src/EVM/SymExec.hs
```diff
@@ -719,42 +795,43 @@ verifyInputs
...
-  when (conf.dumpExprs && conf.simp) $ liftIO $ do
-    let exprSimplified = Expr.simplify expr
-    T.writeFile "simplified.expr" (formatExpr exprSimplified)
-    T.writeFile "simplified-conc.expr" (formatExpr $ Expr.simplify $ mapExpr Expr.concKeccakOnePass exprSimplified)
+  ends <- executeVM fetcher opts.iterConf preState
+  -- when conf.dumpExprs $ liftIO $ T.writeFile "unsimplified.expr" (formatExpr expr)
+  -- when (conf.dumpExprs && conf.simp) $ liftIO $ do
+  --   let exprSimplified = Expr.simplify expr
+  --   T.writeFile "simplified.expr" (formatExpr exprSimplified)
+  --   T.writeFile "simplified-conc.expr" (formatExpr $ Expr.simplify $ mapExpr Expr.concKeccakOnePass exprSimplified)
```
blishko: Probably just delete this?

blishko (commented):
Looks good!
I left a few comments and questions.

test/test.hs
msooseth: I think it didn't work because of some clash. Haskell is kinda dumb. You can try removing it but I think it will fail...

test/test.hs
msooseth: Fair. It will make the PR larger I think. I'm good with `paths` actually.

src/EVM/SymExec.hs
msooseth: Yes, but I wanted to keep it easier to read. I think it's a rather complicated expression.

src/EVM/SymExec.hs
msooseth: I can change it back. I think I added it only because I got tired of all the `liftIO $` but yeah, let me fix this.

msooseth: OK, changed, but will need to wait to push due to airport restrictions :)

src/EVM/SymExec.hs
msooseth: Yep, sorry, fixed. Will push once out of the airport :)

src/EVM/SymExec.hs
msooseth: Ah shit. It's already added here:

```
        pushTo #constraints $ Expr.simplifyProp (addrExpr .== val)
```

for `runAllPaths`. Same for `runBothPaths`:

```haskell
      runBothPaths loc _ (Case v) = do
        assign #result Nothing
        pushTo #constraints $ if v then Expr.simplifyProp (Lit 0 ./= condSimpConc)
                                   else Expr.simplifyProp (Lit 0 .== condSimpConc)

```

So adding the constraint is not needed I think. Removing.

src/EVM/SymExec.hs
msooseth: Good point. As per above, I'll close this and give updates about it there.

src/EVM/SymExec.hs
msooseth: Wow... that thing was wrong. This: `Expr.simplifyProp (Lit 0 ./= condSimpConc)` when replaced by  `PNeg (PEq (Lit 0) cond)` it actually works. Otherwise, nope. Ooops. Fixing. And then we can remove adding this condition, since it's added correctly by all `continue`-s, i.e. for `runAllPaths` and for `runBothPaths`. Nice, thanks!

msooseth: Regarding the shared code... I think they are subtly different? I nevertheless tried to contract things, especially around `newT`. Let me know what you think!

src/EVM/SymExec.hs
msooseth: OK, fixed, thanks!

src/EVM/SymExec.hs
msooseth: Yep, fixed!

test/test.hs
msooseth: It says:

```
test/test.hs:209:23: error: [GHC-99339]
    • Ambiguous record update with field ‘iterConf’
      This field appears in both datatypes ‘VeriOpts’ and ‘InterpTask’
    • In the expression:
        defaultVeriOpts {iterConf = defaultIterConf {maxIter = Just 5}}
      In an equation for ‘veriOpts’:
          veriOpts
            = defaultVeriOpts {iterConf = defaultIterConf {maxIter = Just 5}}
```

test/test.hs
msooseth: Yay, done! Also changed it in `SymExec.hs`. Thanks!

src/EVM/SymExec.hs
msooseth: Fixed, sorry!

test/test.hs
blishko: Right, OK, let's keep this for now.

blishko (commented):
I think this looks good!
Can you rebase on top of current main?
I think you will have to remove some `RealWorld` parameters to be compatible with current main.

I would like to run a few experiments afterwards to have some performance numbers.

sambacha:
> I think this looks good!
> 
> Can you rebase on top of current main?
> 
> I think you will have to remove some `RealWorld` parameters to be compatible with current main.
> 
> 
> 
> I would like to run a few experiments afterwards to have some performance numbers.

yes plz

msooseth:
> I think this looks good! Can you rebase on top of current main? I think you will have to remove some `RealWorld` parameters to be compatible with current main.
> 
> I would like to run a few experiments afterwards to have some performance numbers.

Done! I merged main into this, which is sometimes easier to do. Looks very similar anyway in the git history. I hope it's good :) We can also squash-merge (when merging the PR), if you like, then it will look like a single commit.

msooseth:
(oops, my merge was a bit bad, sorry, I'll fix soon)

msooseth:
I have now fixed the biggest issue. I'll fix the 2 tests that fail tonight when I have more time & better internet.

src/EVM/SymExec.hs
msooseth: OK, I now made it a lot cleaner -- basically single line each, so a lot easier on the eyes! Let me know what you think :)

msooseth:
OK, this should finally pass all tests :) @blishko can you do a final review maybe?

PR #910, opened 2025-10-27
Test: Refactor blockchain tests to depend less on Contract type

blishko:
We introduce a new helper type, Blockchain contract, that is used only locally. It is a simple projection of a concrete contract. The benefit is that it can be trivially parsed from JSON and we can avoid akward definition of FromJSON for Contract.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
LGTM!

PR #913, opened 2025-10-27
Also warn on precompile address fetch

msooseth:
## Description
We should warn on precompile addresses? I am not sure whether we should just return 0 without warning, though. Something for Lexi or someone with deep knowledge could perhaps help with.

The tests are also better, it compiles the code rather than giving raw assembly. So it should be easier to follow what's going on.

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -18,6 +18,10 @@ module EVM.Fetch
   , saveCache
   , RPCContract (..)
   , makeContractFromRPC
+  -- Below 3 are needed for Echidna
+  , fetchSlotWithSession
+  , fetchSlotWithCache
+  , fetchWithSession
   ) where
 
 import Prelude hiding (Foldable(..))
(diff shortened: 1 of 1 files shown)


test/clitest.hs
```diff
@@ -247,14 +247,48 @@ main = do
...
+          contract C {
+            function computeHash() public view {
+                bytes memory data = hex"68656c6c6f";
+                (bool success,) = address(923741898892).staticcall(data);
+                assert(success);
+            }
+          }
+         |])
+        (_, stdout, stderr) <- readProcessWithExitCode "cabal" ["run", "exe:hevm", "--", "equivalence", "--code-a", Types.bsToHex c, "--code-b", Types.bsToHex c2] ""
+        stdout `shouldContain` "Warning: no RPC"
```
blishko: Should we spell out the full warning?
```suggestion
        stdout `shouldContain` "Warning: no RPC info provided"
```

msooseth: Yeah OK :)

msooseth:
Fair enough, let me fix!

msooseth:
Fixed now, sorry. New PR opened for the updated tests.

blishko (approved):
LGMT!

PR #915, opened 2025-10-27
Test: Small reorganization of blockchain tests

blishko:
## Description
We refactor the blockchain tests code a bit.
As part of this, we specify which functions are exported and accessible from outside.
This shows that there are several functions that are unused, so we delete them.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko:
Yes, more tests should run on CI.
I don't see a reason why they should be excluded.

PR #916, opened 2025-10-28
Move to macos15-intel

msooseth:
## Description
The Intel macos-13 runner is being deprecated. This moves to macos-15-intel. Also moves macos-14 to macos-15 (which is arm64). And now it's no longer mixed `-latest` and `-15` in macos.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      main
+      - main
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -3,7 +3,7 @@ on:
   pull_request:
   push:
     branches:
-      main
+      - main
```
msooseth, line 6: this was actually incorrect yaml. Fixed.

blishko (approved):
LGTM!

PR #919, opened 2025-10-28
Update cli tests to use solc rather than raw assemby

msooseth:
## Description
Update clitest to use solc rather than raw assembly

## Checklist

- [ ] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko:
Looks like you have to decide if `no` should be capitalized or not in the warning message :)

msooseth:
Ooops, sorry, yes, fixed!

Issue #917, opened 2025-10-28
Decoding of bytes1 array fails in some cases


gustavo-grieco:
When using arrays, sometimes the decoding is incorrect when a counter example is found. To reproduce:

1. Create a new foundry project: `forge init array-decoding`
2. Add a new file in `src`:
```solidity
contract C {
	bool public IS_TEST = true;
	uint256 state;
	function prop() public {
		assert(state == 0);
	}
	function inc() external { state++; }
        function prove_callAndCheckProperty(bytes1[36] calldata arr) external {
        address target = address(this);
        bool success;

        assembly {
            // Allocate memory for the calldata copy
            let dataPtr := mload(0x40)
            let dataLen := 36 // N = 32 bytes

            // Copy calldata (after 4-byte selector) into memory
            calldatacopy(dataPtr, 0x04, dataLen)

            // Update the free memory pointer
            mstore(0x40, add(dataPtr, dataLen))

            // Perform the low-level call
            success := call(gas(), target, 0, dataPtr, dataLen, 0, 0)
        }

        if (!success) revert("raw call failed");
        prop();
    }

}
```
3. Run hevm: `force clean ; forge build --ast ; hevm test --debug`

Output in the current `master` () is:
```
[RUNNING] prove_callAndCheckProperty(bytes1[36])
   Exploring call prefix 0xd0a773af
   Flattening expression
   Exploration finished, 6 branch(es) to check in call prefix 0xd0a773af
   Keccak preimages in state: 0
   Checking for reachability of 1 potential property violation(s) in call prefix 0xd0a773af
   SMT result: Cex SMTCex {vars = fromList [(Var "arg1-a-0",0x371303c000000000000000000000000000000000000000000000000000000000)], addrs = fromList [(SymAddr "miner",0xfFFfFFfFfFFffFFffFfFfFffffFFffffFFfFFfE5),(SymAddr "origin",0x1A)], BuFfErS = fROmLIsT [(ABsTRaCtBuF "txdata",Comp Base {byte = 0, length = 0x1})], store = fromList [], blockContext = fromList [], txContext = fromList []}
   Found 1 potential counterexample(s) in call prefix 0xd0a773af
[...]

PR #920, opened 2025-10-29
Fix packed array of bytes1

msooseth:
This partially addresses  #917 The issue is that bytes `bytes1[36]` is packed as per ABI spec. From the ABI documentation: https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding

```
bytes, of length k (which is assumed to be of type uint256):

enc(X) = enc(k) pad_right(X), i.e. the number of bytes is encoded as a uint256 followed by the actual value of X as a byte 
sequence, followed by the minimum number of zero-bytes such that len(enc(X)) is a multiple of 32.
```

However, we treated it as a regular Array, which is padded to 32B for each element.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/contracts/fail/bytes1-packed.sol
+++ b/test/contracts/fail/bytes1-packed.sol
@@ -1,0 +1,41 @@
+contract C {
+  bool public IS_TEST = true;
+  uint256 state;
+
+  function prop() public view {
+    assert(state == 0);
+  }
+
+  function inc() external { state++; }
+
+  function prove_callAndCheckProperty(bytes1[36] calldata arr) external {
+    address target = address(this);
+    bool success;
+    assembly {
+      // Allocate memory for the calldata copy
+      let dataPtr := mload(0x40)
+      let dataLen := 36 // N = 32 bytes
+
+      // Copy calldata (after 4-byte selector) into memory
+      calldatacopy(dataPtr, 0x04, dataLen)
+
+      // Update the free memory pointer
+      mstore(0x40, add(dataPtr, dataLen))
+
+      // Perform the low-level call
+      // this should call the inc() function, which has selector:
+      //```
+      //cast keccak "inc()"
+      //0x371303c051bff726100ad13871cababf50c20dd920fca137e519f98f089a74b4
+      //```
+      // which happens to have the first 4 bytes as
+      // 0x37, 0x13 0x03, 0xc0
+      // This will increment `state` by 1, thereby violating the property
+
+      success := call(gas(), target, 0, dataPtr, dataLen, 0, 0)
+    }
+
+    if (!success) revert("raw call failed");
+    prop();
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 - Fix incorrect simplification rule for `PEq (Lit 1) (IsZero (LT a b))`
+- bytes of length k in ABI are packed, as per ABI spec
 
 ## Changed
 - Replaced RPC mocking by a full block cache support. This allows users to cache responses from an RPC
--- a/src/EVM/ABI.hs
+++ b/src/EVM/ABI.hs
@@ -222,6 +222,11 @@ abiValueType = \case
 getAbi :: AbiType -> Get AbiValue
 getAbi t = label (Text.unpack (abiTypeSolidity t)) $
   case t of
+    -- Special case: static array of bytes1 is packed: `bytes, of length k [...] enc(X) = enc(k) pad_right(X)`
+    AbiArrayType n (AbiBytesType 1) -> do
+      bs <- getBytesWith256BitPadding n
+      pure $ AbiArray n (AbiBytesType 1) (Vector.fromList [AbiBytes 1 (BS.singleton (BS.index bs i)) | i <- [0..n-1]])
+
     AbiUIntType n  -> do
       let word32Count = 8 * div (n + 255) 256
       xs <- replicateM word32Count getWord32be
@@ -263,6 +268,7 @@ getAbi t = label (Text.unpack (abiTypeSolidity t)) $
 
     AbiFunctionType ->
       AbiFunction <$> getBytesWith256BitPadding (24 :: Int)
+
 
 putAbi :: AbiValue -> Put
 putAbi = \case
@@ -350,6 +356,8 @@ abiHeadSize x =
         AbiAddress _ -> 32
         AbiBool _    -> 32
         AbiTuple v   -> sum (abiHeadSize <$> v)
+        -- Special case: static array of bytes1 is packed: `bytes, of length k [...] enc(X) = enc(k) pad_right(X)`
+        AbiArray n (AbiBytesType 1) _ -> roundTo32Bytes n
         AbiArray _ _ xs -> sum (abiHeadSize <$> xs)
         AbiFunction _ -> 32
         _ -> internalError "impossible"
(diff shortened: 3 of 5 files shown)


gustavo-grieco:
Why this is a partial fix?

msooseth:
@gustavo-grieco only because it doesn't remove the trailing zeros... not sure they must be removed, but they probably could be removed. Either way, it definitely fixes the bigger issue, and the printing of the output will allow a user to just send in the counterexample that is short.

test/contracts/fail/bytes1-packed.sol
```diff
@@ -0,0 +1,30 @@
+contract C {
+	bool public IS_TEST = true;
+	uint256 state;
+	function prop() public view {
+		assert(state == 0);
+	}
+	function inc() external { state++; }
+      function prove_callAndCheckProperty(bytes1[36] calldata arr) external {
```
blishko: Looks like the indentation is wrong here.

test/contracts/fail/bytes1-packed.sol
```diff
@@ -0,0 +1,30 @@
...
+            let dataLen := 36 // N = 32 bytes
+
+            // Copy calldata (after 4-byte selector) into memory
+            calldatacopy(dataPtr, 0x04, dataLen)
+
+            // Update the free memory pointer
+            mstore(0x40, add(dataPtr, dataLen))
+
+            // Perform the low-level call
+            success := call(gas(), target, 0, dataPtr, dataLen, 0, 0)
```
blishko: Can you explain what is this going to call?

test/contracts/fail/bytes1-packed.sol
msooseth: Sorry, it was a mix of tab and spaces. All spaces now and it should be fine.

test/contracts/fail/bytes1-packed.sol
msooseth: @gustavo-grieco came up with this example, I just copied it :)
@gustavo-grieco can you maybe give some context?

msooseth: I think the idea is to call `inc()`

msooseth: Yes, it calls `inc` in order to violate
```
assert(state == 0);
```

because:
```shell
cast keccak "inc()" 
0x371303c051bff726100ad13871cababf50c20dd920fca137e519f98f089a74b4
```

And indeed the correct CEX, is:
```
0x37, 0x13, 0x03, 0xc0
```
Which are the first 4 bytes of that keccak, i.e. the function selector. 

So the idea is for the system to calculate the function it needs to call that increments the `state` so that the `state == 0` no longer holds.

msooseth: OK, I now added this to the code, can you check again?

test/contracts/fail/bytes1-packed.sol
```diff
@@ -0,0 +1,41 @@
...
+      // this should call the inc() function, which has selector:
+      //```
+      //cast keccak "inc()"
+      //0x371303c051bff726100ad13871cababf50c20dd920fca137e519f98f089a74b4
+      //```
+      // which happens to have the first 4 bytes as
+      // 0x37, 0x13 0x03, 0xc0
+      // This will increment `state` by 1, thereby violating the property
+
+      success := call(gas(), target, 0, dataPtr, dataLen, 0, 0)
```
blishko, line 35: Oh, I see, this basically just calls this contract again, with the calldata passed in as argument to this function.
So the number `36` is rather arbitrary here, no?
`4` would be sufficient, since we only need the selector (which has 4 bytes) to be able to call `inc`, no?

Why was `36` chosen as the magic number?

gustavo-grieco: Yes, sorry, the 36 was selected to allow at least one parameter, but here it can be reduced.

test/contracts/fail/bytes1-packed.sol
gustavo-grieco: Yes, this is correct. This was an experiment to make hevm/echidna run a symbolic transaction and perform some checks after that (e.g. calling a functions with no parameters). It can be generalized to more transactions.

msooseth:
Ooops:

```
2025-10-29T16:28:45.2063660Z hevm>     symbolic-abi encoding-vs-solidity-2-args:                     FAIL (1.06s)
2025-10-29T16:28:45.2064412Z hevm>       *** Failed! (after 95 tests):
2025-10-29T16:28:45.2064837Z hevm>       Exception:
2025-10-29T16:28:45.2065433Z hevm>         Internal Error: Left (Revert (ConcreteBuf "")) -- CallStack (from HasCallStack):
2025-10-29T16:28:45.2066214Z hevm>           internalError, called at test/test.hs:5620:13 in main:Main
2025-10-29T16:28:45.2066601Z hevm>         CallStack (from HasCallStack):
2025-10-29T16:28:45.2067037Z hevm>           error, called at src/EVM/Types.hs:1547:19 in hevm-0.56.0-7NJEoXrfiAACDJcfgNGfWY:EVM.Types
2025-10-29T16:28:45.2067532Z hevm>           internalError, called at test/test.hs:5620:13 in main:Main
2025-10-29T16:28:45.2068110Z hevm>       (SymbolicAbiVal 0x6454be4953aaaf542dbf21db8d4244833ebdda80,SymbolicAbiVal [0x7d, 0xf9, 0xf6, 0x0f, 0x0b])
2025-10-29T16:28:45.2068749Z hevm>       Use --quickcheck-replay="(SMGen 3482784650225922177 7199675813539884525,94)" to reproduce.
2025-10-29T16:28:45.2069481Z hevm>       Use -p '/symbolic-abi encoding-vs-solidity-2-args/' to rerun this test only.
```

and:

```
2025-10-29T16:40:42.3228998Z     abi-encoding-vs-solidity-2-args:                              FAIL (1.31s)
2025-10-29T16:40:42.3230776Z       *** Failed! (after 39 tests):
2025-10-29T16:40:42.3231304Z       Exception:
2025-10-29T16:40:42.3231956Z         Internal Error: Left (Revert (ConcreteBuf "")) -- CallStack (from HasCallStack):
2025-10-29T16:40:42.3232951Z           internalError, called at test\test.hs:5620:13 in hevm-0.56.0-inplace-test:Main
2025-10-29T16:40:42.3233731Z         CallStack (from HasCallStack):
2025-10-29T16:40:42.3234386Z           error, called at src\EVM\Types.hs:1547:19 in hevm-0.56.0-inplace:EVM.Types
2025-10-29T16:40:42.3235315Z           internalError, called at test\test.hs:5620:13 in hevm-0.56.0-inplace-test:Main
2025-10-29T16:40:42.3236082Z       ([0x19, 0x34, 0x1d, 0x0e, 0x3c, 0x0d, 0x01, 0x21, 0x12, 0x1f, 0x0e, 0x15, 0x30, 0x1a, 0x3b, 0x35, 0x06, 0x29, 0x32, 0x3a, 0x21, 0x14, 0x2b, 0x23, 0x00, 0x30, 0x10, 0x1e, 0x38],0x0000000000000000000000000000000000006866)
2025-10-29T16:40:42.3237134Z       Use --quickcheck-replay="(SMGen 9011992352728319783 16502856406791791583,38)" to reproduce.
2025-10-29T16:40:42.3237723Z       Use -p '/abi-encoding-vs-solidity-2-args/' to rerun this test only.
```

blishko:
I confirm this problem is not reproducible on main, so it must be introduced here.

msooseth:
Yes, I know. I am working on it.

msooseth:
Hahah, I vibe-coded! With Claude Code. Actually seems to work. Kinda crazy. It wanted to originally remove the test to fix it :D But then I told it off and now it works!

msooseth:
As per #971 comment, this is not a bug. Ooops. Closing.

PR #922, opened 2025-10-29
Cabal: Remove unneeded private modules from test-common

blishko:
All these modules are already compiled as part of test-utils library which is a dependency of test-common


msooseth:
@blishko do you plan to fix this up?

blishko:
Yes, I plan to get back to this, but I want to go through the refactoring of our tests first.

PR #923, opened 2025-10-30
RPC: Small cleanup

blishko:
## Description

In processing query to fetch a slot, the current code checks the cache twice.
First directly in the handler code, secondly in function `fetchSlotFrom`.
Since this function is not used anywhere else, we inline it and eliminate one unnecessary cache check.

We also remove debug printing that is now actually unreachable.


blishko:
I am pretty sure they are unreachable now.
If `addr <= 17`, then `withSession` is going to hit this case: `a | a <= 0xdeadbeef -> pure def`
And it will just proceed immediately with `continue` (with empty contract/zero value). It never enters the lambda where the debug prints are located.

PR #925, opened 2025-11-03
Refactor RPC handling

blishko:
## Description

The main proposed change in RPC is that we do not require a session if no RPC info is provided.
No RPC info should be interpreted as "unknown address contains unknown contract".
No RPC call should be made in such case and so no session is required here.

We also remove the definition of `RpcInfo` as a `Semigroup`.
I don't think it makes mathematical sense to say this is a semigroup.
Instead we make an explicit function that represents "no RPC info".

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/EVM/Test/BlockchainTests.hs
+++ b/test/EVM/Test/BlockchainTests.hs
@@ -99,17 +99,16 @@ prepareTests = do
   rootDir <- liftIO rootDirectory
   liftIO $ putStrLn $ "Loading and parsing json files from ethereum-tests from " <> show rootDir
   cases <- liftIO allTestCases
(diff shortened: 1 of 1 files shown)


blishko:
Hmm. I only removed the tests that were testing the presence (or absence) of a Warning message that is being removed. We agreed that it should be removed, because the scenario that triggered it is actually a perfectly valid scenario.

I did not add any new errors. I am not sure what kind of tests I should add.

msooseth (approved):
Fair enough :) Approving!

PR #926, opened 2025-11-04
Expr: Remove expensive unnecessary simplification

### LINKED ISSUE #831: Investigate a special simplification case for `WriteWord`
Currently, we have a special case when simplifying `WriteWord` expression.
If we are writing on a concrete index into a concrete buffer, this simplification overwrites the bytes in the concrete buffer to a zero word first, since these bytes will be overwritten anyway.
The idea, as far as I can tell, is to canonicalize the base concrete buffers.
The change was introduced in https://github.com/ethereum/hevm/commit/d2e38f1d4.

I don't quite understand the reasoning behind this.
We should investigate if this change is helpful.
If yes, in what cases?
If no, we can just delete this special case.

blishko:
## Description
This transformation is a bit expensive to check and should not be attempted repeatedly, which is what we are doing now. It might make sense to apply it once at the very end before we encode to SMT.
However, the simplification system does not support that yet. For now the simplification can be removed.
We could attempt to bring it back later.

Resolves #831.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   node via `--cache-dir dir`.
 - Changed `verify*` methods to always require postcodition.
 - Removed type parameter of mutable memory from VM definition.
+- Removed simplification that were rewriting concrete bytes-to-be-overwritten
+  with zero bytes. Benefits were unclear while it had negative effect on
+  analysis' performance.
 
 ## [0.56.0] - 2025-10-13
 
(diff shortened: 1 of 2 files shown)


msooseth (approved):
Yeah, doesn't seem crucial. Would be nice to know why it was written, i.e. if there was a particular reason. I guess not. Let's remove it. We can undo the PR later if need be, but I doubt it will be needed. May be useful to add a ChangeLog item though...

blishko:
Changelog entry added.

PR #929, opened 2025-11-12
Expr: Canonicalize transformations

blishko:
## Description
On expression level, we convert all inequalities to use (signed) less-than operator, (S)LT.

On propositional level, we express inequalities using PLT and PLEq. Here, it might be beneficial to keep leq-or-equal operator, so as to express as much as possible with positive atoms, avoiding negation.

We also take advantage of "bool-like" expressions, i.e., word-expressions that we can express on the propositional level. These include "IsZero", "(S)LT", "Eq".
The result of these operations can only be 0 or 1, so we can simplify a propositon (PEq (Lit 0) e) where e is bool-like (and also if it is equal to Lit 1).

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko:
We should definitely gather more unit tests for simplifications!

PR #934, opened 2025-11-26
Symbolic execution queue should be used to run SMT as well

msooseth:
## Description

This now finally does what I was aiming for :) You can pass in a function that will be run on the `Expr End` and the results will be collated. This is now done with the `handler`.

Notice that we now have to add a task for the `Partial` as well as the `handler` needs to be executed on this, here:

```diff
-                    -- got us to this point and return a partial leaf for the other side
-                    liftIO . atomically $ writeTChan res (Partial [] (TraceContext (Zipper.toForest vm.traces) vm.env.contracts vm.labels) (MaxIterationsReached vm.state.
pc vm.state.contract))
+                    -- got us to this point and queue a task to return a partial leaf for the other side
+                    let partialLeaf = Partial [] (TraceContext (Zipper.toForest vm.traces) vm.env.contracts vm.labels) (MaxIterationsReached vm.state.pc vm.state.contract
)
+                    liftIO $ atomically $ modifyTVar numTasks (+1)
+                    liftIO $ writeChan taskq $ t { vm = vm, stepper = pure partialLeaf }
```

This was actually done bye Claude Code at first, but I had to adjust quite a bit.

Now it finds the Cex for prove_inv_identity in `src/ABDKMath64x64PropertyTests.sol:CryticABDKMath64x64Properties` instantaneously, but keeps running.

I will write a cexHander for equivalence checking as a follow-up PR.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -431,7 +431,7 @@ interpret fetcher iterConf vm stepper handler = do
         nt <- readTVar numTasks
         when (np' == 0 && nt == 0) $ putTMVar allProcessDone ()
 
-getOneExpr :: forall m a . (MonadIO m, ReadConfig m, App m)
+getOneExpr :: forall m a . App m
   => InterpTask m a
   -> Chan ()
   -> Chan (Process m a)


msooseth:
Now runs like:

```
   Counterexample:   [validated]
     calldata: prove_inv_identity(-108003701288662363396026832958930103201)
   [FAIL] prove_inv_identity
```

This will need an update to the docs. But it should be a lot easier to use for users.

gustavo-grieco:
This is great, a leap forward in terms of usability (specially for getting counter examples!)

msooseth:
Yes, this was the actual goal of the symbolic queue that Martin has helped with a lot. So we are finally making progress to be much more user-friendly, and show relevant results immediately.

gustavo-grieco:
Oh, btw, should stop the exploration immediately if a CEX if found?

msooseth:
> Oh, btw, should stop the exploration immediately if a CEX if found?

Yep, we are thinking about adding that option exactly :) I'll add it in a sec :)

msooseth:
> > Oh, btw, should stop the exploration immediately if a CEX if found?
> 
> Yep, we are thinking about adding that option exactly :) I'll add it in a sec :)

OOops, forgot to add that. Will do it tomorrow as a separate PR. Should be very simple :)

src/EVM/SymExec.hs
```diff
@@ -762,47 +807,50 @@ verifyResults preState res = (map snd res, fmap toVRes res)
...
+    pure (smtResult, mPartial)
+
+  let (smtResults, partials) = unzip results
   when conf.debug $ liftIO $ do
-    printPartialIssues ends ("the call " <> call)
-    putStrLn $ "   Exploration finished, " <> show (length ends) <> " branch(es) to check in call " <> call <> " of which partial: "
-                <> show (length partials)
-    when (conf.verb >=2) $ putStrLn $ "   results:\n" <> prettyvmresults ends
+    putStrLn $ "   Exploration finished, " <> show (length results) <> " branch(es) checked in call " <> call <> " of which partial: "
+                <> show (length smtResults)
```
blishko: I think this message here is misleading now.
At this point, not only exploration has finished, but also all the solving, no?
At least before, "exploration" meant only the first phase, excluding the solving.

src/EVM/UnitTest.hs
```diff
@@ -168,7 +168,7 @@ validateCex :: forall m . App m
   -> Fetch.Fetcher Concrete m
   -> VM Concrete
   -> ReproducibleCex
-  -> m Bool
+  -> m (Err Bool)
```
blishko: What is the purpose of this change? Is it necessary?

src/EVM/SymExec.hs
```diff
@@ -321,102 +322,143 @@ freezeVM vm = do
...
+      -- Decrement and check if all done
+      liftIO $ atomically $ do
+        np <- readTVar numProcs
+        let np' = np - 1
+        writeTVar numProcs np'
+        -- Check if both interpretation and processing are done
+        nt <- readTVar numTasks
+        when (np' == 0 && nt == 0) $ putTMVar allProcessDone ()
+
+getOneExpr :: forall m a . (MonadIO m, ReadConfig m, App m)
```
blishko: `ReadConfig` is already included in `App`, no?
Not sure about `MonadIO`, can we remove it as well?

hevm.cabal
```diff
@@ -188,6 +188,7 @@ library
     template-haskell                  >= 2.19.0 && < 3,
     extra                             >= 1.7.14 && < 2,
     aeson-pretty                      >= 0.8.8 && < 0.9.0,
+    monad-loops                       >= 0.4.1 && < 0.5,
```
blishko, line 191: Do we need this?

blishko (commented):
I think this looks good! I have just a few small questions.

hevm.cabal
msooseth: Yes, for `forever`, which I think is really nice.

msooseth: Oh, actually, no, we need it for `whileM`, so we don't have to define it!

src/EVM/SymExec.hs
msooseth: Good point! I now did this:
```diff
commit c5a56bb3ebee5bcf42dc6cf7e1c2bb78c5920e9c (HEAD -> queue-check, origin/queue-check)
Author: Mate Soos <mate.soos@argot.org>
Date:   Thu Dec 4 12:59:01 2025 +0100

    Better printing to the user

diff --git a/src/EVM/SymExec.hs b/src/EVM/SymExec.hs
index d26971b4..d702c11b 100644
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -820,7 +820,9 @@ verifyInputsWithHandler
 verifyInputsWithHandler solvers opts fetcher preState post cexHandler = do
   conf <- readConfig
   let call = mconcat ["prefix 0x", getCallPrefix preState.state.calldata]
-  when conf.debug $ liftIO $ putStrLn $ "   Exploring call " <> call
+  when conf.debug $ liftIO $ do
+    putStrLn $ "   Keccak preimages in state: " <> (show $ length preState.keccakPreImgs)
+    putStrLn $ "   Exploring call " <> call

   results <- executeVM fetcher opts.iterConf preState $ \leaf -> do
     -- Extract partial if applicable
@@ -846,9 +848,8 @@ verifyInputsWithHandler solvers opts fetcher preState post cexHandler = do
   when conf.debug $ liftIO $ do
     putStrLn $ "   Exploration and solving finished, " <> show (length results) <> " branch(es) checked in call " <> call <> " of which partial: "
                 <> show (length smtResults)
-    putStrLn $ "   Keccak preimages in state: " <> (show $ length preState.keccakPreImgs)
     let cexs = filter (\(res, _) -> not . isQed $ res) smtResults
-    putStrLn $ "   Found " <> show (length cexs) <> " potential counterexample(s) in call " <> call
+    putStrLn $ "   Found " <> show (length cexs) <> " counterexample(s) in call " <> call

   pure (smtResults, catMaybes partials)
   where
```

So the printing should be better I think!

src/EVM/SymExec.hs
msooseth: Oh, good point! Removed them all except `App m`

src/EVM/UnitTest.hs
msooseth: Yep, removed it completely, and now it's even simpler, see 5540297eb3e29841ee6e90df9f4b06f2f96128d0 :) Thanks, this really helped!

msooseth:
Thanks, you have made this a lot better  :) Can you take another look maybe?

Issue #850, opened 2025-09-04
Improve our constant propagation to also store and check inequalities


msooseth:
Currently, only constants such as a=10 are stored and checked. a<10 and a>10 will not result in UNSAT.

Could help fix #849

gustavo-grieco:
Should we close this one now that #935 was merged?

blishko:
Yes, I think #935 resolves this!

Issue #849, opened 2025-09-04
hevm gets stuck proving in a simple property


gustavo-grieco:
Take a look to this simple invariant from abdk math:
```solidity
    // Test for negative values, should revert as ln is not defined
    function prove_ln_negative(int128 x) public view {
        require(x < ZERO_FP); // ZERO_FP *is* 0

        try this.ln(x) {
            // Unexpected, should revert
            assert(false);
        } catch {
            // Expected
        }
    }
    ...
    function ln(int128 x) public pure returns (int128) {
        return ABDKMath64x64.ln(x);
    }
```

Let's examine `ln`:

```solidity
  function ln (int128 x) internal pure returns (int128) {
    unchecked {
      require (x > 0);

      return int128 (int256 (
          uint256 (int256 (log_2 (x))) * 0xB17217F7D1CF79ABC9E3B39803F2F6AF >> 128));
    }
  }
```
The `log_2` function is complicated and it contains loops. Not fun for hevm, but it should not necessary if the x > 0 and x < 0 are correctly simplified.

To reproduce with the latest hevm revision, clone the [abdk verification repository](https://github.com/gustavo-grieco/abdk-math-64.64-verification/) and run `make verify-hevm T=prove_ln_negative`.

blishko:
This may be similar to an issue I have encountered as well.
Currently, when exploring, `hevm` is not able to determine that a path condition that contains both `x > 0` and `x < 0` is unsatisfiable and it continues exploring such path.
In this case, I guess it means it goes on exploring `log_2`, even though it would not be necessary.

I was not sure it would be worth implementing more aggressive strategies for detecting infeasibility of path conditions.
If it would help on this example as well, then we should definitely do it!

msooseth:
We could probably "just" improve our constant propagation and discover this infeasibility. Currently, only equalities are propagated and checked. So two conditions like `a=1` and `a=0` will result in UNSAT during simplification -- which is always done during branching. However, `a<1` and `a>1` will not result in UNSAT. BTW, we could also run an SMT solver for each branch, to check if the conditions are not infeasible, but that may be a bit too much... Let me add a TODO. I don't want a full fledged simplex solver, but trivial things like a both <10 and >10 should be UNSAT.

gustavo-grieco:
This is no longer an issue. Should we close this or there is some follow up here?

blishko:
Solved by #935

PR #935, opened 2025-11-27
Propagate basic inequalities to discard some paths faster

gustavo-grieco:
## Description

An initial implementation of fix for #850. It instantly solves #849:

```
[RUNNING] prove_ln_negative(int128)
   Exploring call prefix 0x8909bc39
Interpretation finished, collected 4 results.
   Exploration finished, 4 branch(es) to check in call prefix 0x8909bc39 of which partial: 0
   Keccak preimages in state: 0
   Checking for reachability of 4 potential property violation(s) in call prefix 0x8909bc39
   Found 0 potential counterexample(s) in call prefix 0x8909bc39
   [EXPLORATION COMPLETE] prove_ln_negative -- explored 4 paths.
   symRun -- (cex,warnings,unexpectedAllRevert): (False,False,False)
   [PASS] prove_ln_negative

unitTest individual results: [(True,True)]
```

It clearly needs some tests, but wanted to make sure the approach is good.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -940,8 +940,8 @@ tests = testGroup "hevm"
         assertEqualM "Must not contain PBool False" False ((PBool False) `elem` propagated)
     , test "inequality-tight-bounds-violated" $ do
         let
-          -- x >= 5 && x <= 5 and x == 6 should be inconsistent
-          t = [PGEq (Var "x") (Lit 5), PLEq (Var "x") (Lit 5), PEq (Var "x") (Lit 6)]
+          -- x >= 5 && x <= 5 and x != 5 should be inconsistent
+          t = [PGEq (Var "x") (Lit 5), PLEq (Var "x") (Lit 5), PNeg (PEq (Var "x") (Lit 5))]
           propagated = Expr.constPropagate t
         assertEqualM "Must contain PBool False" True ((PBool False) `elem` propagated)
     , test "inequality-with-existing-equality-consistent" $ do


src/EVM/Expr.hs
```diff
@@ -1687,20 +1689,80 @@ constPropagate ps =
...
           s <- get
           case Map.lookup a s.values of
-            Just l2 -> when (l==l2) $ put ConstState {canBeSat=False, values=mempty}
+            Just l2 -> when (l==l2) $ put ConstState {canBeSat=False, values=mempty, lowerBounds=mempty, upperBounds=mempty}
             Nothing -> pure ()
         PNeg (PEq a b@(Lit _)) -> go $ PNeg (PEq b a)
+        
+        -- inequalities
+        PLT a (Lit b) -> updateUpper a (b - 1)
+        PLT (Lit a) b -> updateLower b (a + 1)
```
msooseth: I think this b+1 and b-1 could wrap around? so we need to be careful?

src/EVM/Expr.hs
```diff
@@ -1687,20 +1689,80 @@ constPropagate ps =
...
+        PLEq a (Lit b) -> updateUpper a b
+        PLEq (Lit a) b -> updateLower b a
+        PGT a (Lit b) -> updateLower a (b + 1)
+        PGT (Lit a) b -> updateUpper b (a - 1)
```
msooseth: again I think b+1/-1 can wrap around maybe?

blishko:
The idea looks good, I'll have a proper look later.

msooseth:
Note to self. I have created a PR to fix the rollaround: https://github.com/gustavo-grieco/hevm/pull/6/files

The PR itself does, as Gustavo mentioned:
```
cabal run -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_ln_negative
[RUNNING] prove_ln_negative(int128)
   [PASS] prove_ln_negative
```

In under 1s.

gustavo-grieco:
Should we create an issue to add a RPC test in a future PR?

msooseth:
I realized we probably want to propagate things like `PGT a b` and `b > 5`, then `a > 6`. Next PR!

test/clitest.hs
```diff
@@ -136,50 +136,6 @@ main = do
         let hexStr = Types.bsToHex c
         (_, stdout, _) <- readProcessWithExitCode "cabal" ["run", "exe:hevm", "--", "symbolic", "--solver", "empty", "--code", hexStr] ""
         stdout `shouldContain` "SMT solver says: Result unknown by SMT solver"
-
```
blishko, line 139: Why are we removing this test?

test/test.hs
```diff
@@ -895,6 +895,68 @@ tests = testGroup "hevm"
...
+          -- x > maxLit is impossible
+          t = [PGT (Var "x") (Lit 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff)]
+          propagated = Expr.constPropagate t
+        assertEqualM "Must contain PBool False" True ((PBool False) `elem` propagated)
+    , test "PGT-overflow-check" $ do
+        let
+          -- 0 > y is impossible
+          t = [PGT (Lit 0) (Var "y")]
+          propagated = Expr.constPropagate t
+        assertEqualM "Must contain PBool False" True ((PBool False) `elem` propagated)
```
blishko, line 922: I am not convinced we need to detect these in `constPropagate`.
I think these should be simplified to `False` in earlier stages, in prop simplification.

test/test.hs
```diff
@@ -895,6 +895,68 @@ tests = testGroup "hevm"
...
+    , test "inequality-tight-bounds-satisfied" $ do
+        let
+          -- x >= 5 && x <= 5 and x == 5 should be consistent
+          t = [PGEq (Var "x") (Lit 5), PLEq (Var "x") (Lit 5), PEq (Var "x") (Lit 5)]
+          propagated = Expr.constPropagate t
+        assertEqualM "Must not contain PBool False" False ((PBool False) `elem` propagated)
+    , test "inequality-tight-bounds-violated" $ do
+        let
+          -- x >= 5 && x <= 5 and x == 6 should be inconsistent
```
blishko: I would change this test to detect that `x >= 5 && x <= 5 and x != 5 is inconsistent`

test/clitest.hs
msooseth: The inequality solves this immediately, so the UNSAT cache is never activated...

msooseth: It was a rather flaky test anyway. It depended on the ordering of things. Most of the time it worked. I'd rather remove anyway.

test/test.hs
msooseth: I agree. However, This makes constPropagate total, which is kinda nice and clean. I saw this, but I thought that it's not too bad, actually.

blishko: I agree that we can keep them.

test/clitest.hs
blishko: That's probably true.
But I think the idea is that we wanted to have a test to check that the UNSAT cache is not totally broken.
We can remove this test, but then we should figure out how to test UNSAT cache functionality.

test/test.hs
msooseth: Written as a PR to the PR, here: https://github.com/gustavo-grieco/hevm/pull/7 once @gustavo-grieco merges it, it will appear here :)

test/clitest.hs
msooseth: Fair... let me try to write a test in solidity.

gustavo-grieco:
@msooseth changes were merged already

test/clitest.hs
msooseth: I can't see to write one. It's annoying. We should think about this a bit...

blishko: Let's create a separate issue for UNSAT cache testing.

gustavo-grieco:
I can handle the missing RPC test in a futuro PR

Issue #928, opened 2025-11-12
Release run is failing due to missing directory dependency


msooseth:
See: https://github.com/argotorg/hevm/actions/runs/19221887897/job/54941324710

```
2025-11-10T07:00:11.7220838Z        > Using Parsec parser
2025-11-10T07:00:11.7221092Z        > Configuring hevm-0.56.0...
2025-11-10T07:00:11.7221672Z        > CallStack (from HasCallStack):
2025-11-10T07:00:11.7222234Z        >   withMetadata, called at libraries/Cabal/Cabal/src/Distribution/Simple/Utils.hs:370:14 in Cabal-3.8.1.0:Distribution.Simple.Utils
2025-11-10T07:00:11.7222843Z        > Error: Setup: Encountered missing or private dependencies:
2025-11-10T07:00:11.7223151Z        > directory >=1.3.8.0 && <=1.3.9.0
2025-11-10T07:00:11.7223373Z        >
2025-11-10T07:00:11.7223823Z        For full logs, run 'nix log /nix/store/9ff0jz3l8xpqpmvdd75mp8z76d1h5srx-hevm-static-x86_64-unknown-linux-musl-0.56.0.drv'.
```

This release run is running once in a while, in order to catch such issues. We need to resolve this before we try to release it next time :)

msooseth:
I'm checking if relaxing it is good enough, here: https://github.com/msooseth/hevm/releases/tag/release%2F21.0.0

It checks whether:
```diff
-    directory                         >=1.3.8.0 && <= 1.3.9.0,
+    directory                         >=1.3.7.0 && <= 1.3.9.0,
```

fixes it. If it does, I'll create a PR here.

msooseth:
NICE. Let me see if the release is fixed by doing a "release" in msooseth :)

PR #943, opened 2025-12-08
This fixes release due to directory being too high

msooseth:
## Description

This fixes the release build issue that we had before. Another issue pops up, but that's for another PR to fix, besides, I think it's already fixed:

```
       > src/EVM/SymExec.hs:17:1: error: [-Wunused-imports, -Werror=unused-imports]
       >     The import of ‘Control.Monad’ is redundant
       >       except perhaps to import instances from ‘Control.Monad’
       >     To import instances alone, use: import Control.Monad()
       >    |
       > 17 | import Control.Monad (when, forM_, forM, forever)
       >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
So I 'd like to merge this.

Coses #928

## Checklist

I tested this as part of the msooseth/hevm release pipeline. It fixes the original problem.

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -143,7 +143,7 @@ library
   install-includes:
     ethjet/tinykeccak.h, ethjet/ethjet.h, ethjet/ethjet-ff.h, ethjet/blake2.h
   build-depends:
-    directory                         >=1.3.7.0 && <= 1.3.9.0,
+    directory                         >= 1.3.6 && < 1.4,
     system-cxx-std-lib                >= 1.0 && < 2.0,
     QuickCheck                        >= 2.13.2 && < 2.16,
     Decimal                           >= 0.5.1 && < 0.6,


elopez:
Are releases being built with GHC 9.6?

msooseth:
Releases are being built with the same nix setup, but with:

```
          nix build .#redistributable --out-link hevm
```

So I think it uses whatever GHC that the normal build is using? You can check it out in the `release.yml` file. I am not an expert at nix...

elopez:
Yeah, unfortunately I think that results in it using GHC 9.6 on Linux release builds due to pkgsStatic (newer GHCs on there have broken TemplateHaskell on the static set, so pkgsStatic defaults to 9.6 which works). We had to do a more complicated thing on Echidna to get static 9.8 builds through pkgsMusl. I saw some recent activity on nixpkgs working on this so maybe it will be better in a few weeks/months without us having to change anything.

I'm ok merging this change if it means 9.6+ builds work 😄

blishko:
I've just realized that it is weird that `1.3.7` works, but `1.3.8` does not.
There should not be a breaking change in the minor update, which I think `1.3.7 -> 1.3.8` is.
What was the problem with `1.3.8`?
Also, we cannot say that upper bound is `1.3.9` if `1.3.8` does not work.

blishko:
I have read some more and I think we should relax the upper bound to `< 1.4`.
Not sure about the lower bound, maybe that can be relaxed, but if `1.3.7` works with `GHC 9.6`, we can use that.

PR #946, opened 2025-12-08
Early abort flag

msooseth:
## Description

Adds `--early-abort` option to all of: test/symbolic/equivalence. Will throw away all queued up processes and tasks when the abort flag has been set. 

Needs test, probably.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -57,7 +57,7 @@ import EVM.Expr qualified as Expr
 import EVM.FeeSchedule (feeSchedule)
 import EVM.Format (formatExpr, formatPartial, formatPartialDetailed, showVal, indent, formatBinary, formatProp, formatState, formatError)
 import EVM.SMT qualified as SMT
-import EVM.Solvers (SolverGroup, checkSatWithProps)
+import EVM.Solvers (SolverGroup(..), checkSatWithProps)
 import EVM.Stepper (Stepper)
 import EVM.Stepper qualified as Stepper
 import EVM.Traversals (mapExpr, mapExprM, foldTerm)


src/EVM/Effects.hs
```diff
@@ -65,6 +66,7 @@ defaultConfig = Config
   , verb = 0
   , simp = True
   , onlyDeployed = False
+  , earlyAbort = False
```
gustavo-grieco, line 69: Should we make it `True` by default?

msooseth: Yes, I think so! Then the option would need to become `--no-early-abort

I think we also need to abort the SMT solvers. On the `abdk-math-64.64-verification`'s `prove_inv_identity`, it finds a CEX and displays it immediately, but waits until all SMT solvers terminate, which is like 5 minutes :S So this PR is not good enough as-is :S

gustavo-grieco: It still a large improvement in usability. Great work! 🙏

msooseth:
@gustavo-grieco OK, this is now kinda working, but it has NOT been pushed to the branch yet:

```
cabal run --enable-shared -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_inv_identity --early-abort
Resolving dependencies...
   Counterexample:   [validated]
     calldata: prove_inv_identity(-101034573947264812908054335791493761808)
   [FAIL] prove_inv_identity
      11x -> Aborted due to early 
```

It exits _immediately_.

src/EVM/Effects.hs
msooseth: I'm closing because it's probably best to discuss this someplace else I think? It's really just 3 lines of code either way. I want to concentrate making this PR work in this PR :) The 3-line change we can have separately :)

gustavo-grieco:
Very excited to test this with the rest of the abdk examples. One question: the model should be built first or Haskell lazyness can give us the counter example before the model is complete?

gustavo-grieco:
What is the status of this PR? Is there anything missing that we can help?

msooseth:
>Very excited to test this with the rest of the abdk examples. One question: the model should be built first or Haskell lazyness can give us the counter example before the model is complete?

I am not using lazyness. It's a monad, and it only issues the abort to other threads when it has completed computing the CTX.

> What is the status of this PR? Is there anything missing that we can help?

Well, it's working... but Martin & I discussed and we may want to do it differently. We may want to have "thowaway" SMT threads, where we can abort them anytime, and re-spin them anytime. This way we can have the same SMT being solved at the same time by multiple SMT solvers, and we could enforce strict time and memory limits (currently, this is NOT done, and some SMT solvers don't always respect the requested limits).

I might have a go at doing this rewrite.

msooseth:
OK, I am closing because with STRICT enforcement of time limits on the SMT solver, we can achieve the same effect in this particular verification task. But this way is much cleaner and more extensible, and allows a future `--early-abort` flag to be added much easier and with less noise and pain. I will open a new PR for this. But for now:


```
time cabal run --enable-shared -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_inv_identity --smttimeout 2
Configuration is affected by the following files:
- cabal.project
   Counterexample:   [validated]
     calldata: prove_inv_identity(-101034573947264812908054335791493761808)
   [FAIL] prove_inv_identity

Command exited with non-zero status 1
        Command being timed: "cabal run --enable-shared -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_inv_identity --smttimeout 2"
        User time (seconds): 7.89
        System time (seconds): 0.63
        Percent of CPU this job got: 356%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.38
```

So 2s wallclock time.

I will cut this PR down to stop the symbolic exploration with `--early-abort`, which is the easy part.

PR #947, opened 2025-12-08
[RFC] Track failures, expose stats, return cache info

gustavo-grieco:
## Description

@elopez kindly provided this WIP for some cache additional code. This allows third-party programs like Echidna to know if  the cache was hit or not. It also keeps tracks of the number of slots fetched per contract as well as track failures. This information is shown in Echidna, but it can also shown in hevm if you are interested (in this PR). After this is merged, we will proceed with the release of Echidna 2.3.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM/Fetch.hs
+++ b/src/EVM/Fetch.hs
@@ -24,6 +24,8 @@ module EVM.Fetch
   , fetchSlotWithCache
   , fetchWithSession
   , getCacheState
+  , FetchStatus(..)
+  , FetchResult(..)
   ) where
 
 import Prelude hiding (Foldable(..))
@@ -47,7 +49,6 @@ import qualified Data.ByteString.Lazy as BSL
 import Data.Bifunctor (first)
 import Control.Exception (try, SomeException)
 
-import Control.Monad.Trans.Maybe
 import Data.Aeson hiding (Error)
 import Data.Aeson.Optics
 import Data.ByteString qualified as BS
@@ -55,7 +56,7 @@ import Data.Text (Text, unpack, pack)
 import Data.Text qualified as T
 import Data.Foldable (Foldable(..))
 import Data.Map.Strict qualified as Map
-import Data.Maybe (fromMaybe, isJust, fromJust, isNothing)
+import Data.Maybe (fromMaybe, fromJust, isNothing)
 import Data.Set qualified as Set
 import Data.Set (Set)
 import Data.Vector qualified as RegularVector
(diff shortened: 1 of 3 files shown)


src/EVM/Fetch.hs
```diff
@@ -460,9 +540,9 @@ oracle solvers preSess rpcInfo q = do
           Nothing -> do
             when (conf.debug) $ liftIO $ putStrLn $ "Fetching contract at " ++ show addr
             let (block, url) = fromJust rpcInfo.blockNumURL
-            contract <- liftIO $ fmap (fmap makeContractFromRPC) $ fetchContractWithSession conf sess block url addr
-            case contract of
-              Just x -> pure $ continue x
+            (rpcContract, _wasCached) <- liftIO $ fetchContractWithSession conf sess block url addr
```
blishko: Why give a name here, but not elsewhere?

```suggestion
            (rpcContract, _) <- liftIO $ fetchContractWithSession conf sess block url addr
```

blishko (commented):
I would like to suggest to use strong types more.
Instead of returning `Bool`, we should return a new type with two values, for example
`CachedResult | FreshResult` or `CachedFetch | FreshFetch`.
Ideally, the code should be self-documenting.

src/EVM/Fetch.hs
```diff
@@ -270,26 +277,46 @@ fetchWithSession url sess x = do
...
+          balance <- MaybeT $ fetch (QueryBalance addr)
+          pure $ RPCContract (ByteStringS code) nonce balance
+        case result of
+          Just contr | contr.code /= ByteStringS BS.empty -> do
+            -- Success: cache it
+            modifyMVar_ sess.sharedCache $ \c ->
+              pure $ c { contractCache = Map.insert addr contr c.contractCache }
+            pure (Just contr, False)  -- New fetch
+          Just _ -> do
+            -- Empty contract = "not found", cache this failure
```
blishko: I don't understand why failure is indicated by returning contract with empty code.
It feels like there should be a better way to indicate that the contract was not found.

src/EVM/Fetch.hs
gustavo-grieco: This was replaced by the usage of the new type

src/EVM/Fetch.hs
gustavo-grieco: This was replaced by the usage of the new type

gustavo-grieco:
Now it is using a proper algebraic type:
```haskell
data FetchStatus = Cached | Fresh
  deriving (Show, Eq)

data FetchResult a
  = FetchSuccess a FetchStatus
  | FetchFailure FetchStatus
  | FetchError Text
  deriving (Show, Eq)
 ```

msooseth:
I’ll check in 1h :)

PR #949, opened 2025-12-10
Encode symbolic power of 2 as bit-shift in SMT

gustavo-grieco:
## Description

This branch shows some code to handle symbolic exponent of pow, concretize them up to 255. It allows to prove the following code in seconds:
```solidity
contract PowTest {
  function nonNegative(uint256 exp) external pure {
    uint256 v = 2 ** exp; 
    assert(v >= 1 && v <= 2 ** 255);
  }
}
```

(you can play with the assert condition, adding `&& v != 2 ** 128` results in the correct counter example). If you like this approach, I can add some tests and polish it. Also worth mentioning that we can now fully verify the following [Algebra invariant](https://github.com/cryptoalgebra/Algebra/blob/integral-v1.2.2/src/core/contracts/test/echidna/BitMathEchidnaTest.sol) in seconds:
```solidity
contract BitMathEchidnaTest {
  function leastSignificantBitInvariant(uint256 input) external pure {
    unchecked {
      require(input > 0);
      uint8 lsb = TickTree.getSingleSignificantBit((0 - input) & input);
      assert(input & (uint256(2) ** lsb) != 0);
      assert(input & (uint256(2) ** lsb - 1) == 0);
    }
  }
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inequality propagation in constant propagation to prune impossible execution paths earlier.
   The constraint solver now tracks lower and upper bounds for symbolic values and detects
   conflicts (e.g., x < 5 && x > 10), significantly reducing the number of paths explored
-- Concretize symbolic EXP exponents up to 255 in SMT encoding
+- Encode symbolic power of 2 as bit-shift in SMT encoding.
 
 ## Fixed
 - Fix incorrect simplification rule for `PEq (Lit 1) (IsZero (LT a b))`


src/EVM/SMT.hs
```diff
@@ -666,6 +666,45 @@ expandExp base expnt
     n <- expandExp base (expnt - 1)
     pure $ "(bvmul " <> b `sp` n <> ")"
 
+-- | Unrolls a symbolic exponentiation into a series of multiplications
+expandSymbolicExp :: Expr EWord -> Expr EWord -> Err Builder
+expandSymbolicExp (Lit 2) expnt = do
+  eEnc <- exprToSMT expnt
+  pure $ "(bvshl " <> one `sp` eEnc <> ")"
```
blishko: I like this specialization for powers of two.

src/EVM/SMT.hs
```diff
@@ -666,6 +666,45 @@ expandExp base expnt
...
+            let bitCheck = "(= ((_ extract " <> Data.Text.Lazy.Builder.Int.decimal i `sp` Data.Text.Lazy.Builder.Int.decimal i <> ") e) #b1)"
+            in "(ite " <> bitCheck `sp` p `sp` one <> ")"
+
+         factors = map mkTerm terms
+         multiply [] = one
+         multiply [x] = x
+         multiply (x:xs) = "(bvmul " <> x `sp` multiply xs <> ")"
+       in multiply factors
+
+  pure $ "(let ((e " <> eEnc <> ")) " <> buildPowers 0 bEnc constructProduct <> ")"
```
blishko: It will take me some time to understand what is going on here 😅

src/EVM/SMT.hs
blishko: This special case can be handled directly in `exprToSMT`.

gustavo-grieco: Done!

src/EVM/SMT.hs
gustavo-grieco: I removed this part to include it in a future PR

src/EVM/SMT.hs
```diff
@@ -438,7 +438,11 @@ exprToSMT = \case
       pure $ "(ite (= " <> benc `sp` zero <> " ) " <> one `sp` zero <> ")"
     _ -> case b of
       Lit b' -> expandExp a b'
-      _ -> Left $ "Cannot encode symbolic exponent into SMT. Offending symbolic value: " <> show b
+      _ -> case a of
+        Lit 2 -> do
+          benc <- exprToSMT b
+          pure $ "(bvshl " <> one `sp` benc <> ")"
```
blishko, line 444: This belongs one level up, in the first `case a of`, between `Lit 0` and `_`

src/EVM/SMT.hs
```diff
@@ -666,6 +670,7 @@ expandExp base expnt
     n <- expandExp base (expnt - 1)
     pure $ "(bvmul " <> b `sp` n <> ")"
 
+
```
blishko: Don't add this extra empty line.

blishko (commented):
Can you also update the PR title and description?

Can we also include a test for this?

src/EVM/SMT.hs
gustavo-grieco: Are you sure that will keep the concretization correct when both values are concrete?

blishko: Ah, I see.
I have not thought about that.
But actually, I think it would be better. Currently, it unrolls multiplication, while now, if base is `2`, it will become a single shift.
Maybe we should turn this into a case on `(a,b)`.

We can also keep it as you are suggesting, and modify it later.

gustavo-grieco:
Should we include the 256 as well? if the code is using unsafe Solidity...

blishko:
What do you mean?

gustavo-grieco:
Oh, ignore my comment, only applies to the general case base != 2 😅 . I will add the missing description and test soon.

msooseth:
I did the following:
* added tests
* added one more rule:` go (Exp (Lit 2) (Lit k)) = shl (Lit 1) (Lit (k-1)) -- notice that k == 0 is handled above`
* refined the `shl` rule: when we shift by 256, it's always 0. Doesn't need to shift by MORE than 256. Shifting by 256 is enough.
* moved EXP rules to the same place
* added changelog

src/EVM/Expr.hs
```diff
@@ -1275,13 +1269,20 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
...
     -- NOTE: we can't simplify (Lit 0)^k. If k is 0 it's 1, otherwise it's 0.
     --       this is encoded in SMT.hs instead, via an SMT "ite"
-
+    --
+    -- A special pattern sometimes generated from Solidity that uses exponentiation to simulate bit shift.
+    -- We can rewrite the exponentiation into a bit-shift under certain conditions.
+    go (Exp (Lit 0x100) offset@(Mul (Lit a) (Mod _ (Lit b))))
+      | a * b <= 32 && (maxWord256 `Prelude.div` a) > b = shl (mul (Lit 8) offset) (Lit 1)
+    go (Exp (Lit 0x100) offset@(Mod _ (Lit 32))) = (shl (mul (Lit 8) offset)) (Lit 1)
+    go (Exp (Lit 2) (Lit k)) = shl (Lit 1) (Lit (k-1)) -- notice that k == 0 is handled above
```
blishko: It should be this, shouldn't it?
```suggestion
    go (Exp (Lit 2) (Lit k)) = shl (Lit 1) (Lit k)
```

msooseth:
Yes. You are right. Let me add some tests for that.

msooseth:
OK, I removed my additions to this PR and will create a new PR with my additions here: https://github.com/argotorg/hevm/pull/958

The reason is that I can't test this PR without fixes to `exp`, because haskell runs out of memory. I also can't test the SMT side, because it runs out of stack. 

First because `^` actually computes the actual value in memory and then truncates to mod 2^256. So it doesn't fit the memory.

The second (SMT) issue is because the `expandExp` could be called 2^256-1 times, so that's not gonna work.

msooseth:
OK, added tests and changelog

PR #958, opened 2026-01-07
Fix issues with EXP handling, add testing, add more rules

msooseth:
## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Encode symbolic power of 2 as bit-shift in SMT encoding.
 - Limit the expansion of the EXP operation to avoid blow-up in the size of the SMT expressions.
 - New EXP rewrite rule for base-2 exponents
-- Use `bvshl` in SMT for base-2 exponentiation so we can explore more.
 
 ## Fixed
 - Fix incorrect simplification rule for `PEq (Lit 1) (IsZero (LT a b))`


src/EVM/SMT.hs
```diff
@@ -432,12 +432,16 @@ exprToSMT = \case
   Sub a b -> op2 "bvsub" a b
   Mul a b -> op2 "bvmul" a b
   Exp a b -> case a of
-    -- Lit 1 has already been handled via Expr.simplify
     Lit 0 -> do
       benc <- exprToSMT b
       pure $ "(ite (= " <> benc `sp` zero <> " ) " <> one `sp` zero <> ")"
+    Lit 1 -> pure one
```
msooseth, line 438: Note: I added this here, because while `simplify` indeed handles special cases, we ALSO have fuzz-testing and I want to be able to fuzz-test, which does NOT use the `simplify` since we are testing if the `simplify` is correct :)

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The constraint solver now tracks lower and upper bounds for symbolic values and detects
   conflicts (e.g., x < 5 && x > 10), significantly reducing the number of paths explored
 - Encode symbolic power of 2 as bit-shift in SMT encoding.
+- Limit the expansion of the EXP operation to avoid blow-up in the size of the SMT expressions.
+- New EXP rewrite rule for base-2 exponents
+- Use `bvshl` in SMT for base-2 exponentiation so we can explore more.
```
blishko: I think this refers to the change from the previous PR, no?
It should be covered by `Encode symbolic power of 2 as bit-shift in SMT encoding.`

msooseth: OK, removing last line!

PR #962, opened 2026-01-11
[RFC] Alow to use yices as solver

gustavo-grieco:
## Description

Very experimental branch to allow to use `yices` as solver. There are some limitations in the way that memory is represented that must be fixed yet..

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -1119,7 +1119,7 @@ tests = testGroup "hevm"
         (Expr.indexWord (Lit 2) (Lit 0xff22bb4455667788990011223344556677889900112233445566778899001122))
     , test "encodeConcreteStore-overwrite" $
       assertEqualM ""
-        (pure "(store (store ((as const Storage) #x0000000000000000000000000000000000000000000000000000000000000000) (_ bv1 256) (_ bv2 256)) (_ bv3 256) (_ bv4 256))")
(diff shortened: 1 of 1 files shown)


blishko:
What are the changes to the SMT encoding and why are they needed?

I can try to resurrect my own work on integrating Yices.

gustavo-grieco:
This PR was made by Claude using halmos source code as a guide to understand how they use yices. It replaces the constant array by a normal array and force zero into all the read data.

gustavo-grieco:
If you think this is useful, feel free to take over/merge with your code.

blishko:
I am a bit skeptical about this, because it seems to me that the encoding might not be sound. It looks like a heuristic that might miss an assertion that for some index, the array must be zero.
I am afraid it would bite us in the future.
Designing more complex encoding in order to get around something that should really be solved at the level of the solver is not the way to go, in my opinion.

I have not seen compelling evidence that we need Yices, especially with Bitwuzla making excellent improvements in the last couple of years.

gustavo-grieco:
We can close this one, there could be some parts of the code that could be merged in a different context (e.g. the code to select which solver to use in testing).

PR #964, opened 2026-01-12
`Solvers.hs` should run separate processes for each query

msooseth:
## Description

This runs a separate process for each query and properly handles when the SMT solver dies. It also **strictly** enforces memory and time limits for the SMT solver. Hence, no more ooops-we-ran-longer issues, which plagued our system. We now get:

```
time cabal run --enable-shared -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_inv_identity --smttimeout 2
Configuration is affected by the following files:
- cabal.project
   Counterexample:   [validated]
     calldata: prove_inv_identity(-101034573947264812908054335791493761808)
   [FAIL] prove_inv_identity

Command exited with non-zero status 1
        Command being timed: "cabal run --enable-shared -f run exe:hevm -- test --root /home/matesoos/development/abdk-math-64.64-verification --max-buf-size 12 --max-iterations 2 --max-depth 70 --solver bitwuzla --match prove_inv_identity --smttimeout 2"
        User time (seconds): 8.03
        System time (seconds): 0.68
        Percent of CPU this job got: 359%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.42
```

Due to STRICT time enforcement.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [x] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Solvers.hs
+++ b/src/EVM/Solvers.hs
@@ -517,50 +517,37 @@ checkCommand inst cmd = do
 stripCarriageReturn :: Text -> Text
 stripCarriageReturn t = fromMaybe t $ T.stripSuffix "\r" t
 
+-- | Helper function for when we want to try an IO action and get Nothing on exception
+tryIO :: forall a. IO a -> MaybeT IO a
+tryIO action = MaybeT $ either (const Nothing) Just <$> (try action :: IO (Either IOException a))
+
 -- | Sends a string to the solver and appends a newline, returns the first available line from the output buffer
 -- Returns "unknown" if the solver process has died (timeout, crash, etc.)
 sendCommand :: SolverInstance -> SMTEntry -> IO Text
 sendCommand _ (SMTComment _) = internalError "Attempting to send a comment as a command to SMT solver"
 sendCommand (SolverInstance _ stdin stdout _) (SMTCommand cmd) = do
-  result <- try $ T.hPutStrLn stdin $ toLazyText cmd
-  case result of
-    Left (_ :: IOException) -> pure "unknown"  -- Write failed
-    Right _ -> do
-      flushResult <- try $ hFlush stdin
-      case flushResult of
-        Left (_ :: IOException) -> pure "unknown"  -- Flush failed
-        Right _ -> do
-          lineResult <- try $ stripCarriageReturn <$> T.hGetLine stdout
-          case lineResult of
-            Left (_ :: IOException) -> pure "unknown"  -- Process died
-            Right txt -> pure txt
+  result <- runMaybeT $ do
+    tryIO $ T.hPutStrLn stdin $ toLazyText cmd
+    tryIO $ hFlush stdin
+    tryIO $ stripCarriageReturn <$> T.hGetLine stdout
+  pure $ fromMaybe "unknown" result
 
 -- | Returns a string representation of the model for the requested variable
 -- Returns Nothing if the solver process has died (timeout, crash, etc.)
 getValue :: SolverInstance -> Text -> IO (Maybe Text)
-getValue (SolverInstance _ stdin stdout _) var = do
-  result <- try $ T.hPutStrLn stdin (T.append (T.append "(get-value (" var) "))")
-  case result of
-    Left (_ :: IOException) -> pure Nothing
-    Right _ -> do
-      flushResult <- try $ hFlush stdin
-      case flushResult of
-        Left (_ :: IOException) -> pure Nothing
-        Right _ -> runMaybeT $ do
-          val <- MaybeT $ readSExpr stdout
-          pure $ (T.unlines . reverse) val
+getValue (SolverInstance _ stdin stdout _) var = runMaybeT $ do
+  tryIO $ T.hPutStrLn stdin (T.append (T.append "(get-value (" var) "))")
+  tryIO $ hFlush stdin
+  val <- MaybeT $ readSExpr stdout
+  pure $ (T.unlines . reverse) val
 
 -- | Reads lines from h until we have a balanced sexpr
 -- Returns Nothing if the solver process has died or parsing fails
 readSExpr :: Handle -> IO (Maybe [Text])
-readSExpr h = do
-  result <- try $ go 0 0 []
-  case result of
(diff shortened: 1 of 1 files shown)


msooseth:
Windows can't do memory or time limiting of processes -- not easily.

Mac can only do time with `ulimit`. For memory, we'd need:
```haskell
import Foreign.C.Types (CLong(..))
import Foreign.Ptr (nullPtr)
import System.Posix.Process (forkProcess, executeFile)

foreign import ccall unsafe "sys/resource.h setrlimit"
    c_setrlimit :: CInt -> Ptr () -> IO CInt

-- | Set the memory limit (in KB) for the current process
setMemoryLimit :: Integer -> IO Bool
setMemoryLimit limitInKB = do
    -- Create an rlimit struct
    withArray [fromIntegral limitInKB, fromIntegral limitInKB] $ \ptr -> do
        -- resource = RLIMIT_RSS (value is 5 on macOS)
        result <- c_setrlimit 5 (castPtr ptr)
        return $ result == 0
```

But I'm not doing that, that looks very fragile.

Linux can reliable limit both.

msooseth:
@blishko I think FINALLY this is ready for review. Notice that the one failure we have correctly reports:
```
2026-01-15T13:24:17.8008102Z hevm>     simplify-joinbytes:                                           Error: SMT solver reported unsupported operation: (error "Array theory solver does not yet support write-chains connecting two different constant arrays")
2026-01-15T13:24:17.8031374Z hevm> FAIL (4.86s)
```

And all other types of failures that are due to killing of the SMT solver mid-way a response etc. are correctly ignored.

src/EVM/Solvers.hs
```diff
@@ -361,47 +384,77 @@ getModel inst cexvars = do
           -- TODO: do I need to check the write idx here?
           (Write _ idx next) -> idx <= 32 && go (Comp next)
 
-mkTimeout :: Maybe Natural -> Text
-mkTimeout t = T.pack $ show $ (1000 *)$ case t of
-  Nothing -> 300 :: Natural
-  Just t' -> t'
+mkTimeout :: Maybe Natural -> Natural
```
blishko, line 403: Can you add a comment explaining which unit this is? Seconds?

src/EVM/Solvers.hs
```diff
@@ -361,47 +384,77 @@ getModel inst cexvars = do
...
+spawnSolver :: Solver -> Maybe (Natural) -> Natural -> IO SolverInstance
+#if defined(mingw32_HOST_OS)
+spawnSolver solver _ _ = do
   (readout, writeout) <- createPipe
-  let cmd
-        = (proc (show solver) (fmap T.unpack $ solverArgs solver threads timeout ))
+  let solverCmd = show solver
+      solverArgsStr = fmap T.unpack $ solverArgs solver
+      -- Windows: no ulimit available
+      cmd = (proc solverCmd solverArgsStr)
```
blishko, line 440: I think we should still pass at least the timeout to the solver, if it has a command-line flag for that.

src/EVM/Solvers.hs
msooseth: Yep, seconds, sorry! Fixed :)

src/EVM/Solvers.hs
msooseth: Yes, fair. I just did that now :)

msooseth:
I now fixed those two things, they were good points, to be honest. Also, I realized that `--smtdebug` should be `--smt-debug`, so I fixed that too. Change is at d22163403c4278ad9dff329d9d3bfa49f35aab31 so it may be easier to review.

Note: UPDATED, due to solvers needing millisecs :)

src/EVM/Solvers.hs
```diff
@@ -435,36 +492,56 @@ sendScript solver (SMTScript entries) = do
...
+        Right _ -> fmap (T.unlines . reverse) (readSExpr stdout)
 
 -- | Reads lines from h until we have a balanced sexpr
+-- Returns ["unknown"] if the solver process has died
 readSExpr :: Handle -> IO [Text]
-readSExpr h = go 0 0 []
+readSExpr h = do
+  result <- try $ go 0 0 []
+  case result of
+    Left (_ :: IOException) -> pure ["unknown"]  -- Solver died (timeout, crash, etc.)
```
blishko: I am not sure returning `unknown` is the proper way.
I think it would be better to propagate the error somehow (if we don't want to just let the caller's handle the exception).

test/rpc.hs
```diff
@@ -86,7 +86,7 @@ tests = testGroup "rpc"
           calldata' = ConcreteBuf $ abiMethod "transfer(address,uint256)" (AbiTuple (V.fromList [AbiAddress (Addr 0xdead), AbiUInt 256 wad]))
         sess <- mkSessionWithoutCache
         vm <- weth9VM sess testBlockNumber (calldata', [])
-        postVm <- withSolvers Z3 1 1 Nothing $ \solvers ->
+        postVm <- withSolvers Z3 1 Nothing 1024 $ \solvers ->
```
blishko: Could we introduce a name to the default memory limit?
It will be much easier to change it later if we decide to do it.

msooseth: OK, fair, this has now been fixed :)

msooseth:
OOopps, sorry about the MacOS issue. Also, I now made `Solvers.hs` into a module, so only some things are exported. This means I had to remove the namings of some of the structure's elements, otherwise they'd be reported as unused. I think it's actually cleaner this way anyway.

src/EVM/Solvers.hs
msooseth: OOps, OK, new commit is a bit bigger, but fixes it: c826b82d01452e726dbf2da4aa3eec00dd2bb71e

msooseth:
OK, I tried to make the handling of `IOException` a lot nicer via 7ba2ee5f17d04f84eae18bef79087d44211151fa It actually removes code, which is nice :)

PR #965, opened 2026-01-12
`--early-abort` flag to abort symbolic execution *only*

msooseth:
## Description

This aborts the symbolic execution ONLY (i.e. NOT the SMT solving) when a CEX is found. I'll do another PR to kill the threads once the Solver.hs PR has been merged related to process management.

TODO: add test

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -257,8 +257,7 @@ common test-base
     optics-core,
     witch,
     unliftio-core,
-    exceptions,
-    stm
+    exceptions
 
 library test-utils
   import: test-base


msooseth:
I don't see how to test this, so I'm marking this ready for review. We can test this properly once we have SMT solvers in separate processes, and we can kill them all.

hevm.cabal
```diff
@@ -266,6 +266,8 @@ library test-utils
     EVM.Test.Utils
     EVM.Test.FuzzSymExec
     EVM.Test.BlockchainTests
+  build-depends:
```
blishko: Just add this to `test-base` instead of adding it to both `test-utils` and `test-common`.
And is it really needed?

src/EVM/SymExec.hs
```diff
@@ -395,29 +398,37 @@ interpret fetcher iterConf vm stepper handler = do
...
       _ <- liftIO $ readChan avail
       task <- liftIO $ readChan taskQ
-      runTask' <- toIO $ getOneExpr task avail processQ numTasks numProcs
-      liftIO $ forkIO runTask'
+      abortFlag <- liftIO $ atomically $ readTVar shouldAbort
+      if abortFlag
+        then liftIO $ writeChan avail ()
+        else do
+          runTask' <- toIO $ getOneExpr task avail processQ numTasks numProcs
+          void $ liftIO $ forkIO runTask'
```
blishko, line 411: Why do you need `void` here?

src/EVM/SymExec.hs
```diff
@@ -483,19 +494,25 @@ interpretInternal t@InterpTask{..} = eval (Operational.view stepper)
...
         conf <- readConfig
         frozen <- liftIO $ stToIO $ freezeVM vm
         let newDepth = vm.exploreDepth+1
         (ra, vma) <- liftIO $ stToIO $ runStateT (continue True) frozen { result = Nothing, exploreDepth = newDepth }
-        liftIO $ atomically $ modifyTVar numTasks (+1)
-        liftIO $ writeChan taskQ $ t { vm = vma, stepper = (k ra) }
-        when (conf.debug && conf.verb >= 2) $ liftIO $ putStrLn $ "Queued new task for Fork at depth " <> show newDepth
+        -- Check abort flag before queuing new task
+        abortFlag <- liftIO $ atomically $ readTVar shouldAbort
+        unless (conf.earlyAbort && abortFlag) $ do
```
blishko: I don't think we need to consult the config here, no?
Checking the flag should be sufficient.
Same for line 499.

src/EVM/SymExec.hs
```diff
@@ -338,10 +339,11 @@ interpret :: forall m a . App m
   => Fetch.Fetcher Symbolic m
   -> IterConfig
   -> VM Symbolic
+  -> TVar Bool
   -> Stepper Symbolic (Expr End)
   -> (Expr End -> m a)
   -> m [a]
-interpret fetcher iterConf vm stepper handler = do
+interpret fetcher iterConf vm shouldAbort stepper handler = do
```
blishko: Is there a reason why `interpret` should take the `TVar` as an argument?
Wouldn't it be better to create it here?

src/EVM/SymExec.hs
msooseth: Before, it returned `b`, now it returns void, i.e. `()`. We don't need to return anything from here. The thread will be forked with `forkIO`, and that `threadId` will be killed later:

```haskell
  taskOrchestrate' <- toIO $ taskOrchestrate taskQ availableInstances processQ numTasks numProcs
  taskOrchestrateId <- liftIO $ forkIO taskOrchestrate'
```

It's a bit mind-bending, that we return an `IO ()` and that gets passed to `forkIO`... but that's how it seems to work. Weird. Note that we can't return the `forkIO runTask` because that's the thread of the **taks**  not the **orchestrator**. Annoying...

src/EVM/SymExec.hs
msooseth: Aaaaa.... good point! Fixed both :)

src/EVM/SymExec.hs
msooseth: Good idea. `verifyInputsWithHandler` launches an `interpret` (through `executeVM`):

```
  results <- executeVM fetcher opts.iterConf preState shouldAbort $ \leaf -> do
```

And this function, that is passed `leaf` is the one that sets the `shouldAbort`:
```
            when conf.earlyAbort $ liftIO $ atomically $ writeTVar shouldAbort True
```
 So I now added `shouldAbort` as a parameter to that function. Which means that instead of `pure` we now need `(pure . pure)` when we just need the list of `Expr End`-s. Should it be `(leaf, shouldAbort)`? Then it's only `pure`... but I think that's not very functional.

hevm.cabal
msooseth: Good point, done!

src/EVM/SymExec.hs
blishko: Oh, I missed that!
Let me think about it. Though it looks reasonable now.

msooseth: I actually like that `interpret` is more standalone this way...

hevm.cabal
```diff
@@ -257,7 +257,8 @@ common test-base
     optics-core,
     witch,
     unliftio-core,
-    exceptions
+    exceptions,
+    stm
```
blishko: This should no longer be needed, actually.

blishko (approved):
LGTM!
I have an idea for some clarification, I'll put that as a new PR.
Also, I think we no longer need to add the new dependency.

hevm.cabal
msooseth: OK, I removed it, tested it, and now I'm squashing and merging!

PR #966, opened 2026-01-12
CI: Add build on arm linux

blishko:
## Description

We want to be able to release executable for arm linux as well, so we should build hevm in CI for this architecture.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,14 +13,13 @@ jobs:
   nix-build:
     strategy:
       matrix:
-        runtime: [ linux-x64, osx-x64, osx-arm64 ]
         include:
         - runtime: linux-x64
           os: ubuntu-latest
-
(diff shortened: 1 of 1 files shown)


msooseth:
Yep, this `       error: attribute 'solc_0_8_26' missing` was the issue I bumped into the last time I tried :S

blishko:
It's because the overlay for solc binaries that we use does not export linux-arm binaries (likely because `solc` was not releasing those prior to `0.8.31`).
I have opened an issue and a prototype PR for that: https://github.com/hellwolf/solc.nix/pull/26

My plan is to wait a few days to see if things get moving over there.
Otherwise, we might consider forking the repository and maintaining it ourselves.

PR #968, opened 2026-01-16
SymEx: Introduce a type synonym and noop helper for path handler

blishko:
## Description

The type synonym and helper will enable making change in one place instead of updating type and/or usage of noop handler at every location they are used.
It will also hopefully make it a bit more clear what this type represents.

This is a follow up on #965.


msooseth (approved):
LGTM

PR #971, opened 2026-01-20
Implement etch() cheatcode + Support Bytes dynamic type

### LINKED ISSUE #970: Support for etch cheatcode as per Foundry and Medusa
When running echidna, we may need to etch some bytecode into a specific address

This is useful for 2 reasons:
- Deterministic deployments / code
- Macros to perform storage extractions

## Example use case

```solidity
    /// @notice Deploys MockUniswapV2Factory and etches it to the hardcoded address
    /// @dev Required for createLaunchPad to work without mainnet fork
    function _deployMockUniswapFactory() internal {
        // Deploy the mock factory
        MockUniswapV2Factory mockFactory = new MockUniswapV2Factory();
        
        // Get the bytecode of the deployed mock
        bytes memory factoryCode = address(mockFactory).code;
        
        // Etch the mock factory code to the hardcoded Uniswap V2 factory address
        vm.etch(LaunchPadUtils.UNISWAP_POOL_FACTORY_, factoryCode);
        
        // Copy storage from mock factory to the etched address (for allPairs array initialization)
        // Note: The factory starts with empty state, which is correct for testing
    }

address public constant UNISWAP_POOL_FACTORY_ = 0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f
```

## Error

Image

## Request

Implement vm.etch to be equivalent with Foundry and Medusa

msooseth:
## Description
Fixes #970 

TENTATIVE, needs test case provided by @GalloDaSballo to make sure we are doing the right thing.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/ABI.hs
+++ b/src/EVM/ABI.hs
@@ -660,7 +660,8 @@ instance Arbitrary AbiValue where
     AbiBytesDynamic b -> AbiBytesDynamic . BS.pack <$> shrinkList shrinkIntegral (BS.unpack b)
     AbiString b -> AbiString . BS.pack <$> shrinkList shrinkIntegral (BS.unpack b)
     AbiBytes n a | n <= 32 -> shrink $ AbiUInt (n * 8) (word256 a)
-    AbiBytes n a -> AbiBytes n . BS.pack <$> shrinkList shrinkIntegral (BS.unpack a)
+    --bytesN for N > 32 don't really exist right now anyway..
+    AbiBytes _ _ | otherwise -> []
     AbiArray _ t v ->
       Vector.toList v ++
         map (\x -> AbiArray (length x) t (Vector.fromList x))
@@ -669,7 +670,7 @@ instance Arbitrary AbiValue where
     AbiUInt n a -> AbiUInt n <$> (shrinkIntegral a)
     AbiInt n a -> AbiInt n <$> (shrinkIntegral a)
     AbiBool b -> AbiBool <$> shrink b
-    AbiAddress a -> [AbiAddress 0xacab, AbiAddress a, AbiAddress 0xbabeface]
+    AbiAddress a -> [AbiAddress 0xacab, AbiAddress a, AbiAddress 0xbabeface, AbiAddress 0xdeadbeef]
     AbiFunction b -> shrink $ AbiBytes 24 b
 
 -- A modification of 'arbitrarySizedBoundedIntegral' quickcheck library


GalloDaSballo:
Happy new year @msooseth!
Thank you for the prompt response.

Here's a sample repo you can use for testing:
https://github.com/GalloDaSballo/chimera-etch-test

forge test --match-test test_crytic -vvvvv --decode-internal - Works
medusa fuzz - Works
echidna . --contract CryticTester --config echidna.yaml --format text --workers 16 --test-limit 1000000 - Fails

msooseth:
Hi,

I see, thanks! Unfortunately, testing this through Echidna is not really what I meant. I'll try to write a test case so I can have it inside hevm, to make sure we don't accidentally break this functionality in the future. What I meant was like a 10-line Solidity code to check :) But I have an idea how to write one. Thanks again,

Mate

msooseth:
As always, this looks simpler than it is :) We cannot currently decode dynamic calldata. So I'm working on that. Kind of a rabbit hole, but maybe there is a light at the end of the tunnel...

msooseth:
@GalloDaSballo here's a test case, let me know if this is what you think should work:

```solidity
            interface Vm {
              function etch(address,bytes calldata) external;
            }
            contract A {
              function getX() external returns (uint) {
                  return 10;
              }
            }
            contract B {
              function getX() external returns (uint) {
                  return 20;
              }
            }
            contract C {
              A a;
              B b;
              function fun() external {
                  a = new A();
                  b = new B();
                  Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);

                  assert(a.getX() == 10);
                  vm.etch(address(a), address(b).code);
                  assert(a.getX() == 20);
              }
            }
```

The above now succeeds. With `assert(a.getX() == 10);` on both sides of the `etch` it fails.

GalloDaSballo:
> @GalloDaSballo here's a test case, let me know if this is what you think should work:
> 
> ```solidity
>             interface Vm {
>               function etch(address,bytes calldata) external;
>             }
>             contract A {
>               function getX() external returns (uint) {
>                   return 10;
>               }
>             }
>             contract B {
>               function getX() external returns (uint) {
>                   return 20;
>               }
>             }
>             contract C {
>               A a;
>               B b;
>               function fun() external {
>                   a = new A();
>                   b = new B();
>                   Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
> 
>                   assert(a.getX() == 10);
>                   vm.etch(address(a), address(b).code);
>                   assert(a.getX() == 20);
>               }
>             }
> ```
> 
> The above now succeeds. With `assert(a.getX() == 10);` on both sides of the `etch` it fails.

LG. 
I think you can also assert that the bytecode length and content matches, ideally symbolically.

msooseth:
> > The above now succeeds. With `assert(a.getX() == 10);` on both sides of the `etch` it fails.
> 
> LG. I think you can also assert that the bytecode length and content matches, ideally symbolically.

Thanks, good points. I added these checks now:

```solidity
            interface Vm {
              function etch(address,bytes calldata) external;
            }
            contract A {
              function getX() external returns (uint) {
                  return 10;
              }
            }
            contract B {
              function getX() external returns (uint) {
                  return 20;
              }
            }
            contract C {
              A a;
              B b;
              function fun() external {
                  a = new A();
                  b = new B();
                  bytes memory bCode = address(b).code;
                  Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);

                  assert(a.getX() == 10);
                  vm.etch(address(a), bCode);
                  assert(address(a).code.length == address(b).code.length);
                  bytes memory aCode = address(a).code;
                  assert(keccak256(aCode) == keccak256(bCode));
                  assert(a.getX() == 20);
              }
            }
```

GalloDaSballo:
I believe that this ensures the feature is correct for all EOAs / Normal Contracts.

Perhaps worth exploring what happens when a precompile is etched? I'm not sure I have ever used this for foundry / echidna, but I think this can lead to edge cases.

src/EVM.hs
```diff
@@ -1977,6 +1977,15 @@ cheatActions = Map.fromList
...
         _ -> vmError (BadCheatCode "setEnv(string,string) parameter decoding failed" sig)
 
+  , action "etch(address,bytes)" $
+      \sig input ->  case decodeBuf [AbiAddressType, AbiBytesDynamicType] input of
+          (CAbi valsArr,"") -> case valsArr of
+            [AbiAddress addr, AbiBytesDynamic bytes] -> fetchAccount addr $ \_ -> do
+                replaceCodeEtch addr (RuntimeCode (ConcreteRuntimeCode bytes))
+                doStop
+            _ -> vmError (BadCheatCode "etch(address,bytes) address decoding failed" sig)
+          _ -> vmError (BadCheatCode "etch(address,bytes) address decoding failed" sig)
```
blishko: If we want the same error message, we can combine the two cases into one, no?

```suggestion
      \sig input ->  case decodeBuf [AbiAddressType, AbiBytesDynamicType] input of
          (CAbi [AbiAddress addr, AbiBytesDynamic bytes],"") -> fetchAccount addr $ \_ -> do
                replaceCodeEtch addr (RuntimeCode (ConcreteRuntimeCode bytes))
                doStop
          _ -> vmError (BadCheatCode "etch(address,bytes) address decoding failed" sig)
```

src/EVM/ABI.hs
```diff
@@ -593,8 +665,7 @@ instance Arbitrary AbiValue where
     AbiBytesDynamic b -> AbiBytesDynamic . BS.pack <$> shrinkList shrinkIntegral (BS.unpack b)
     AbiString b -> AbiString . BS.pack <$> shrinkList shrinkIntegral (BS.unpack b)
     AbiBytes n a | n <= 32 -> shrink $ AbiUInt (n * 8) (word256 a)
-    --bytesN for N > 32 don't really exist right now anyway..
-    AbiBytes _ _ | otherwise -> []
+    AbiBytes n a -> AbiBytes n . BS.pack <$> shrinkList shrinkIntegral (BS.unpack a)
```
msooseth: This seems more thorough. Also, we now DO support bytes that are more than 32 long.

src/EVM/ABI.hs
```diff
@@ -603,8 +674,7 @@ instance Arbitrary AbiValue where
     AbiUInt n a -> AbiUInt n <$> (shrinkIntegral a)
     AbiInt n a -> AbiInt n <$> (shrinkIntegral a)
     AbiBool b -> AbiBool <$> shrink b
-    AbiAddress a -> [AbiAddress 0xacab, AbiAddress 0xdeadbeef, AbiAddress 0xbabeface]
-      <> (AbiAddress <$> shrinkIntegral a)
+    AbiAddress a -> [AbiAddress 0xacab, AbiAddress a, AbiAddress 0xbabeface]
```
msooseth: I am surprised that `a` was not used in the RHS of this. I fixed it.

msooseth:
This is now ready for review. I tried to make the change as small as possible. The system now ONLY works with Foundry and on addresses created via `setUp`. Does not work with addresses created during symbolic execution, only during the concrete, i.e. `setUp` (or construction) phases. I believe this should still be good enough for @GalloDaSballo as it will work with e.g.:

```solidity
contract D is Test {
  A a;
  B b;
  function setUp() public {
      a = new A();
      b = new B();
  }
  function prove_etch_fail() external {
      bytes memory bCode = address(b).code;
      Vm vm = Vm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);

      assert(a.getX() == 10);
      vm.etch(address(a), bCode);
      assert(a.getX() == 10);
  }
}
```

Which is what I am suspecting all fuzzing systems do -- i.e. set up the environment in `setUp()` rather than during the run.

src/EVM/ABI.hs
blishko: I think this represents Solidity's `bytesN` type and `bytes32` is the largest one.
So I think the comment was correct.

src/EVM/ABI.hs
blishko: Maybe we can also keep the `deadbeef`?

msooseth:
@blishko  Aaaa... good point about the BytesN. I'll fix that.  I'll also add back the `0xdeadbeef`

msooseth:
OK, maybe you can check https://github.com/argotorg/hevm/pull/971/commits/2cf60fcec421e1428b6aa276db266c1c4782c7ef and then it can perhaps be merged :)

Issue #763, opened 2025-06-12
Branch explosion due to missing state merging


gustavo-grieco:
hevm takes a very long time to explore the following function (which despite how it looks, it is used frequently when dealing with ticks on dexes):

```solidity
  function getSqrtRatioAtTick(int24 tick) internal pure returns (uint160 price) {
    unchecked {
      // get abs value
      int24 absTickMask = tick >> (24 - 1);
      uint256 absTick = uint24((tick + absTickMask) ^ absTickMask);
      if (absTick > uint24(MAX_TICK)) revert IAlgebraPoolErrors.tickOutOfRange();

      uint256 ratio = 0x100000000000000000000000000000000;
      if (absTick & 0x1 != 0) ratio = 0xfffcb933bd6fad37aa2d162d1a594001;
      if (absTick & 0x2 != 0) ratio = (ratio * 0xfff97272373d413259a46990580e213a) >> 128;
      if (absTick & 0x4 != 0) ratio = (ratio * 0xfff2e50f5f656932ef12357cf3c7fdcc) >> 128;
      if (absTick & 0x8 != 0) ratio = (ratio * 0xffe5caca7e10e4e61c3624eaa0941cd0) >> 128;
      if (absTick & 0x10 != 0) ratio = (ratio * 0xffcb9843d60f6159c9db58835c926644) >> 128;
      if (absTick & 0x20 != 0) ratio = (ratio * 0xff973b41fa98c081472e6896dfb254c0) >> 128;
      if (absTick & 0x40 != 0) ratio = (ratio * 0xff2ea16466c96a3843ec78b326b52861) >> 128;
      if (absTick & 0x80 != 0) ratio = (ratio * 0xfe5dee046a99a2a811c461f1969c3053) >> 128;
      if (absTick & 0x100 != 0) ratio = (ratio * 0xfcbe86c7900a88aedcffc83b479aa3a4) >> 128;
      if (absTick & 0x200 != 0) ratio = (ratio * 0xf987a7253ac413176f2b074cf7815e54) >> 128;
      if (absTick & 0x400 != 0) ratio = (ratio * 0xf3392b0822b70005940c7a398e4b70f3) >> 128;
      if (absTick & 0x800 != 0) ratio = (ratio * 0xe7159475a2c29b7443b29c7fa6e889d9) >> 128;
      if (absTick & 0x1000 != 0) ratio = (ratio * 0xd097f3bdfd2022b8845ad8f792aa5825) >> 128;
[...]

Issue #719, opened 2025-04-29
Experiment with state merging


gustavo-grieco:
Manticore used to [do state merging when performing symbolic execution](https://blog.trailofbits.com/2019/01/25/symbolic-path-merging-in-manticore/). While this was never implemented for EVM, it could be a good idea to consider it for hevm.

msooseth:
While on paper this may be interesting, I think this is currently not the most important thing to work on. I'd like to make sure we can deal with loops, recursion, and unknown code (i.e. `CALL` to transfer funds) at least to _some_ extent before we start making solving faster. I find that solving faster is not the most important issue right now. Most of the time, we can just throw more CPU at the problem. And if the problem is unsolvable, because of complicated maths (mod, mul, etc over 256b integers) then state merging won't help anyway.

gustavo-grieco:
We already have this merged, closing this.

PR #973, opened 2026-01-20
State merging

gustavo-grieco:
## Description

This branch contains state merging, which allows to solve/mitigate issues #719 and #763. This is a minimal version of this PR. It introduces:
- State merging with forward merging only, and no branching within speculative execution (i.e. no nested execution). Whenever an RPC/SMT call occurs, or a revert/error, we abandon speculation immediately.
- Makes source code lookup more generic, so we can print source lines at any point, including during merging. This means we can get outputs as below.
- This version has a slightly cleaner set of comments, which may make it easier to grasp what's going on.

 To test, use [abdk-math-64.64-verification repository](https://github.com/gustavo-grieco/abdk-math-64.64-verification). Just clone it and execute `T=prove_pow_base_one make verify-hevm`. If you do this without state merging, it will never finish in practice (unclear how long it can take).

## Example output
```
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 436 : "result >= MIN_64x64 && result <= MAX_64x64"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 382 : "if (y & 0x2 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 387 : "if (y & 0x4 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 392 : "if (y & 0x8 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 436 : "result >= MIN_64x64 && result <= MAX_64x64"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 377 : "if (y & 0x1 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 382 : "if (y & 0x2 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 387 : "if (y & 0x4 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 392 : "if (y & 0x8 != 0) {\n            absResult = absResult * absX >> 127;\n          }"
Merged forward jump in file "lib/abdk-libraries-solidity/ABDKMath64x64.sol" on line 436 : "result >= MIN_64x64 && result <= MAX_64x64"
Parsed contract: "ABDKMath64x64" file: /home/matesoos/development/abdk-math-64.64-verification/out/ABDKMath64x64.sol/ABDKMath64x64.json
Parsed contract: "CryticABDKMath64x64Properties" file: /home/matesoos/development/abdk-math-64.64-verification/out/ABDKMath64x64PropertyTests.sol/CryticABDKMath64x64Properties.json
Found 1 unit test contract(s) to test:
  --> src/ABDKMath64x64PropertyTests.sol:CryticABDKMath64x64Properties  ---  functions: [Sig "prove_pow_base_one" [uint256]]

[RUNNING] prove_pow_base_one(uint256)
   Keccak preimages in state: 0
   Exploring call prefix 0xd185a620
Interpretation finished, collected 66 results.
   Exploration and solving finished, 66 branch(es) checked in call prefix 0xd185a620 of which partial: 66
   Found 0 potential counterexample(s) in call prefix 0xd185a620
   [EXPLORATION COMPLETE] prove_pow_base_one -- explored 66 paths.
   symRun -- (cex,warnings,unexpectedAllRevert): (False,False,False)
   [PASS] prove_pow_base_one

unitTest individual results: [(True,True)]
```

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -3993,8 +3993,9 @@ tests = testGroup "hevm"
           e <- withDefaultSolver $ \s -> getExpr s c sig [] defaultVeriOpts
           pure $ length e
         assertBoolM "Merging should reduce number of paths" (merge < noMerege)
+     -- Checked arithmetic reverts in one of the branches, which is not supported by
+     -- the current state merging implementation
      , expectFail $ testCase "merge-simple-branches-revert" $ do
-        -- This cannot be merged due to revert unfortunately.
         Just c <- solcRuntime "C"
           [i|
           contract C {
@@ -4051,20 +4052,11 @@ tests = testGroup "hevm"
           }
           |]
         let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
-        (paths, []) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        paths <- withDefaultSolver $ \s -> getExpr s c sig [] defaultVeriOpts
         let numPaths = length paths
         -- Without merging: 16 paths. With unchecked + merging: should be <= 4
         liftIO $ assertBool ("Expected at most 4 paths with unchecked merging, got " ++ show numPaths) (numPaths <= 4)
-        putStrLnM $ "Many branches merged: " ++ show numPaths ++ " paths (vs 16 without merging)"
-    -- TODO: Checked arithmetic merging doesn't work well with solcRuntime's standard-json output
-    -- The bytecode structure differs from direct solc --bin-runtime compilation
-    -- Direct compile: 4 merges, 4 paths. Standard-json: 0 merges, 18 paths
-    -- This needs investigation into why standard-json produces different bytecode structure
-    -- , test "merge-many-branches-checked" $ do ...
     , test "merge-with-unchecked-arithmetic" $ do
-        -- Unchecked arithmetic should merge cleanly (no overflow check JUMPIs)
-        -- 4 branches = 2^4 = 16 paths without merging
-        -- With unchecked + merging, should be minimal paths (ideally 2)
         Just c <- solcRuntime "C"
           [i|
           contract C {
@@ -4085,7 +4077,6 @@ tests = testGroup "hevm"
         let numPaths = length paths
         -- Without merging: 16 paths. With unchecked + merging: should be <= 4
         liftIO $ assertBool ("Expected at most 4 paths with unchecked merging, got " ++ show numPaths) (numPaths <= 4)
-        putStrLnM $ "Unchecked arithmetic merged: " ++ show numPaths ++ " paths (vs 16 without merging)"
     , test "merge-counterexample-three-branches" $ do
         -- Find counterexample with 3 merged branches
         Just c <- solcRuntime "C"
@@ -4102,8 +4093,8 @@ tests = testGroup "hevm"
           }
           |]
         let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
-        (_, [Cex _]) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
-        putStrLnM "Counterexample found with three merged branches"
+        (_, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        assertBoolM "Expected a counterexample with merged branches" (exactlyCex 1 ret)
     , test "no-false-positive-nested-branches" $ do
         -- Verify that nested branches don't cause false positives
         -- This tests soundness: if both paths of a nested branch don't converge
@@ -4168,8 +4159,8 @@ tests = testGroup "hevm"
           }
           |]
         let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
-        (_, []) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
-        putStrLnM "Zero-mul loop merging works without expression blowup"
+        (_, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        assertEqualM "Zero-mul loop merging works without expression blowup" [] ret
     , test "no-merge-with-memory-write" $ do
         -- Branches with memory writes should not cause issues
         Just c <- solcRuntime "C"
@@ -4186,8 +4177,8 @@ tests = testGroup "hevm"
           }
           |]
         let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
-        (_, []) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
-        putStrLnM "Memory write branches handled correctly"
+        (_, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        assertEqualM "Merging should not cause issues with memory writes" [] ret
     ]
   , testGroup "simple-checks"
     [ test "simple-stores" $ do


src/EVM/Expr.hs
```diff
@@ -1053,6 +1053,14 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
     go (IsZero (Xor x y)) = eq x y
     go (IsZero a) = iszero a
 
+    -- ITE (if-then-else) simplification for path merging
+    go (ITE (Lit 0) _ f) = f                        -- false condition: take else branch
+    go (ITE (Lit _) t _) = t                        -- true condition (non-zero): take then branch
+    go (ITE _ t f) | t == f = t                     -- branches identical: eliminate ITE
+    go (ITE c (ITE c' a _) d) | c == c' = ITE c a d -- nested same condition in then
+    go (ITE c a (ITE c' _ d)) | c == c' = ITE c a d -- nested same condition in else
+    go (ITE c t f) = ITE c t f                      -- no simplification possible
```
msooseth: I think this last line is not needed

src/EVM/Merge.hs
```diff
@@ -0,0 +1,250 @@
...
+              if length trueStack == length falseStack && soundnessOK
+                then do
+                  -- Merge stacks using ITE expressions
+                  let condSimp = Expr.simplify cond
+                      mergeExpr t f
+                        | t == f    = t
+                        | otherwise = ITE condSimp t f
+                      mergedStack = zipWith mergeExpr trueStack falseStack
+                  -- Use vm0 as base and update only PC and stack
+                  when conf.debug $ traceM $ "Merged forward jump at PC " ++ show jumpTarget
```
msooseth: Hmmm, the `debug $ traceM` is actually kinda neat, even though I prefer no `traceM` in real code, but in debug mode I think it's OK.

src/EVM/Expr.hs
msooseth: In other words, `go (ITE c t f) = ITE c t f` does nothing, and is not necessary for our simplifier.

src/EVM.hs
```diff
@@ -813,14 +815,51 @@ exec1 conf = do
 
         OpJumpi -> {-# SCC "OpJumpi" #-}
           case stk of
-            x:y:xs -> forceConcreteLimitSz x 2 "JUMPI: symbolic jumpdest" $ \x' ->
+            x:y:xs ->
+              forceConcreteLimitSz x 2 "JUMPI: symbolic jumpdest" $ \x' ->
```
msooseth: Cutting this into two lines is unnecessary, and generates noise

src/EVM.hs
```diff
@@ -813,14 +815,51 @@ exec1 conf = do
...
+                        if alreadyMerging
+                          then do
+                            -- Inside speculation: try both paths using budget-based exploration
+                            result <- Merge.exploreNestedBranch conf (exec1 conf) jumpTarget ms.msTargetPC y xs
+                            case result of
+                              Just vmFinal -> put vmFinal
+                              Nothing -> do
+                                -- Neither path converged - signal failure by setting a result
+                                -- This will cause speculateLoop to return Nothing
+                                assign #result $ Just $ VMFailure BadJumpDestination
```
msooseth: Hmmm can't the `result` be also assigned by the `exec1`? Maybe it'd be safer to use something other than `result`?

src/EVM/Merge.hs
```diff
@@ -0,0 +1,250 @@
...
+      , msRemainingBudget = budget
+      , msNestingDepth = 0
+      }
+    result <- speculateLoop conf exec1Step targetPC
+    -- Reset merge state
+    modifying #mergeState $ \ms -> ms { msActive = False }
+    return result
+
+-- | Inner loop for speculative execution with budget tracking
+speculateLoop
```
msooseth, line 47: I don't think "loop" is the right name for this? It's a bit misleading.

src/EVM.hs
gustavo-grieco: Fixed

src/EVM/Merge.hs
gustavo-grieco: Renamed

src/EVM.hs
gustavo-grieco: Fixed

src/EVM/Merge.hs
gustavo-grieco: I think we need this for experimentation, but we can think in a better way to do it..

src/EVM/Expr.hs
gustavo-grieco: Fixed

gustavo-grieco:
@msooseth this branch was rebase with master. I also added the option to disable state merging using `--merge-max-budget 0`. We can use that value as default for the next release if we are not sure about it.

gustavo-grieco:
This is already rebased and the test are passing 🎉

src/EVM.hs
```diff
@@ -866,14 +868,51 @@ exec1 conf = do
 
         OpJumpi -> {-# SCC "OpJumpi" #-}
           case stk of
-            x:y:xs -> forceConcreteLimitSz x 2 "JUMPI: symbolic jumpdest" $ \x' ->
+            x:y:xs ->
+              forceConcreteLimitSz x 2 "JUMPI: symbolic jumpdest" $ \x' ->
               burn g_high $
+                -- Note: We must NOT check tryInto x' unconditionally!
+                -- Per EVM semantics, if condition is 0, we just fall through
+                -- without validating the destination. Only validate when jumping.
```
msooseth: BTW, this is a really good point. We were wrong here! Like, current hevm is incorrect in symbolically executing. Wow.

gustavo-grieco: Let's preserve this in an issue so we can fix it later

src/EVM/Merge.hs
```diff
@@ -0,0 +1,144 @@
...
+                soundnessOK = checkNoSideEffects vm0 vmFalse
+
+            -- Check merge conditions: same stack depth AND no side effects
+            if length trueStack == length falseStack && soundnessOK
+              then do
+                -- Merge stacks using ITE expressions, simplifying to prevent growth
+                let condSimp = Expr.simplify cond
+                    mergeExpr t f
+                      | t == f    = t
+                      | otherwise = Expr.simplify (ITE condSimp t f)
```
blishko: This is part of the `Expr.simplify` for `ITE`, maybe we can just do 
```
mergeExpr t f = Expr.simplify (ITE condSimp t f)
```

src/EVM/Merge.hs
```diff
@@ -0,0 +1,144 @@
...
+                    mergedStack = zipWith mergeExpr trueStack falseStack
+                -- Use vm0 as base and update only PC and stack
+                when conf.debug $ traceM $ "Merged forward jump"
+                  ++ runSrcLookup vm0.config.srcLookup vm0.env.contracts vm0.state.contract jumpTarget
```
blishko:
```suggestion
                  <> runSrcLookup vm0.config.srcLookup vm0.env.contracts vm0.state.contract jumpTarget
```

cli/cli.hs
```diff
@@ -131,6 +132,7 @@ commonOptions = CommonOptions
   <*> (switch $ long "only-deployed" <> help "When trying to resolve unknown addresses, only use addresses of deployed contracts")
   <*> (optional $ strOption $ long "cache-dir" <> help "Directory to save and load RPC cache")
   <*> (switch $  long "early-abort" <> help "Stop exploration immediately upon finding the first counterexample")
+  <*> (option auto $ long "merge-max-budget" <> showDefault <> value 10000 <> help "Max instructions for speculative merge exploration during path merging")
```
blishko: This default is different than the value in defaultConfig.
We should have one source of truth for the default value.
I would actually try with much smaller default, maybe `100`?

src/EVM/Types.hs
```diff
@@ -717,10 +730,25 @@ data BaseState
...
+
+instance Show SrcLookup where
+  show _ = "<SrcLookup>"
+
+-- | Run a SrcLookup to get source location info, with a fallback for when
+-- no SrcLookup is available.
+runSrcLookup :: Maybe SrcLookup -> Map (Expr EAddr) Contract -> Expr EAddr -> Int -> String
+runSrcLookup Nothing _ addr pc = " at addr: " <> show addr <> " at pc: " <> show pc
+runSrcLookup (Just (SrcLookup f)) contracts addr pc = f contracts addr pc
+
```
blishko: Could we separate this change regarding source code lookup into a different PR?
Seems like it is an independent change, no?
It would shrink the changes in this PR significantly.

cli/cli.hs
msooseth: Ah, that's my fault, sorry, not the original PR. I'd say 100 is good, Gustavo I think may want 1000? The abdk tests and the tests in tests.hs both pass with 9, but with 8 we fail the abkd tests. So IMO 100 is fair.

src/EVM/Merge.hs
msooseth: Good point, fixed.

cli/cli.hs
msooseth: @gustavo-grieco what do you think? I made the default and this default both 100 with a commit, but I'm happy to be swayed either way.

src/EVM/Types.hs
msooseth: Yes, good point. I really wanted to the source code lookup, but you are right, it's not necessary. However, I really want it in another PR :)

msooseth: OK, removed. New pr at #1010 as a follow-up once this PR ha been merged.

src/EVM.hs
msooseth: I think it'll be fixed now -- I think this PR is getting through in the coming days/hours :)

src/EVM/Merge.hs
msooseth: Sorry. The corresponding line, with the updated PR without the source line lookup is using `<>` now. Thanks!

cli/cli.hs
gustavo-grieco: Let's go with 100

src/EVM.hs
blishko: I am not sure the comment is right.
In the sense that I don't think we were doing anything wrong before, no?

test/test.hs
```diff
@@ -3963,6 +3963,200 @@ tests = testGroup "hevm"
...
+              uint256 result = 1;
+              if (x & 0x1 != 0) result = result * 2;
+              if (x & 0x2 != 0) result = result * 3;
+              assert(result > 0);
+            }
+          }
+          |]
+        let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
+        (_, []) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        putStrLnM "Simple branch merging works"
```
blishko: How do we know if branch merging occurred or not?

test/test.hs
```diff
@@ -3963,6 +3963,200 @@ tests = testGroup "hevm"
...
+              if (x & 0x1 != 0) result = result * 2;
+              if (x & 0x2 != 0) result = result * 3;
+              // Bug: result == 6 when both bits are set
+              assert(result != 6);
+            }
+          }
+          |]
+        let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
+        (_, [Cex _]) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        putStrLnM "Counterexample found with merged paths"
```
blishko: Can we avoid unconditional print statements in tests?

test/test.hs
```diff
@@ -3963,6 +3963,200 @@ tests = testGroup "hevm"
...
+        let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
+        (paths, []) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        let numPaths = length paths
+        -- Without merging: 16 paths. With unchecked + merging: should be <= 4
+        liftIO $ assertBool ("Expected at most 4 paths with unchecked merging, got " ++ show numPaths) (numPaths <= 4)
+        putStrLnM $ "Many branches merged: " ++ show numPaths ++ " paths (vs 16 without merging)"
+    -- TODO: Checked arithmetic merging doesn't work well with solcRuntime's standard-json output
+    -- The bytecode structure differs from direct solc --bin-runtime compilation
+    -- Direct compile: 4 merges, 4 paths. Standard-json: 0 merges, 18 paths
+    -- This needs investigation into why standard-json produces different bytecode structure
```
blishko: It seems to me we should investigate this first.

gustavo-grieco: I suspect that the solc optimizer is not running, and therefore, the code produced different paths that cannot be merged, but I agree we should investigate it. We could even produce some recommendations to users on how to compile in order to make state merging more likely to happen.

msooseth:
Checked arithmetic probably fails because it reverts in one of the branches and that is a Nothing and we abort on that (see comment in code — it explicitly lists this as something it cannot handle). We’d need to handle Nothing for that to work. Changing solc options won’t help AFAIK, other than turning off checked arithmetic in solc.

Mate

test/test.hs
msooseth: This was actually a wrong test. Due to revert issues in checked arithmetic, this didn't even merge. This has now been rectified. The number of End Expr-s are checked as being reduced.

src/EVM.hs
msooseth: Sorry, I was wrong, removing this comment.

test/test.hs
msooseth: I have now added a failing tests specifically to test this. I'll adjust the comment etc. to reflect the knowledge. It's because of revert which leads to Nothing, which I have already mentioned in a separate comment. I'll update this and then resolve :)

msooseth: OK, can  you please check this now?

test/test.hs
```diff
@@ -3963,6 +3967,219 @@ tests = testGroup "hevm"
...
+          }
+          |]
+        let sig = Just (Sig "f(uint256)" [AbiUIntType 256])
+        noMerege <- runEnv testEnv { config = testEnv.config {mergeMaxBudget = 0}} $ do
+          e <- withDefaultSolver $ \s -> getExpr s c sig [] defaultVeriOpts
+          pure $ length e
+        merge <- runEnv testEnv { config = testEnv.config {mergeMaxBudget = 1000}} $ do
+          e <- withDefaultSolver $ \s -> getExpr s c sig [] defaultVeriOpts
+          pure $ length e
+        assertBoolM "Merging should reduce number of paths" (merge < noMerege)
```
blishko, line 3995: Nice!
This is a good way to test the behaviour!

PR #975, opened 2026-01-21
Test: Collect more Expr/Prop related tests to a dedicated test module

blishko:
## Description

This is a continuation of #972. We are still extracting and collecting in one module unit tests related to Expr/Prop construction and simplification.

Next step will be to move property tests. Actually, some of the unit tests would be better expressed as property tests.


msooseth:
wow, nice. Yeah, this is a ton of pretty fragile work. Thanks for doing it! The test module has been pretty unregulated for a while.

PR #984, opened 2026-01-27
Test: Add framework for concrete execution test

blishko:
We aim to separate concrete execution tests to a separate module.
Here we prepare the ground for testing the expected result of executing a single call to a specific function of a solidity contract.
We define several helpers and demonstrate with three simple examples how to create new test cases.

### DIFF
--- a/test/EVM/ConcreteExecution/ConcreteExecutionTests.hs
+++ b/test/EVM/ConcreteExecution/ConcreteExecutionTests.hs
@@ -1,0 +1,131 @@
+{-# LANGUAGE QuasiQuotes #-}
+
+module EVM.ConcreteExecution.ConcreteExecutionTests (tests) where
+
+import Control.Monad.ST (stToIO)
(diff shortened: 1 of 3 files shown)


blishko:
I plan to use this to define new tests for `assume` cheatcode in the concrete execution.

test/EVM/ConcreteExecution/ConcreteExecutionTests.hs
```diff
@@ -0,0 +1,131 @@
...
+    , prevRandao     = 0
+    , schedule       = feeSchedule
+    , chainId        = 1
+    , create         = False
+    , baseState      = EmptyBase
+    , txAccessList   = mempty
+    , allowFFI       = False
+    , freshAddresses = 0
+    , beaconRoot     = 0
+    }
```
blishko, line 72: I think we need to have some default "empty" `VMOpts`. I'll try to do that separately.

PR #987, opened 2026-01-28
Solvers: Do not specify bv-solver option for bitwuzla

blishko:
Seems like the default choice for bv-solver (bit-blast) performs
better, specially on benchmarks that can be solved quickly, like in our
expression equivalence property tests.

With this change, our tests run faster with Bitwuzla than with cvc5.


msooseth:
Nice!

PR #988, opened 2026-01-28
Move from `ethereum/tests` to `ethereum/execution-spec-tests`

elopez:
## Description
This moves hevm from `ethereum/tests` to `ethereum/execution-spec-tests`

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

PR #992, opened 2026-01-29
EVM: fix MCOPY stack corruption on out-of-gas in nested calls

elopez:
## Description

The MCOPY opcode handler had `assign' (#state % #stack) xs` placed outside the `burn` continuation block. When a nested call ran out of gas during MCOPY execution:

1. burn detected insufficient gas and called vmError
2. vmError called finishFrame which correctly restored the parent's stack
3. Control returned to MCOPY and the code continued past the burn block
4. `assign' (#state % #stack) xs` executed, overwriting the parent's restored stack with the callee's remaining stack

This caused the parent contract to lose its stack state, leading to an incorrect execution. This issue was detected with the newer Ethereum tests in #988 

The fix moves `assign' (#state % #stack) xs` inside the burn continuation so it only executes when burn succeeds.

This also adds an early return for zero-size MCOPY operations.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -658,7 +658,7 @@ exec1 conf = do
 
         OpMcopy -> {-# SCC "OpMcopy" #-}
           case stk of
-            dstOff:srcOff:sz:xs ->  do
+            dstOff:srcOff:sz:xs ->
               case sz of
                 Lit sz' -> do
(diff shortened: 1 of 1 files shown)


blishko (approved):
Great find!

Issue #995, opened 2026-01-30
Use uninterpreted functions for division


gustavo-grieco:
SMT bitvector division (`bvudiv`, `bvsdiv`, `bvurem`, `bvsrem`) is expensive for solvers. Halmos uses **uninterpreted functions** during symbolic execution, deferring exact division semantics to a refinement phase. This dramatically speeds up path exploration while remaining sound.

msooseth:
We had a version of this but we removed it. It may be useful for exploration though! Didn’t think of that. We could also just limit the timeout for the solver when it’s running an exploration query? Maybe that’s an easier solution? Modern solvers do this kind of transformation already usually. Now that we can have individual timeouts for queries, this would be a VERY easy thing to do… and may be simpler? Also can be done ALONGSIDE this solution!

gustavo-grieco:
It seems that we need at least some of this for `prove_div_negative_divisor` and  `prove_div_values` from [abdk-math-64.64-verification](https://github.com/gustavo-grieco/abdk-math-64.64-verification). Feel free to experiment with these, these takes seconds to get the problematic smt queries and they are not very big.

blishko:
As Mate said, this would ideally be implemented at the level of the solvers. Bitwuzla should be doing this kind of abstraction internally.
Let's see if we can figure out why they would be struggling on these examples.

gustavo-grieco:
I'm totally in favor of using bitwuzla infrastructure, if it works as good as our own, and we don't have to maintain additional code. If this is exclusive from bitwuzla, however, we must consider strongly recommend user to use that solver over others. In any case, here are the zip unsolved queries for each abdk invariant related with division:

[prove_div_negative_divisor.zip](https://github.com/user-attachments/files/25003989/prove_div_negative_divisor.zip)
[prove_div_values.zip](https://github.com/user-attachments/files/25003990/prove_div_values.zip)

I don't know if any configuration of bitwuzla can solve this in a reasonable time. If you want, I can also provide the augmented smts that can be easily solved in seconds using #996.

gustavo-grieco:
@blishko please check if this zip file contains the abstract arithmetic queries that you need. 

[prove_div_values_abs.zip](https://github.com/user-attachments/files/25013213/prove_div_values_abs.zip)

msooseth:
## Note to self
- It takes 5 mins to `prove_div_values` with the PR, but we get`8x -> SMT solver says: Result unknown by SMT solver` without it
[...]

PR #996, opened 2026-01-30
[PARKED] Use uninterpreted functions for 2-phase solving approach

gustavo-grieco:
## Description

A crude prototype for #995 The system now does 2-phase solving. It can now deal with these two ABDK verification problems:
```sh
make verify-hevm T=prove_div_negative_divisor
make verify-hevm T=prove_div_values
```
Once abstraction-refinement is turned on. See [this repo ](https://github.com/gustavo-grieco/abdk-math-64.64-verification )for details. 

This change needs a flag to be set in the CLI for it to be enabled.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,377 @@
...
+    mkAssertion (fname, a, b) = do
+      aenc <- exprToSMTWith AbstractDivision a
+      benc <- exprToSMTWith AbstractDivision b
+      let result = "(" <> fname `sp` aenc `sp` benc <> ")"
+      if fname == "evm_bvudiv"
+        -- (x / y) <= x
+        then pure $ SMTCommand $ "(assert (bvule " <> result `sp` aenc <> "))"
+        -- (x % y) <= y (ULE not ULT because y could be 0 and 0 % 0 = 0)
+        else pure $ SMTCommand $ "(assert (bvule " <> result `sp` benc <> "))"
+
```
msooseth: This is nice

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,377 @@
...
+-- | Encode props using uninterpreted functions for div/mod (Phase 1 of two-phase solving)
+assertPropsAbstract :: Config -> [Prop] -> Err SMT2
+assertPropsAbstract conf ps = do
+  let mkBase s = assertPropsHelperWith AbstractDivision s divModAbstractDecls
+  base <- if not conf.simp then mkBase False ps
+          else mkBase True (decompose conf ps)
+  bounds <- divModBounds ps
+  pure $ base
+      <> SMT2 (SMTScript bounds) mempty mempty
```
msooseth:
```suggestion
  pure $ base <> SMT2 (SMTScript bounds) mempty mempty
```

src/EVM/Solvers.hs
```diff
@@ -128,9 +129,46 @@ checkSatWithProps sg props = do
   if psSimp == [PBool False] then pure Qed
   else do
     let concreteKeccaks = fmap (\(buf,val) -> PEq (Lit val) (Keccak buf)) (toList $ Keccak.concreteKeccaks props)
-    let smt2 = assertProps conf (if conf.simp then psSimp <> concreteKeccaks else psSimp)
```
msooseth, line 131: Wow this was a bug in the original code! I'll push a PR to fix. Should state:
```
 (if conf.simp then psSimp <> concreteKeccaks else props)
```

msooseth: Ahh, actually, not. This was fine!

src/EVM/Solvers.hs
```diff
@@ -128,9 +129,46 @@ checkSatWithProps sg props = do
...
+      -- Two-phase solving with abstract division
+      -- Phase 1: Use uninterpreted functions (overapproximation)
+      let smt2Abstract = assertPropsAbstract conf allProps
+      if isLeft smt2Abstract then pure $ Error $ getError smt2Abstract
+      else do
+        phase1 <- liftIO $ checkSat sg (Just props) smt2Abstract
+        case phase1 of
+          Qed -> pure Qed            -- UNSAT with abstractions => truly UNSAT (sound)
+          Error e -> pure (Error e)
+          Unknown u -> pure (Unknown u)
```
msooseth:
```suggestion
          e@(Error _) -> pure e
          u@(Unknown _) -> pure u
```

src/EVM/Solvers.hs
```diff
@@ -128,9 +129,46 @@ checkSatWithProps sg props = do
...
+                  -- SAT/Unknown results are discarded (fall back to phase2 Unknown).
+                  when conf.debug $ liftIO $ putStrLn "Phase 2 unknown, trying shift-based bounds..."
+                  let smt2Shift = assertPropsShiftBounds conf allProps
+                  if isLeft smt2Shift then pure phase2
+                  else do
+                    when conf.dumpQueries $ liftIO $ writeSMT2File (getNonError smt2Shift) "." "shift-bounds"
+                    phase3 <- liftIO $ checkSat sg (Just props) smt2Shift
+                    case phase3 of
+                      Qed -> pure Qed  -- UNSAT with shift bounds => truly UNSAT
+                      _   -> pure phase2  -- SAT/Unknown from shift bounds is not reliable
```
msooseth: Do we _really_ need this 3-phase refinement? Can we just do 2? Is this important?

src/EVM/SMT.hs
```diff
@@ -434,23 +451,23 @@ exprToSMT = \case
   Mul a b -> op2 "bvmul" a b
   Exp a b -> case a of
     Lit 0 -> do
-      benc <- exprToSMT b
+      benc <- exprToSMTWith enc b
```
msooseth: Can you please refactor by declaring a lambda function
`exprToSMT` that is equal to `exprToSMTWith enc` -- this would make the diff a lot smaller and more readable I think!

src/EVM/SMT.hs
```diff
@@ -587,17 +604,52 @@ exprToSMT = \case
...
+    smodOp abstractOp a b = case enc of
+      AbstractDivision -> op2 abstractOp a b
+      ConcreteDivision -> do
+        aenc <- exprToSMTWith enc a
+        benc <- exprToSMTWith enc b
+        let absa = "(ite (bvsge " <> aenc `sp` zero <> ")" `sp` aenc `sp` "(bvsub" `sp` zero `sp` aenc <> "))"
+            absb = "(ite (bvsge " <> benc `sp` zero <> ")" `sp` benc `sp` "(bvsub" `sp` zero `sp` benc <> "))"
+            urem = "(bvurem" `sp` absa `sp` absb <> ")"
+        pure $ "(ite (=" `sp` benc `sp` zero <> ")" `sp` zero `sp`
+                "(ite (bvsge" `sp` aenc `sp` zero <> ")" `sp` urem `sp` "(bvsub" `sp` zero `sp` urem <> ")))"
```
msooseth: Wow this actually modifies the CONCRETE (i.e. phase2) system, too. Interesting! Maybe we could even have this as a separate PR, although I understand that's a bit annoying. But yeah, this is very interesting.

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,376 @@
...
+    collectDivOps = \case
+      Div a b  -> [(0, a, b)]
+      SDiv a b -> [(1, a, b)]
+      Mod a b  -> [(2, a, b)]
+      SMod a b -> [(3, a, b)]
+      _        -> []
+
+    eqDivOp :: DivOp -> DivOp -> Bool
+    eqDivOp (k1, a1, b1) (k2, a2, b2) =
+      k1 == k2 && show a1 == show a2 && show b1 == show b2
```
msooseth: Is this `show x == show y` efficient? Do we need the `show`?

msooseth: Maybe we can make `DivOp` deriving `Eq` instead?

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,376 @@
...
+    entries <- concat <$> mapM (uncurry mkGroupAxioms) indexedGroups
+    let links = mkCongruenceLinks indexedGroups
+    pure $ (SMTComment "division/modulo ground-instance axioms (CSE'd)") : entries <> links
+  where
+    collectDivOps :: forall a . Expr a -> [DivOp]
+    collectDivOps = \case
+      Div a b  -> [(0, a, b)]
+      SDiv a b -> [(1, a, b)]
+      Mod a b  -> [(2, a, b)]
+      SMod a b -> [(3, a, b)]
```
msooseth: Having them as numbers is pretty confusing. Can we have an enum instead?

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,376 @@
...
+  bounds <- divModBounds ps
+  axioms <- divModShiftBoundAxioms ps
+  pure $ base
+      <> SMT2 (SMTScript bounds) mempty mempty
+      <> SMT2 (SMTScript axioms) mempty mempty
+
+-- | Generate shift-based bound axioms (no bvudiv/bvurem).
+-- For each group of signed div/mod ops, if the dividend has a SHL(k, _) structure,
+-- generates bounds using bvlshr instead of bvudiv.
+divModShiftBoundAxioms :: [Prop] -> Err [SMTEntry]
```
msooseth: Do we really need all these complex axioms? Wouldn't be best to have the SMT solver generate these? I feel like the solver _should_ be able to generate these on its own? Do we have substantially better/more information than the solver because of our AST knowledge?

msooseth (commented):
Some thoughts and comments. I'm reluctant because:
- There are all these axioms and they'll take time to review
- I am not sure we need 3 phases

There are some thought-provoking parts here, though:
- The improvement to the concrete SMT translation (i.e. even forgetting about all this abstraction)
- The axioms may _truly_ reflect that we have more information at hand than the SMT solver, and adding them may actually be sufficient, without the abstraction-refinement?

These 2 could actually be added as-is to the system, I think!

msooseth:
Can you also maybe fix the failures of the checks? I think it's failing only on warnings, so they should be easy to fix, and then we can see if the fuzz tests go through!

src/EVM/SMT.hs
gustavo-grieco: Done

src/EVM/SMT/DivEncoding.hs
gustavo-grieco: Done

src/EVM/SMT/DivEncoding.hs
gustavo-grieco: Done

src/EVM/SMT/DivEncoding.hs
gustavo-grieco: Claude said:

The axioms encode knowledge that the solver cannot derive on its own because we replace bvsdiv/bvsrem with uninterpreted functions. Without axioms, the uninterpreted functions are completely unconstrained — the solver doesn't know that evm_bvudiv(a, b) <= a or that signed division decomposes into unsigned division on absolute values. These aren't things the solver can "generate" since it has no theory for uninterpreted functions beyond congruence. The key axioms are:

1. Bounds (divModBounds): q <= a for div, r <= b for mod — prunes impossible models
2. Ground-instance axioms (divModGroundAxioms): pin each uninterpreted function call to its exact bvudiv/bvurem semantics with CSE'd intermediates
3. Signed decomposition: express sdiv/smod in terms of bvudiv/bvurem on absolute values

The bounds alone (Phase 1) give us a cheap UNSAT check. The ground axioms (Phase 2) give exact semantics. Without the axioms, the abstract encoding would be unsound for SAT results and imprecise for UNSAT.

src/EVM/Solvers.hs
gustavo-grieco: Claude said:
Phase 3 (shift-based bounds) is an optimization for a specific pattern: when the dividend has the form SHL(k, x), we can derive tight quotient bounds using bvlshr instead of bvudiv. This avoids bvudiv entirely, which bitwuzla struggles with at 256 bits. It only fires when Phase 2 returns Unknown (solver timeout on exact bvudiv), and only UNSAT results are trusted (overapproximation). That said, if this pattern doesn't arise often enough to justify the complexity, we can drop Phase 3 and keep just 2 phases. Happy to remove it if you'd prefer simplicity.

src/EVM/Solvers.hs
gustavo-grieco: Done

gustavo-grieco:
Keep in mind that abstract arithmetics are not enabled by default, I don't know if we should enable it for all the test 🤔

src/EVM/Solvers.hs
msooseth: Let's go for simplicity for now, and keep this in mind, perhaps as a follow-up PR? If you could ask Claude to cut it out of this PR, we could review this PR and the follow-up can be reviewed later? I'd rather do this piecemeal than in one big blob

src/EVM/SMT/DivEncoding.hs
msooseth: OK, fair, let me close. Then we need to review these axioms.

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,385 @@
...
+    mkAssertion :: (Builder, Expr EWord, Expr EWord) -> Err SMTEntry
+    mkAssertion (fname, a, b) = do
+      aenc <- exprToSMTWith AbstractDivision a
+      benc <- exprToSMTWith AbstractDivision b
+      let result = "(" <> fname `sp` aenc `sp` benc <> ")"
+      if fname == "evm_bvudiv"
+        -- (x / y) <= x
+        then pure $ SMTCommand $ "(assert (bvule " <> result `sp` aenc <> "))"
+        -- (x % y) <= y (ULE not ULT because y could be 0 and 0 % 0 = 0)
+        else pure $ SMTCommand $ "(assert (bvule " <> result `sp` benc <> "))"
```
msooseth: Wow this `if` does nothing :D I'll have to investigate @gustavo-grieco

src/EVM/SMT.hs
msooseth: I hope this PR can go in as-is and then we don't have to do 2 PRs :)

src/EVM/SMT/DivEncoding.hs
```diff
@@ -0,0 +1,302 @@
...
+    mkAssertion (fname, a, b) = do
+      aenc <- exprToSMTAbst a
+      benc <- exprToSMTAbst b
+      let result = "(" <> fname `sp` aenc `sp` benc <> ")"
+      pure $ SMTCommand $ "(assert (bvule " <> result `sp` aenc <> "))"
+
+data DivModOp = IsDiv | IsMod
+  deriving (Eq, Ord)
+
+data DivOpKind = UDiv | USDiv | UMod | USMod
```
blishko: What is `USDiv`? Shouldn't it be `SDiv`?

test/test.hs
```diff
@@ -848,6 +1279,437 @@ tests = testGroup "hevm"
...
+              assembly { q := mod(x, y) }
+              assert(q != 0);
+            }
+          } |]
+        (_, res) <- withBitwuzlaSolver $ \s -> checkAssert s defaultPanicCodes c Nothing [] defaultVeriOpts
+        assertBoolM "Expected counterexample" (any isCex res)
+    ]
+  , testGroup "Abstract-Arith"
+    -- "make verify-hevm T=prove_div_negative_divisor" in https://github.com/gustavo-grieco/abdk-math-64.64-verification
+    [ testCase "prove_div_values-abdk" $ do
```
msooseth, line 868: @blishko what timed out before.

test/test.hs
```diff
@@ -848,6 +1279,437 @@ tests = testGroup "hevm"
...
+        runEnv testEnvAbstract $ do
+          (_, res) <- withShortBitwuzlaSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        -- with abstract arith, we prove it
+          assertEqualM "Must be QED" res []
+        runEnv testEnv $ do
+          (_, res) <- withShortBitwuzlaSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+          -- without abstract arith, we time out
+          liftIO $ assertBool "Must be unknown" (all isUnknown res)
+    -- "make verify-hevm T=prove_div_negative_divisor" in https://github.com/gustavo-grieco/abdk-math-64.64-verification
+    , testCase "prove_div_negative_divisor" $ do
```
msooseth, line 923: @blishko what timed out before.

msooseth (commented):
Just two notes about what times out without this abstraction-refinement. The tests actually prove they time out

msooseth:
Let's park this for now. I wanna hear back from the Bitwuzla developers and I'm also a bit tired of trying to make this work :) We can come back to this another time.

gustavo-grieco:
Replaced by https://github.com/argotorg/hevm/pull/1075

PR #997, opened 2026-02-01
Test: Refactor and adjust property tests for Expr/Prop

blishko:
We move property tests for Expr/Prop to a dedicated ExprTests module. We also move generators for these types to a separate module.


msooseth (approved):
If this is just rearranging things, then it's good!

PR #998, opened 2026-02-02
Switch tests to Osaka

elopez:
## Description

This switches hevm to the test suite for Osaka. Some basic Osaka support (EIP-2935, EIP-7883) is implemented to keep the ignored test list "small".

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -226,7 +226,8 @@ jobs:
 
       - name: Download execution-spec-tests fixtures
         run: |
-          curl --retry 5 -fsSL https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_stable.tar.gz | tar xz
+          ARTIFACT_URL="$(grep ethereum/execution-spec-tests flake.nix | cut -f2 -d'"')"
+          curl --retry 5 -fsSL "$ARTIFACT_URL" | tar xz
 
       - uses: actions/checkout@v4
         with:


elopez:
The "stable" fixture does not have Osaka tests, as it goes only up to Prague, so we'd need "develop" for now: https://github.com/ethereum/execution-specs/blob/tests-v5.4.1/.github/configs/feature.yaml#L2-L8

flake.nix
```diff
@@ -27,8 +27,8 @@
     flake-utils.lib.eachDefaultSystem (system:
       let
         execution-spec-tests-fixtures = builtins.fetchTarball {
-          url = "https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_stable.tar.gz";
-          sha256 = "0a0i20f611aghvh1vghvzcxilbh80bd4174n7ih7iarn5x71aq1p";
+          url = "https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_develop.tar.gz";
+          sha256 = "0dsg7lcj591i0p7s93xv0vin2dwricgq29v8xakqnx1kfj3fcyvq";
```
blishko, line 31: This needs to be updated also for the windows build, no? I think there it is still "stable"?

blishko: Yes, we need to do that.
We should also add some assertion that the number of tests we are going to run is more than 0 :)

Right now, if we mess something up and there are no test cases to run, the job will happily succeed with 0 passing and 0 failing tests.

elopez: Good catch! I replaced it with some bash to extract the URL from the flake, hopefully that prevents mismatches going forward.

blishko (approved):
LGTM!

msooseth:
:partying_face:

PR #999, opened 2026-02-02
Tests: Signal failure when we inadvertently lose some ethereum tests

blishko:
## Description
It has happened on two occasions before that there was an error and suddenly we were not getting any tests cases from the ethereum test repository.
(One was erroneus change in the parsing, second was an error in the url.)
Previously, such a bug could (and did) go unnoticed, because the CI job would run successfully with 0 tests passing and 0 tests failing. Therefore, we specify the expected number of tests in the code and error out if the actual number is lower than the expected number.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Wow yes, thanks, this was 100% needed

PR #1003, opened 2026-02-03
EVM: Introduce deafult VMOpts

blishko:
This value represents decent choice for the VMOpts fields. It can be easily customized using record update syntax. This should make it easy if we need to introduce a new field there.

### DIFF
--- a/src/EVM/SymExec.hs
+++ b/src/EVM/SymExec.hs
@@ -48,13 +48,12 @@ import Options.Generic (ParseField, ParseFields, ParseRecord)
 import Text.Printf (printf)
 import Witch (into, unsafeInto)
 
-import EVM (makeVm, abstractContract, initialContract, getCodeLocation, isValidJumpDest)
+import EVM (makeVm, abstractContract, initialContract, getCodeLocation, isValidJumpDest, defaultVMOpts)
 import EVM.Exec (exec)
 import EVM.Fetch qualified as Fetch
 import EVM.ABI
 import EVM.Effects
 import EVM.Expr qualified as Expr
-import EVM.FeeSchedule (feeSchedule)
 import EVM.Format (formatExpr, formatPartial, formatPartialDetailed, showVal, indent, formatBinary, formatProp, formatState, formatError)
 import EVM.SMT qualified as SMT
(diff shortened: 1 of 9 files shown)


src/EVM/UnitTest.hs
```diff
@@ -545,15 +541,8 @@ initialUnitTestVm (UnitTestOptions {..}) theContract = do
            , priorityFee = testParams.priorityFee
            , maxCodeSize = testParams.maxCodeSize
            , prevRandao = testParams.prevrandao
-           , schedule = feeSchedule
            , chainId = testParams.chainId
            , create = True
-           , baseState = EmptyBase
-           , txAccessList = mempty -- TODO: support unit test access lists???
-           , allowFFI = ffiAllowed
```
msooseth, line 553: I think here we had this set to something other than the default?

src/EVM/SymExec.hs
```diff
@@ -272,34 +253,18 @@ loadSymVM
...
-    , number = Lit 0
-    , timestamp = Lit 0
     , blockGaslimit = 0
-    , gasprice = 0
     , prevRandao = 42069
-    , gas = ()
-    , gaslimit = 0xffffffffffffffff
-    , baseFee = 0
-    , priorityFee = 0
-    , maxCodeSize = 0xffffffff
```
msooseth, line 294: This is quite a bit smaller than maxBound I think? However, it probably doesn't matter given that this is a symbolic VM, so probably it's fine, just wanted to be a bit precise.

src/EVM/UnitTest.hs
blishko: Oops! Good catch, thanks!

src/EVM/SymExec.hs
blishko: Right, I am making `0xffffffff` the default value. That should match precisely what we had before.
Though I am not really sure why we are using this value.

msooseth: Yeah, it's probably huge anyway. Thanks :)

src/EVM/UnitTest.hs
msooseth: Nice, thanks!

PR #1008, opened 2026-02-05
Tests: Move equivalence tests to dedicated module (and simplify)

blishko:
In addition, we no longer print debug information by default.
Some output is now hidden behind the `debug` flag.
Equivalence for solc's yul optimization tests now has its own verbosity flag.
By default we no longer print anything. This should reduce visual clutter when running tests.


msooseth (approved):
LGTM

PR #1009, opened 2026-02-05
Full Osaka support

elopez:
## Description

This fixes all the things for Osaka support. All tests should be passing here.

It is also all AI-generated in an unsupervised manner, so please consider changes with care 😄 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


elopez:
Branch ai-fixes-split here should have the same changes but split per-EIP, in case it is easier to separate: https://github.com/argotorg/hevm/compare/ai-fixes-split?expand=1

PR #1010, opened 2026-02-05
Source code line display for state merging

msooseth:
## Description
This brings a more generic system that allows us to print source code lines at any point during EVM execution. I think this can be _very_ helpful. It also happens to do it for state merging.

To be applied after #973 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth:
Closing so I can open a much cleaner PR

PR #1017, opened 2026-02-10
Add 2 more Expr simplifications related to negative numbers, enable NOT concretization

msooseth:
## Description

Adds two new rewrites:
- [x] Mul (Lit 0xFF..FF) x = -x
- [x] Add (Lit 1) (Xor (Lit 0xFF..FF) x) = -x
- [x] Enable NOT concretization, which was accidentally not enabled

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -1226,7 +1226,7 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | otherwise = sub a b
 
     -- XOR normalization
-    go (Xor (Lit a) b) | a == maxLit = Not b
+    go (Xor (Lit a) b) | a == maxLit = EVM.Expr.not b
     go (Xor a  b) = EVM.Expr.xor a b
 
     -- Not simplification


src/EVM/Expr.hs
```diff
@@ -1212,6 +1212,10 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | b == c = a
       | otherwise = sub (add a b) c
 
+    -- Negation: ~x + 1 = 0 - x (two's complement)
+    go (Add (Lit 1) (Not b)) = Sub (Lit 0) b
+    go (Add (Lit 1) (Xor (Lit a) b)) | a == maxLit = Sub (Lit 0) b
```
blishko:
```suggestion
    go (Xor (Lit a) b) | a == maxLit = Not b
```

I think we should add this Xor rule (if we don't have it).
Then this xor rule, together with your first rule subsumes your second rule.

test/EVM/Expr/ExprTests.hs
```diff
@@ -279,6 +279,15 @@ basicSimplificationTests = testGroup "Basic simplification tests"
         a = successPath [PEq (Add (Lit 1) (Lit 2)) (Sub (Lit 4) (Lit 1))]
         b = Expr.simplify a
       assertEqual "Must simplify down" (successPath []) b
+  , testCase "simp-mul-neg1" $ do
+      let simp = Expr.simplify $ Mul (Lit Expr.maxLit) (Var "x")
+      assertEqual "mul by -1 is negation" simp (Sub (Lit 0) (Var "x"))
```
blishko:
```suggestion
      assertEqual "mul by -1 is negation" (Sub (Lit 0) (Var "x")) simp
```

First the expected result, then the obtained result :)
https://hackage-content.haskell.org/package/tasty-hunit-0.10.2/docs/Test-Tasty-HUnit.html#v:assertEqual

blishko (commented):
Nice!

src/EVM/Expr.hs
msooseth: OOo good point. Committing.

test/EVM/Expr/ExprTests.hs
msooseth: Yess sorry!

src/EVM/Expr.hs
```diff
@@ -1212,6 +1212,10 @@ simplifyNoLitToKeccak e = untilFixpoint (mapExpr go) e
       | b == c = a
       | otherwise = sub (add a b) c
 
+    -- Negation: ~x + 1 = 0 - x (two's complement)
+    go (Add (Lit 1) (Not b)) = Sub (Lit 0) b
+    go (Xor (Lit a) b) | a == maxLit = Not b
```
blishko: Let's move the Xor rule to the Xor section below.
Also, not sure what will happen if `b` is a literal. I guess the result should actually be `Expr.not b`? WDYT?

test/EVM/Expr/ExprTests.hs
```diff
@@ -279,6 +279,15 @@ basicSimplificationTests = testGroup "Basic simplification tests"
         a = successPath [PEq (Add (Lit 1) (Lit 2)) (Sub (Lit 4) (Lit 1))]
         b = Expr.simplify a
       assertEqual "Must simplify down" (successPath []) b
+  , testCase "simp-mul-neg1" $ do
+      let simp = Expr.simplify $ Mul (Lit Expr.maxLit) (Var "x")
+      assertEqual "mul by -1 is negation" (Sub (Lit 0) (Var "x")) simp
+  , testCase "simp-not-plus1" $ do
+      let simp = Expr.simplify $ Add (Lit 1) (Not (Var "x"))
+      assertEqual "~x + 1 is negation" simp (Sub (Lit 0) (Var "x"))
```
blishko:
```suggestion
      assertEqual "~x + 1 is negation" (Sub (Lit 0) (Var "x")) simp
```

test/EVM/Expr/ExprTests.hs
```diff
@@ -279,6 +279,15 @@ basicSimplificationTests = testGroup "Basic simplification tests"
...
+  , testCase "simp-xor-maxlit-plus1" $ do
+      let simp = Expr.simplify $ Add (Lit 1) (Xor (Lit Expr.maxLit) (Var "x"))
+      assertEqual "xor(maxLit, x) + 1 is negation" simp (Sub (Lit 0) (Var "x"))
```
blishko:
```suggestion
      assertEqual "xor(maxLit, x) + 1 is negation" (Sub (Lit 0) (Var "x")) simp
```

src/EVM/Expr.hs
msooseth: I'm moving it lower down. It'll make it into `Not b` which SHOULD call `EVM.Expr.not` to simplify concrete values but that's missing! WOW. Oooooops. We literally never simplify negated concrete values :S Adding that into this PR.

PR #1019, opened 2026-02-10
Add missing simplifications, add tests for all concretizations, remove SHA256

msooseth:
## Description
This PR:
- [x] Adds concretization tests for all ops
- [x] Adds PEq simplification tests for all(?) ops. I tried all...
- [x] Add missing `PEq (ConcreteStore x) (ConcreteStore y)` simplification when they were not equal
- [x] Missing LT simplification for when  `maxLit < a` which is always false
- [x] Missing `Mod x (Lit 1)` simplification, which is always 0
- [x] Missing `SMod x (Lit 1)` simplification, which is always 0
- [x] Missing `Xor a a` simplification, which is always 0
- [x] Missing `SHL a (Lit >=256)` which is always 0
- [x] Missing `SHR a (Lit >=256)` which is always 0
- [x] Missing `Or _ maxLit` = maxLit
- [x] Missing `Or maxLit _` = maxLit

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - State merging via speculative execution of both branches of a JUMPI, joining the
   resulting states whenever possible. Amount of speculative execution is
   controlled via `merge-max-budget`
-- Missing simplifications for Eq, Mod, SMod, XOR, SHL, SHR, Or and SHA256
+- Missing simplifications for Eq, Mod, SMod, XOR, SHL, SHR, and Or
 
 ## Changed
 - Simplifier now rewrites `Mul(-1, x)` and `~x + 1` to `Sub(0, x)`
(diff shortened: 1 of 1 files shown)


test/EVM/Expr/ExprTests.hs
```diff
@@ -635,6 +634,567 @@ signExtendTests = testGroup "SEx tests"
...
+
+    -- WriteWord
+  , testCase "conc-writeword" $ do
+      let simp = Expr.simplify $ ReadWord (Lit 0) (WriteWord (Lit 0) (Lit 0xab) (AbstractBuf "abc"))
+      assertEqual "WriteWord then ReadWord" (Lit 0xab) simp
+
+    -- CopySlice of zero size should be a no-op
+  , testCase "conc-copyslice-zero-size" $ do
+      let simp = Expr.simplify $ CopySlice (Lit 0) (Lit 0) (Lit 0) (AbstractBuf "src") (AbstractBuf "dst")
+      assertEqual "CopySlice size 0 is noop" (ConcreteBuf "dst") simp
```
blishko:
```suggestion
      assertEqual "CopySlice size 0 is noop" (AbstractBuf "dst") simp
```

msooseth:
Ooops, thanks for spotting that bug. The property tests are not happy because we don't inject SHA256 constraints to the solver, so SHA256 conretization is throwing it off. I am removing SHA256 conretization for the moment. It's the easiest fix, and I don't think we need SHA256 support for the moment.

PR #1024, opened 2026-02-12
Test: Less verbose test runs

blishko:
When comparing hevm and evm runs on random benchmarks, only print out information if the behaviour does not match or if verbose output is explicitly requested.


blishko:
That's my plan :)
I have been working on it step by step.

PR #1025, opened 2026-02-12
SymExec: Fix dumping formatted expression to file

blishko:
In equivalence checking, when comparing two branches, we can dump the end states to a file.
The information about the end state is collected as (Set Prop, String), where the first entry represents the some conditions under which the end states are different and the second entry is the formatted Expr representing the end state.
Previously, when dumping this information to a file,  we would call `show` on this pair to convert it into a string, which is then written to a file.
However, the second entry (the formatted expression) already contains newlines (and other special characters), and `show` converts that to literal string, so we lose the nice formatting.
Compare the behaviour:
```
ghci> let s :: String = "hello\nworld"
ghci> putStrLn s
hello
world
ghci> putStrLn (show s)
"hello\nworld"
```

The fix is to call `show` only on the first entry and keep the second entry as is.


msooseth:
Waaait.. isn't there some nice formatting in our `Format.hs` for these? That would probably do it even better?

blishko:
`msg` is already the formatted `String` obtained using these formatting helpers.
The only problem is that `show` destroys this nice formatting.

PR #1026, opened 2026-02-16
Test: Move more tests to dedicated modules

blishko:
We start a new model for symbolic execution over solidity source code and move a first batch of tests related to storage there.

We also move two tests related to storage simplification to ExprTests and remove unnecessary auxiliary functions.


msooseth:
I'm fine with this refactor, though I am not a huge fan of too many subdirectories :)

PR #1029, opened 2026-02-16
flake: keep only part of the execution spec test fixture

elopez:
## Description
This reduces the on-disk size from about 8GB to around 3GB

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,7 +227,7 @@ jobs:
       - name: Download execution-spec-tests fixtures
         run: |
           ARTIFACT_URL="$(grep ethereum/execution-spec-tests flake.nix | cut -f2 -d'"')"
-          curl --retry 5 -fsSL "$ARTIFACT_URL" | tar xz
+          curl --retry 5 -fsSL "$ARTIFACT_URL" | tar xz "fixtures/blockchain_tests"
 
       - uses: actions/checkout@v4
         with:


flake.nix
```diff
@@ -31,9 +31,19 @@
...
           config = { allowBroken = true; };
         });
-        execution-spec-tests-fixtures = pkgs.fetchzip {
-          url = "https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_develop.tar.gz";
-          hash = "sha256-eHvmhnQzdIun6mgngR+LmTdh4wa7j6TPBTGkIhk9Tzc=";
+        execution-spec-tests-fixtures = pkgs.stdenv.mkDerivation {
+          name = "execution-spec-tests-fixtures";
+          src = pkgs.fetchurl {
+            url = "https://github.com/ethereum/execution-spec-tests/releases/download/v5.4.0/fixtures_develop.tar.gz";
+            hash = "sha256-PisC1J/pA+2k/Yyspcvw0TnEcOl+HemoUpmxsDT5cJk=";
```
blishko, line 38: Why is the hash different?

elopez: I think the fetchurl hash is computed over the raw file while the fetchzip hash is calculated over the unpacked contents.

PR #1030, opened 2026-02-16
Compare solidity repo and solc versions between CI jobs

### LINKED ISSUE #1016: CI: Add a simple job to check that we are using the same version of solc in all places.
Currently we need to maintain a reference to solidity repository in both nix flake and widows setup. Also, we need to separately maintain a reference to binary version, also both in windows setup and nix flake.
We need to make sure that we use the same version everywhere.

elopez:
## Description

Closes #1016

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
Great, thanks!

PR #1033, opened 2026-02-19
Source code lines displayed during state merging

msooseth:
## Description
This prints source code lines during state merging debug. ALSO, and MORE importantly, this allows us to trivially display source code lines at ANY point during execution as long as we have access to `VM`, `addr`, and `PC`. So basically all the time.  Just write:

```
traceM $ "we are at:" <> runSrcLookup vm0.srcLookup vm0.env.contracts vm0.state.contract jumpTarget
```

Output looks like:
```
   Keccak preimages in state: 0
   Exploring call prefix 0xacce097a
Merged forward jump in file "src/loop-detect-fail-854.sol" on line 45 : "result >= MIN_64x64 && result <= MAX_64x64"
Merged forward jump in file "src/loop-detect-fail-854.sol" on line 63 : "result >= MIN_64x64 && result <= MAX_64x64"
Merged forward jump in file "src/loop-detect-fail-854.sol" on line 29 : "result >= MIN_64x64 && result <= MAX_64x64"
Merged forward jump in file "src/loop-detect-fail-854.sol" on line 63 : "result >= MIN_64x64 && result <= MAX_64x64"
```

I have added tests for both `symbolic`, i.e. when we don't have the source code, and when we have the source code via Forge.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog

### DIFF
--- a/src/EVM/Types.hs
+++ b/src/EVM/Types.hs
@@ -719,6 +719,9 @@ data ForkState = ForkState
   }
   deriving (Show, Generic)
 
+deriving instance Show (VM Symbolic)
+deriving instance Show (VM Concrete)
+
 -- | Alias for the type of e.g. @exec1@.
 type EVM (t :: VMType) a = StateT (VM t) (ST RealWorld) a
 
@@ -746,6 +749,7 @@ data RuntimeConfig = RuntimeConfig
   { allowFFI :: Bool
   , baseState :: BaseState
   }
+  deriving (Show)
 
 -- | An entry in the VM's "call/create stack"
 data Frame (t :: VMType) = Frame


src/EVM/Types.hs
```diff
@@ -718,9 +719,6 @@ data ForkState = ForkState
   }
   deriving (Show, Generic)
 
-deriving instance Show (VM Symbolic)
-deriving instance Show (VM Concrete)
```
blishko, line 722: Why do you need to remove the `Show` instances?

src/EVM/Types.hs
```diff
@@ -730,12 +728,24 @@ data BaseState
...
+runSrcLookup :: Maybe SrcLookup -> Map (Expr EAddr) Contract -> Expr EAddr -> Int -> String
+runSrcLookup Nothing _ addr pc = " at addr: " <> show addr <> " at pc: " <> show pc
+runSrcLookup (Just (SrcLookup f)) contracts addr pc = f contracts addr pc
+
 -- | Configuration options that need to be consulted at runtime
 data RuntimeConfig = RuntimeConfig
   { allowFFI :: Bool
   , baseState :: BaseState
   }
-  deriving (Show)
```
blishko, line 738: Why do you need to remove the `Show` instance?

blishko (commented):
Looks good!

I would just like to keep the `Show` instances.

src/EVM/Types.hs
msooseth: Because then I need a show instance for the `srcLookup`, but I'll make a simple one for it, I don't think we should have a full-fledged display for it.

msooseth: Note that I had one for it, but I thought you'd like that even less than removing `Show` :laughing:

msooseth:
OK, the `show`-s are there now, and the show for SrcLookup is

```
+instance Show SrcLookup where
+  show _ = "<SrcLookup Info>"
```

I hope that's OK. I don't think we need the whole thing displayed, it'd probably be extremely noisy for the viewer and likely basically useless.

PR #1036, opened 2026-02-23
Expr: Fix unsigned wrapping bugs in expression simplifier

elopez:
## Description

The expression simplifier had several locations where 256-bit unsigned integer wrapping was not handled, causing incorrect simplifications that the SMT solver could distinguish from the original expressions. This manifested as intermittent QuickCheck failures in simplify-joinbytes and simplifyProp-equivalence-sym.

These fixes and tests have been developed with Claude Code and I'm not familiar with this part of the codebase, so please review with care.

## Checklist

- [x] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -530,8 +530,10 @@ exec1 conf = do
             next >> pushSym vm.state.callvalue
 
         OpCalldataload -> {-# SCC "OpCalldataload" #-} stackOp1 g_verylow $
+        -- Reading past call data length should always return 0. We force this in the concrete case.
+        -- However, symbolic mode relies on `Expr.readWord` which has wrap-around semantics
           \ind -> case (ind, vm.state.calldata) of
-            (Lit i, ConcreteBuf bs) | i > (fromIntegral $ BS.length bs) -> Lit 0 -- Reading past call data length returns 0; `Expr.readWord` has wrap-around semantics
+            (Lit i, ConcreteBuf bs) | i > (fromIntegral $ BS.length bs) -> Lit 0
             _ -> Expr.readWord ind vm.state.calldata
 
         OpCalldatasize -> {-# SCC "OpCalldatasize" #-}


test/EVM/Expr/ExprTests.hs
```diff
@@ -39,6 +39,7 @@ unitTests = testGroup "Unit tests"
   , comparisonTests
   , signExtendTests
   , concretizationTests
+  , wrappingTests
```
blishko: Let's put these as part of `simplificationTests`.

elopez: done 👍

msooseth:
Wow, nice.

src/EVM/Expr.hs
```diff
@@ -299,9 +307,28 @@ readWord idx b@(WriteWord idx' val buf)
...
--- reading a Word that is lower than the dstOffset-32 of a CopySlice, so it's just reading from dst
-readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst) | dstOffset >= x+32 = readWord i dst
-readWord i@(Lit x) (CopySlice _ (Lit dstOffset) (Lit size) _ dst) | x >= dstOffset + size = readWord i dst
+-- reading a Word that is before the CopySlice destination region, so just read from dst
+-- We must ensure x+32 and dstOffset+size do not wrap past maxBound.
+-- When size is symbolic, we bound dstOffset to ensure no wrapping is possible.
+readWord i@(Lit x) (CopySlice _ (Lit dstOffset) _ _ dst)
+  | dstOffset >= x + 32
+  , x + 32 >= x
+  , dstOffset <= (maxBound :: W256) - maxBytes
```
msooseth, line 316: I honestly don't get this one.

```
maxBytes = into (maxBound :: Word32) `Prelude.div` 8
```

So this is going to be always 7/8*maxBound, right? Does that help, though? I'm confused.

src/EVM.hs
```diff
@@ -530,7 +530,9 @@ exec1 conf = do
             next >> pushSym vm.state.callvalue
 
         OpCalldataload -> {-# SCC "OpCalldataload" #-} stackOp1 g_verylow $
-          \ind -> Expr.readWord ind vm.state.calldata
+          \ind -> case (ind, vm.state.calldata) of
+            (Lit i, ConcreteBuf bs) | i > (fromIntegral $ BS.length bs) -> Lit 0 -- Reading past call data length returns 0; `Expr.readWord` has wrap-around semantics
+            _ -> Expr.readWord ind vm.state.calldata
```
msooseth, line 537: This only fixes the case when the values are concrete right? I'm fine with that, but we should document it perhaps? Because the symbolic version will wrap around, right?

src/EVM/Expr.hs
blishko: Right, I agree it is a bit confusing.
Not sure if there is something better we can do since the size might be symbolic.
It is definitely more conservative than what we had before. I think we want to be more conservative here, right?

src/EVM.hs
blishko: Yes, that is true.
How do you suggest to document it?
Also, I believe we have the same case for memory reads.

src/EVM/Expr.hs
msooseth: Sure, sorry, we definitely want to be conservative. I just didn't understand how this was useful? Maybe I need to read this again.

blishko: I think the point is that it will only skip the copy slice and go to `dst` buffer directly, if `dstOffset` is "far enough" from the wraparound boundary. Otherwise the expression does not simplify.

msooseth: OOOOO I understand now!!!

src/EVM.hs
msooseth: Maybe just above this `OpCalldataload` we can document it in a comment line?

blishko: I'll try

msooseth: Sounds great, thanks!

PR #1039, opened 2026-02-25
Optional Echidna compatibility check as a Github Action

gustavo-grieco:
## Description

The workflow is located at  hevm/.github/workflows/echidna-compat.yml. It performs the following steps when a Pull Request is opened or updated in the hevm repository:

1. Checkouts: It clones both the current hevm PR branch and the crytic/echidna master branch.
2. Patching: It uses a Python script to safely modify echidna/stack.yaml, replacing the remote hevm git dependency with the local checkout (./hevm).
3. Environment: It leverages Nix to obtain Stack and GHC, and ensure that hevm and echidna's C dependencies (like libsecp256k1 and libff) are correctly provided to the build environment.
4. Compilation: It runs stack build --fast --test --no-run-tests to verify that echidna still compiles with the changes made in the hevm PR.

Optional Status: The job uses continue-on-error: true, which means it will act as a warning. If compatibility is broken, the PR will show a warning icon but won't be blocked (unless you specifically mark this check as required in GitHub settings). The job will also post a status update with a build log in case the build fails.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/.github/workflows/echidna-compat.yml
+++ b/.github/workflows/echidna-compat.yml
@@ -46,30 +46,11 @@ jobs:
           with open('stack.yaml', 'r') as f:
               data = yaml.safe_load(f)
 
-          if 'extra-deps' in data:
-              new_deps = []
-              found_hevm = False
-              for dep in data['extra-deps']:
-                  # Identify if this is the remote hevm dependency
-                  is_hevm = (isinstance(dep, dict)
-                             and 'git' in dep
-                             and 'hevm' in dep['git'])
-                  if is_hevm:
-                      if not found_hevm:
-                          new_deps.append('./hevm')
-                          found_hevm = True
-                      continue
-                  if dep == './hevm':
-                      found_hevm = True
-                      new_deps.append(dep)
-                      continue
-                  new_deps.append(dep)
-
-              if not found_hevm:
-                  # Insert hevm at the beginning of extra-deps
-                  new_deps.insert(0, './hevm')
-
-              data['extra-deps'] = new_deps
+          deps = data.get('extra-deps', [])
+          deps = [d for d in deps if not (isinstance(d, str) and 'hevm' in d)
+                                  and not (isinstance(d, dict) and 'git' in d and 'hevm' in d['git'])]
+          deps.insert(0, './hevm')
+          data['extra-deps'] = deps
 
           with open('stack.yaml', 'w') as f:
               yaml.dump(data, f)


msooseth:
This would be really nice actually! So we know when we break Echidna :)

gustavo-grieco:
Btw, it is expected that this fails as echidna fails with the latest hevm revision

msooseth:
OH, it's good now? I'll review today :) Thanks!

msooseth:
@blishko do you also wanna have a go at this? I'm actually OK with it -- it doesn't add anything to hevm, only adds another hook to build. Ultimately, I think it's fine as long as it doesn't break by accident, and only breaks if it really ought to break because we broke Echidna :D

.github/workflows/echidna-compat.yml
```diff
@@ -0,0 +1,144 @@
...
+                      found_hevm = True
+                      new_deps.append(dep)
+                      continue
+                  new_deps.append(dep)
+
+              if not found_hevm:
+                  # Insert hevm at the beginning of extra-deps
+                  new_deps.insert(0, './hevm')
+
+              data['extra-deps'] = new_deps
```
blishko: Why is this complicated logic needed?
Isn't the goal just to point echidna to the local hevm version? Isn't there a simpler way to do it?

blishko:
Wouldn't it be better to just fail the workflow instead of marking it as successful and just posting a comment?
The comment can be difficult to spot. Failed workflow is spot immediately.

.github/workflows/echidna-compat.yml
gustavo-grieco: I think there is a simpler way to do it using `awk`/`sed` but it won't be robust if we change some newlines/comments. Maybe @elopez has more ideas.

gustavo-grieco:
@blishko this check is optional to avoid imposing a new burden to the hevm team. You can decide to fail the workflow in the future if you feel the ABI is stable enough.

blishko:
I would prefer the strong indication that this workflow is failing.
It is optional, true, so we can just decide to ignore the failure. But at least the failure will be visible.

.github/workflows/echidna-compat.yml
blishko: Why not just always prepend `./hevm` to existing dependencies? Why all the checks?

elopez: hevm is an existing dependency referenced via git, so then there would be two hevms listed potentially with the same version number. But I agree this script is too complicated for what it does. I simplified it a bunch on the latest commit.

elopez:
I removed the comment stuff now. The only con of just letting it fail is that whenever a breaking change is introduced to hevm, all commits, PRs, etc will then start to be marked with  ❌  in the GitHub UI. That might make it hard to tell apart actual failures that need attention from the known compatibility breakage.

.github/workflows/echidna-compat.yml
blishko: Great, thanks! This version is much more understandable!

blishko:
Having the breakage be more visible is better in this case, but that is my personal opinion.
Let's try and if it is too distractive, we can go back to the PR comment.

blishko:
BTW, what was the reason for the current breakage? Do you plan to fix it in Echidna soon?

blishko (approved):
LGMT!

blishko:
I think we can squash this to a single commit, no?

gustavo-grieco:
Yes

PR #1042, opened 2026-03-02
Test: Disallow FFI in forge tests by default

blishko:
It seems to me that by default FFI should be disabled.


msooseth:
YES. Thank you!

PR #1044, opened 2026-03-07
EVM: optimize OpSwap to use direct list manipulation

elopez:
## Description

Replace optics-based ix/zoom swap with direct splitAt-based list swap. The old code used two ix traversals + zoom + two .= assignments. The new code uses a single splitAt + list construction.

bench-perf shows a ∼10% reduction in primes and loop. The optimization was suggested by Claude Code.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

PR #1047, opened 2026-03-09
Tweak default RTS options

elopez:
## Description

This tweaks -A4m -> -A32m; see https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-flag-A-size

bench-perf shows 60%+ improvements across tests due to less time spent in GC. I haven't tested symbolic execution through hevm CLI but it's likely to benefit too.

For comparison, [we ship -A64m as default in echidna](https://github.com/crytic/echidna/blob/master/package.yaml#L101) but even higher values like -A1g show improvement for very large machines.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -381,7 +381,7 @@ benchmark bench-perf
   hs-source-dirs:
     bench
   ghc-options:
-    -O2 -threaded "-with-rtsopts=-N -A32m" -fproc-alignment=64
+    -O2 -threaded -rtsopts "-with-rtsopts=-N1 -A32m" -fproc-alignment=64
   other-modules:
     Paths_hevm
   autogen-modules:
--- a/hevm.cabal
+++ b/hevm.cabal
@@ -200,7 +200,7 @@ executable hevm
     cli
   main-is:
     cli.hs
-  ghc-options: -threaded -with-rtsopts=-N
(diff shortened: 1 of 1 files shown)


msooseth:
I'm not against this. In general, symbolic execution speed is mostly governed by SMT solver speed. So I'd say this is OK from my side. @blishko ?

msooseth (approved):
LGTM, but let's wait for Martin!

blishko:
Seems OK to me.
I trust @elopez expertise here; I don't plan to delve into GHC internals :)

blishko:
I would like to run `bench-perf` on my machine, just to confirm.
Let's merge https://github.com/argotorg/hevm/pull/1044 first and then this one.

blishko:
@elopez, can you rebase this one?
Let's focus on this one now.

blishko (approved):
Should be OK!

msooseth:
@elopez do you know why `Echidna with local hevm` is failing?

elopez:
@msooseth it's expected; Echidna does not build with current hevm main. It should resolve itself once we update hevm in Echidna (likely via https://github.com/crytic/echidna/pull/1499)

Issue #980, opened 2026-01-25
Internal Error when parsing user-defined enum types in function signatures


Mike4751:
## Description

When Echidna processes Solidity libraries with functions that have enum type parameters (e.g., `Order.OrderType`), hevm throws an internal error:

```
echidna: Internal Error: method type: Order.OrderType
```

This crash prevents Echidna from generating coverage reports, similar to #979.

## Reproduction

1. Create a Solidity library with functions that accept enum parameters:

```solidity
library Order {
    enum OrderType {
        MarketSwap,
        LimitSwap,
        MarketIncrease,
        MarketDecrease,
        // ...
    }

    function isMarketOrder(OrderType _orderType) internal pure returns (bool) {
        return _orderType == OrderType.MarketSwap ||
               _orderType == OrderType.MarketIncrease ||
               _orderType == OrderType.MarketDecrease;
    }
}
```

2. Deploy the library in Echidna's config (`deployContracts` in echidna.yaml)
3. Run Echidna and terminate (Ctrl+C)
4. Observe the internal error when coverage is being written

## Root Cause

In `src/EVM/ABI.hs`, the `basicType` parser only supports primitive Solidity types:

```haskell
basicType :: Vector AbiType -> P.Parsec () Text AbiType
basicType v =
  P.choice
    [ P.string "address" $> AbiAddressType
    , P.string "bool"    $> AbiBoolType
    , P.string "string"  $> AbiStringType
    , sizedType "uint" AbiUIntType
    , sizedType "int"  AbiIntType
    , sizedType "bytes" AbiBytesType
    , P.string "bytes" $> AbiBytesDynamicType
    , P.string "tuple" $> AbiTupleType v
    , P.string "function" $> AbiFunctionType
    -- No handler for user-defined types
    ]
```

When `parseTypeName` encounters `Order.OrderType`, all parsers fail and hevm throws an internal error.

## Proposed Fix

Add a fallback parser that treats unknown types as `uint8` (since Solidity enums are stored as uint8 internally):

```haskell
basicType :: Vector AbiType -> P.Parsec () Text AbiType
basicType v =
  P.choice
    [ P.string "address" $> AbiAddressType
    , P.string "bool"    $> AbiBoolType
    , P.string "string"  $> AbiStringType
    , sizedType "uint" AbiUIntType
    , sizedType "int"  AbiIntType
    , sizedType "bytes" AbiBytesType
    , P.string "bytes" $> AbiBytesDynamicType
[...]

PR #1055, opened 2026-04-02
Parse correctly enumerated types

gustavo-grieco:
## Description

Fix for #980 

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/test/test.hs
+++ b/test/test.hs
@@ -230,6 +230,15 @@ tests = testGroup "hevm"
         case readP_to_S (parseAbiValue (AbiBytesType 4)) threeBytes of
           [] -> pure ()  -- Expected: parsing should fail
           _ -> internalError "Should reject 3-byte value for bytes4"
+    , test "ABI-user-defined-enum-type" $ do
+        -- User-defined types like enums (e.g. "Order.OrderType") should parse as uint8
+        assertEqualM "qualified enum" (Just (AbiUIntType 8)) (parseTypeName mempty "Order.OrderType")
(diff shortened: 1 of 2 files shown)


msooseth:
This is nice actually! Can you maybe add some test cases? I think they'd be good as examples, too! Perhaps an LLM can dream up some nice use-cases :)

gustavo-grieco:
You mean more test cases, using complete solidity contracts?

msooseth:
Yes, perhaps we could add them to `test.hs` or if that doesn't work, then `clitest.hs` ? I prefer `test.hs` because it's faster, but either works :)

test/test.hs
```diff
@@ -230,6 +230,15 @@ tests = testGroup "hevm"
         case readP_to_S (parseAbiValue (AbiBytesType 4)) threeBytes of
           [] -> pure ()  -- Expected: parsing should fail
           _ -> internalError "Should reject 3-byte value for bytes4"
+    , test "ABI-user-defined-enum-type" $ do
```
blishko, line 233: Maybe we should rename this to `ABI-parsing-types`? And add tests for all the standard types?

blishko (commented):
Looks good to me!

test/test.hs
gustavo-grieco: To avoid this PR to stall, let's merge and address this in upcoming PR.

PR #1057, opened 2026-04-02
Allow to concretize dynamic arrays in calldata up to N values

### LINKED ISSUE #847: Implement concretization of dynamic data structures (e.g. dynamic arrays)
Dynamic array such as `bytes` are used to pass calldata. These are the source of a large number of exploits, where the data passed is not properly sanitized, yet fuzzers have a lot of trouble to produce reasonable values for them. In hevm, the current dynamic array make them intractable. Halmos instead, concertizes the size of the dynamic array up to a certain number selected by the user, and that's usually good enough.

gustavo-grieco:
## Description

This is an attempt to introduce size concretization for dynamic arrays to fix #847 with a similar approach to halmos. Some limitations: (1) it will only work with `calldata` parameters and (2) it will not work with recursive/complex dynamic types.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


msooseth:
Nice! Let's add 2-3 test cases where the calldata would need to be longer than 100 to be able to find the CEX. In these cases, the system should give the user an `Expr End Partial`, so the user knows that we couldn't explore everything. In other words, I want to assert via test cases, that the system doesn't give false sense of security just because the fault is beyond len 100 of the calldata.

test/test.hs
```diff
@@ -536,6 +536,117 @@ tests = testGroup "hevm"
...
+                  for (uint i = 0; i < 128; i++) {
+                    sum += uint8(data[i]);
+                  }
+                  assert(sum < 1000);
+                }
+              }
+            } |]
+        let sig = Just $ Sig "fun(bytes)" [AbiBytesDynamicType]
+        (e, ret) <- withDefaultSolver $ \s -> checkAssert s defaultPanicCodes c sig [] defaultVeriOpts
+        assertBoolM "must report partial due to bounded bytes" $ any isPartial e
```
msooseth: ❤️

msooseth:
Looking a LOT better, thanks!!!

msooseth:
I think this is okay-ish? @blishko do you wanna have a look? 

The only thing I don't like is the "insert a Partial in the result to warn" because it feels like that should be kinda automatic the way the system works. But it's possible that I am wrong, so I approve.

blishko:
@msooseth, you are bringing up a very good point. I don't understand why should the analysis be inserting this extra `Partial` result. Seems weird.

gustavo-grieco:
Maybe we need a different kind of partial type?

blishko:
Can someone give me a brief description of what would hevm do on these examples with `bytes calldata` arguments before this change?

Regarding the information to display to the user, I don't think this should be an extra `Partial` in the results.
It's not like there is a path in the program that we were not able to explore, no?
It's more like we are restricting the input values considered for *all* paths. So to me, it seems like this should be some kind of warning applied to the whole program, not an individual path.

gustavo-grieco:
> Can someone give me a brief description of what would hevm do on these examples with `bytes calldata` arguments before this change?

Pre-change behavior: symAbiArg had no case for AbiBytesDynamicType or AbiStringType, they fell through to the catch-all t -> internalError $ "TODO: symbolic abi encoding for " <> show t. So before this PR, calling hevm symbolic execution on any function whose signature contained bytes calldata or string calldata just crashed with that internal error. There was no handling at all.

> Regarding the information to display to the user, I don't think this should be an extra `Partial` in the results. It's not like there is a path in the program that we were not able to explore, no? It's more like we are restricting the input values considered for _all_ paths. So to me, it seems like this should be some kind of warning applied to the whole program, not an individual path.

On the warning vs. Partial: As you say, every path is fully explored; what's bounded is the universe of inputs. Treating that as a per-path Partial is a conceptual error. The natural representation is a single program-wide soundness caveat returned from verify (or attached to the result), not an extra entry in the [Expr End] list. What is the best approach, some kind of warning? We already have other constraints that limit the exploration but do not produce Partials (e.g. assume no reentrancy).

blishko:
> On the warning vs. Partial: As you say, every path is fully explored; what's bounded is the universe of inputs. Treating that as a per-path Partial is a conceptual error. The natural representation is a single program-wide soundness caveat returned from verify (or attached to the result), not an extra entry in the [Expr End] list. What is the best approach, some kind of warning? We already have other constraints that limit the exploration but do not produce Partials (e.g. assume no reentrancy).

My opinion is that it should not be a `Partial`, but  extra warning attached to the result.
No reentrancy is a bit different because it is an option that is opt-in.
I would say that the success result should have some extra (optional) field for "caveats" where this information would be stored and later shown to the user.

gustavo-grieco:
@blishko please take a look to the code to see if this matches your expectation. the output looks like this:

```
Checking 2 function(s) in contract src/unit-tests.t.sol:DynBoundDemo
[RUNNING] prove_beyond_bound(bytes)
   [FAIL] prove_beyond_bound

   [CAVEAT] Dynamic (bytes/string) arguments were bounded to 64 bytes; counterexamples requiring longer inputs may be missed.
[RUNNING] prove_within_bound(bytes)
   Counterexample:   [validated]
     calldata: prove_within_bound(0xff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000)
   [FAIL] prove_within_bound

   [CAVEAT] Dynamic (bytes/string) arguments were bounded to 64 bytes; counterexamples requiring longer inputs may be missed.
```

src/EVM/SymExec.hs
```diff
@@ -170,17 +183,38 @@ symCalldata sig typesignature concreteArgs base = do
     sizeConstraints
       = (Expr.bufLength withSelector .>= cdLen calldatas)
       .&& (Expr.bufLength withSelector .< (Lit (2 ^ conf.maxBufSize)))
-  pure (withSelector, sizeConstraints : props)
+    -- A dynamic (bytes/string) argument is turned into a 'Dy' fragment with its
+    -- length bounded to maxDynSize. That is a restriction of the input domain
+    -- (every path is still explored fully), so we record it as a program-wide
+    -- 'Caveat' rather than a per-path 'Partial'. Derived from the fragments we
+    -- actually built, so a concretely-supplied argument produces no caveat.
+    caveats = [DynArgBounded conf.maxDynSize | any isDy calldatas]
```
msooseth, line 191: I am not 100% sure we couldn't use the Partial system here? I am a bit confused. Can you explain how the Partial would not make sense here?

msooseth (commented):
It's actually cool but I'm a bit worried about the Caveat thing -- if we could do this with partials, we would avoid this complexity regarding Caveats. I mean, if the Caveats are completely unavoidable, then Caveats _could_ help us do interesting things, e.g. collect caveats about multi-transaction symbolic execution. But I just wanna make sure the Partial system really isn't sufficient here before adding complexity :)

src/EVM/SymExec.hs
gustavo-grieco: isn't this related to @blishko's comments or you mean renaming Caveats to Partial? 🤔

msooseth: Ah yes, shit. OK. I mean, I think I'm okay with this :) Let's dicuss over chat, but this looks reasonable I think.

msooseth:
Haha, this is too good and is breaking our expected fails:
```
2026-06-18T08:32:37.9130087Z hevm>   Dynamic-bytes-concretization
2026-06-18T08:32:37.9498591Z hevm>     bytes-length-cex:                                             FAIL (0.04s)
2026-06-18T08:32:37.9498981Z hevm>       test/test.hs:588:
2026-06-18T08:32:37.9499244Z hevm>       should be partial due to bounded bytes
2026-06-18T08:32:37.9499644Z hevm>       Use -p '/bytes-length-cex/' to rerun this test only.
2026-06-18T08:32:37.9945989Z hevm>     bytes-content-cex:                                            FAIL (0.05s)
2026-06-18T08:32:37.9946362Z hevm>       test/test.hs:603:
2026-06-18T08:32:37.9946623Z hevm>       should be partial due to bounded bytes
2026-06-18T08:32:37.9947016Z hevm>       Use -p '/bytes-content-cex/' to rerun this test only.
2026-06-18T08:32:38.0012808Z hevm>     bytes-no-cex:                                                 FAIL
2026-06-18T08:32:38.0013187Z hevm>       test/test.hs:616:
2026-06-18T08:32:38.0013455Z hevm>       should be partial due to bounded bytes
2026-06-18T08:32:38.0013796Z hevm>       Use -p '/bytes-no-cex/' to rerun this test only.
2026-06-18T08:32:38.0413577Z hevm>     bytes-with-uint-arg:                                          FAIL (0.04s)
2026-06-18T08:32:38.0413989Z hevm>       test/test.hs:631:
2026-06-18T08:32:38.0414252Z hevm>       should be partial due to bounded bytes
2026-06-18T08:32:38.0414644Z hevm>       Use -p '/bytes-with-uint-arg/' to rerun this test only.
```

Can you fix? Also, the Echidna compatibility check is failing due to the Caveat I think? It changes the API I believe

PR #1065, opened 2026-05-21
Expr: sound disjointness rule for readWord/WriteWord with bounded offset

msooseth:
## Summary

Adds a new `readWord` clause that elides a `WriteWord` whose offset is of the form `Add (Lit c) X`, when `X` has a derivable static upper bound and the write region is provably disjoint from the read in EVM (mod 2^256) arithmetic.

- New helper `upperBound :: Expr EWord -> Maybe W256` — conservative static upper bound handling `Lit`, `And`-with-`Lit`, `Mod`-by-`Lit`, `Div`-by-`Lit`, `Mul`/`Add` with one `Lit` operand, and `SHR` by a `Lit` shift.
- New helper `writeDisjointFromReadWord :: W256 -> Expr EWord -> Bool` — uses `Integer` arithmetic to detect wrap on both the read and write byte ranges before concluding the write can be skipped, so the rule does not assume non-wrapping addition.
- Hook in `readWord` at the `WriteWord` case.

Split out from #1063 so the disjointness rule is reviewable on its own; #1063 will be rebased on top of this.

## Tests

5 new unit tests in `test/EVM/Expr/ExprTests.hs`:

- positive: `WriteWord` at `Add(Lit, And-bounded)` skipped
- positive: `WriteWord` at `Add(Lit, Mul-And-bounded)` skipped
- negative: unbounded `X` — not skipped
- negative: wrap-risk near `maxBound` — not skipped
- negative: read region overlaps write min offset — not skipped

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs — N/A, internal simplifier rule
- [x] updated the changelog

### DIFF
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -418,7 +418,7 @@ readBytes (Prelude.min 32 -> n) idx buf
 
 -- | Static upper bound on a word expression (unsigned, no wrap), or Nothing.
 upperBound :: Expr EWord -> Maybe W256
-upperBound a = \case
+upperBound e = case (simplifyNoLitToKeccak e) of
   Lit n         -> Just n
   And (Lit m) _ -> Just m
   Mod _ (Lit b)
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -416,12 +416,62 @@ readBytes :: Int -> Expr EWord -> Expr Buf -> Expr EWord
 readBytes (Prelude.min 32 -> n) idx buf
   = joinBytes [readByte (add idx (Lit . unsafeInto $ i)) buf | i <- [0 .. n - 1]]
 
+-- | Static upper bound on a word expression (unsigned, no wrap), or Nothing.
+upperBound :: Expr EWord -> Maybe W256
+upperBound = \case
+  Lit n         -> Just n
+  And (Lit m) _ -> Just m
+  And _ (Lit m) -> Just m
+  Mod _ (Lit b)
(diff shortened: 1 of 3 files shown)


src/EVM/Expr.hs
```diff
@@ -416,12 +416,62 @@ readBytes :: Int -> Expr EWord -> Expr Buf -> Expr EWord
...
+    | Just ux <- upperBound x ->
+        let cI        = toInteger c
+            uxI       = toInteger ux
+            iI        = toInteger i
+            maxI      = toInteger (maxBound :: W256)
+            writeMaxI = cI + uxI + 31
+            readMaxI  = iI + 31
+        in readMaxI <= maxI
+           && writeMaxI <= maxI
+           && (cI >= iI + 32 || writeMaxI < iI)
```
blishko:
```suggestion
           && (cI > readMaxI || writeMaxI < iI)
```

blishko (commented):
Looks good!

I am suggesting one small change that I think makes the condition more readable.

src/EVM/Expr.hs
msooseth: Wait, doesn't that modify the code? 

```
readMaxI  = iI + 31
```

But you also changed `>=` to `>` ?

blishko: Yes.
`readMaxI = iI + 31` means you can rewrite `cI > readMaxI` as `cI > iL + 31`and that's equivalent to `cI >= iL + 32`, which is you original expression, no?

PR #1067, opened 2026-05-30
Add assertApproxEqRel cheatcode (uint256 and int256)

gustavo-grieco:
Implement foundry's assertApproxEqRel as a native hevm cheatcode, mirroring the existing assertApproxEqAbs. The relative percent delta is computed as |a - b| * 1e18 / b (denominator is the second/reference argument; 1e18 == 100%), and the assertion passes when percentDelta <= maxPercentDelta.

Matches foundry's native cheatcode semantics:
- int256 delta accounts for signs (same sign: ||a|-|b||, opposite: |a|+|b|), denominator is |b|
- b == 0 && a == 0 -> pass; b == 0 && a != 0 -> fail with "real delta: undefined"
- concrete path computes in arbitrary-precision Integer to avoid the intermediate overflow foundry sidesteps with U512
- best-effort symbolic path

Adds pass/fail test contracts and FoundryTests entries.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog

### DIFF
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -2298,24 +2298,19 @@ cheatActions = Map.fromList
     scale1e18 = 10^(18::Int)
     scale1e18W :: W256
     scale1e18W = 10^(18::Int)
-    -- largest absDelta for which (absDelta * 1e18) does not overflow 256 bits.
-    -- Used by the symbolic relative-equality path to detect when the truncating
-    -- Expr.mul would produce a wrong result.
+    -- largest absDelta with absDelta*1e18 < 2^256 (symbolic overflow guard)
     maxSafeRelDelta :: W256
     maxSafeRelDelta = (maxBound :: W256) `Prelude.div` scale1e18W
-    -- shared failure message for the relative-equality assertions; realDelta is
-    -- already rendered (a scaled integer, or the literal "undefined")
+    -- shared rel-assertion failure message; realDelta is pre-rendered
     relFailMsg :: Show a => a -> a -> Word256 -> String -> ByteString
     relFailMsg a b maxPercentDelta realDelta = "assertion failed: " <>
       BS8.pack (show a) <> " !~= " <> BS8.pack (show b) <>
       " (max delta: " <> BS8.pack (show maxPercentDelta) <>
       ", real delta: " <> BS8.pack realDelta <> ")"
     genAssert = genAssertN []
-    -- as genAssert, but ignores `extra` trailing ABI args (foundry's `err`-string
-    -- overloads). Used directly only for the string-operand variants; static
-    -- operands instead strip the trailing arg upstream via withMsg so they keep
-    -- the symbolic comparison path (decodeBuf yields no SAbi once a dynamic type
-    -- like the string is present in the signature).
+    -- like genAssert but ignores `extra` trailing ABI args (the err-string
+    -- overloads). Only used directly for string operands; static operands strip
+    -- the err arg upstream via withMsg to keep the symbolic path.
     genAssertN extra comp exprComp invComp name abitype sig input = do
       case decodeBuf ([abitype, abitype] ++ extra) input of
         (CAbi (a:b:_),"") | a `comp` b -> doStop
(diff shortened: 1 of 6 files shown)


msooseth:
Hey hey! Thanks :) Is this ready to review you think?

gustavo-grieco:
Actually, push one additional commit related with cheatcodes. It augments most of the cheatcodes with their versions that have `string` as the last parameter. In each case, the parameter is unused, except for showing in the console, so there is nothing else to implement.

src/EVM.hs
```diff
@@ -2255,17 +2293,50 @@ cheatActions = Map.fromList
...
     revertErr a b comp = frameRevert $ "assertion failed: " <>
       BS8.pack (show a) <> " " <> comp <> " " <> BS8.pack (show b)
-    genAssert comp exprComp invComp name abitype sig input = do
-      case decodeBuf [abitype, abitype] input of
-        (CAbi [a, b],"") | a `comp` b -> doStop
-        (CAbi [a, b],"")-> revertErr a b invComp
-        (SAbi [ew1, ew2],"") -> case (Expr.simplify (Expr.iszero $ exprComp ew1 ew2)) of
+    -- 1e18 (100% in foundry's 18-decimal fixed point) as Integer and W256
+    scale1e18 :: Integer
+    scale1e18 = 1000000000000000000
```
msooseth:
```suggestion
    scale1e18 = 10^18
```

src/EVM.hs
```diff
@@ -2255,17 +2293,50 @@ cheatActions = Map.fromList
...
+    scale1e18W :: W256
+    scale1e18W = 1000000000000000000
```
msooseth:
```suggestion
    scale1e18W = 10^18
```

src/EVM.hs
```diff
@@ -2331,6 +2402,76 @@ cheatActions = Map.fromList
...
+                percentDelta = (toInteger absDelta * scale1e18) `Prelude.div` toInteger b
+            in if percentDelta > toInteger (maxBound :: Word256)
+               then frameRevert "assertion failed: overflow in delta calculation"
+               else if percentDelta <= toInteger maxPercentDelta then doStop
+               else frameRevert $ relFailMsg a b maxPercentDelta (show percentDelta)
+        (SAbi [ew1, ew2, ew3],"") ->
+          -- percentDelta = (max(a,b) - min(a,b)) * 1e18 / b; check <= maxPercentDelta
+          let absDelta = Expr.sub (Expr.max ew1 ew2) (Expr.min ew1 ew2)
+              percentDelta = Expr.div (Expr.mul absDelta (Lit scale1e18W)) ew2
+          in case Expr.simplify (Expr.iszero $ Expr.leq percentDelta ew3) of
```
msooseth: This looks like a lot of overflow issues can happen here. Are we sure this is right? Looks a bit suspicious to me? The `Expr.mul` with `absDelta` and the `scale1e18W` sounds likes something that could overflow?

src/EVM.hs
```diff
@@ -2331,6 +2402,76 @@ cheatActions = Map.fromList
...
+               let percentDelta = (toInteger absDelta * scale1e18) `Prelude.div` toInteger absB
+               in if percentDelta > toInteger (maxBound :: Word256)
+                  then frameRevert "assertion failed: overflow in delta calculation"
+                  else if percentDelta <= toInteger maxPercentDelta then doStop
+                  else frameRevert $ relFailMsg a b maxPercentDelta (show percentDelta)
+        (SAbi [ew1, ew2, ew3],"") ->
+          -- best-effort symbolic handling (same-sign approximation, see assertApproxEqAbs)
+          let absDelta = Expr.sub (Expr.max ew1 ew2) (Expr.min ew1 ew2)
+              percentDelta = Expr.div (Expr.mul absDelta (Lit scale1e18W)) ew2
+          in case Expr.simplify (Expr.iszero $ Expr.leq percentDelta ew3) of
```
msooseth: Same worries as above

test/contracts/fail/assertApproxEqRel.sol
msooseth, line 1: We should add some  very large differences and see if it could overflow. I have a bad feeling about the overflow with Expr.mul&div

test/contracts/pass/assertApproxEqRel.sol
msooseth, line 1: We should add some potential overflow test cases here.

msooseth (changes requested):
Actually, the string-based asssertEq stuff is SUPER nice, i.e. `assertNotEq(bool,bool,string)` and friends. We have been missing these for forever, thank you for adding them. 

The approx versions I have some worry about in the symbolic setting. I think the concrete ones are fine.

We must also add some tests for the new assertEq cheatcodes that allow the user to add a message, e.g. `assertEq(bool,bool,string)`. These have been added but there are no tests for them I think? I guess an LLM will generate them quite easily, but I'd like to have them, just to be sure we don't break them.

src/EVM.hs
gustavo-grieco: Fixed

src/EVM.hs
gustavo-grieco: Fixed

gustavo-grieco:
@msooseth I have changed to code to fail on overflow, even if that mean allowing some FP. If your properties relies on an `assertApproxEqRel` overflow, then it needs to be fixed anyway.

msooseth:
Actually, a lot better! Can you please ask Claude to compress the comments, though? They are sooooo verbose :)

gustavo-grieco:
@msooseth done!

msooseth (approved):
LGTM

PR #1075, opened 2026-06-22
Two-phase SMT abstraction for symbolic division/modulo and multiplication

gustavo-grieco:
Combines the two stacked proof-of-concept PRs into a single branch against `main`, for easier review:

- #996 — uninterpreted div/mod for two-phase solving
- #1073 — abstract symbolic multiplication (stacked on #996)

Together they add an opt-in SMT **arithmetic abstraction** (`--abstract-arith`) that lets hevm prove multiply-then-divide properties — vault share prices, swap quotes, fixed-point math — that otherwise never return because the solver bit-blasts a 256-bit symbolic multiplier/divider.

## The problem

For symbolic `a`, `b` the only decision procedure SMT has for `a * b` / `a / b` over `(_ BitVec 256)` is bit-blasting: a full 256×256 multiplier (~10^5 gates) or a divider, per operation. Multiply two unknowns and divide by a third — the shape at the heart of vault accounting — and the query explodes (`unknown`/OOM).

## The approach (two phases)

1. **Abstract** `bvmul`/`bvudiv`/`bvsdiv`/`bvurem`/`bvsrem` over symbolic operands with uninterpreted functions (`abst_evm_*`).
2. **Constrain** with only *sound* facts:
   - **div/mod** is equated to the native op in the first query (with the EVM's `x/0 = 0`), next to an absolute-value + shift-bound + congruence encoding (unsigned results exact; signed reconstructed from magnitudes). Division is therefore still bit-blasted.
   - **multiplication** is uninterpreted in the first query, plus a catalogue of sound algebraic lemmas: commutativity, 0/1 identities, mul/div/divisor monotonicity, the div–mul link `(a/b)*b ≤ a`, mulDiv bounds, and the constant-scaling family — const-cancellation `(c1·x)/c2 == (c1/c2)·x`, fraction-reduce, nested-division collapse, ceilDiv-cancel, and scaled-product telescoping.
3. **Refine**: if the first query is SAT and contains an abstract product, the multiplication ground truth is asserted into the same running solver and it is checked again. A counterexample is only reported from that refined check.

### Soundness

Every lemma is true of ordinary arithmetic, so anything proved from them holds for the real operations (in effect you prove the property about a more general contract and the deployed one is a special case). `abst_evm_bvmul` has **no** ground truth in the first query, so a satisfying model there may assign products values real multiplication never would. When that query is SAT and contains an abstract product, hevm asserts every abstract product equal to native `bvmul` into the same running solver and checks again, so a reported counterexample uses real multiplication. The re-check bit-blasts the products, so it may come back Unknown.

## Structure

The lemma machinery lives in three modules so the lemmas are reviewable in one place:

- `EVM.SMT.AbstractBase` — SMT primitives, signed helpers, the div/mod taxonomy, term collectors/matchers, and `saturate` (closes the set of div/mul terms the lemmas range over).
- `EVM.SMT.AbstractLemmas` — the lemma catalogue: a `LemmaInst` sum type, `collectLemmas` (triggers), and `emitLemma` (SMT emission + per-lemma soundness note), one banner per lemma; GHC exhaustiveness ties each constructor to its emitter.
- `EVM.SMT.DivModEncoding` — orchestration (`mulEncoding`) + the div/mod ground-truth encoding.

## Usage

Opt-in via `hevm test --abstract-arith` (off by default); also drivable through echidna's verification mode.

## Validation

Representative properties prove under `--abstract-arith` (const-cancel, scaled const-cancel, fraction-reduce, nested-div, mul/div monotonicity, mulDiv bound, telescoping), and deliberately false identities correctly fail, confirming no lemma over-fires. The abstraction has been used to verify the full stateless conversion math of a production vault (Spark PSM3) and a large subset of the ABDK Math 64.64 property suite.

Supersedes #996 and #1073 (left open for now).

## Try it on a real target: spark-psm

The Spark PSM3 verification mentioned above is public and wired for one-command repro, so it is easy to experiment with and without the abstraction on real production code:

```bash
git clone --recurse-submodules -b symbolic-conversion-proofs https://github.com/gustavo-grieco/spark-psm
cd spark-psm    # needs foundry + bitwuzla on PATH, and hevm built from this branch
```

```bash
# all ~60 prove_* properties, with the abstraction (~30 min; most prove in seconds)
make verify-hevm

# one property or a family, by name regex
make verify-hevm M=prove_convertToAssets_susds
make verify-hevm M=roundtrip

# the same queries WITHOUT the abstraction: mul/div bitblast natively and the
# nonlinear properties come back unknown instead of PASS
make verify-hevm M=roundtrip HEVM_FLAGS="--solver bitwuzla --smt-timeout 300 --max-iterations 5000"
```

`make verify-hevm` runs `forge build --ast` followed by `hevm test --root . --match "<M>" --solver bitwuzla --abstract-arith --smt-timeout 300 --max-iterations 5000`; set `HEVM=/path/to/hevm` to pick the binary. Every `prove_` assumes only the abstraction's operand budget, `require(x <= type(uint128).max)` per input. The properties in `test/symbolic/ProveOriginals.t.sol` delegate to the repo's **unmodified** `testFuzz_*` functions, so what is proven is Spark's own test content; `test/symbolic/README.md` documents the full property set, the lemmas each one exercises, and the eleven known-out-of-reach cases.


msooseth:
Ah, thankfully the majority of the LoC change is in `test.hs`, I was getting extremely worried for the +1000 LoC.... :laughing:

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+
+-- | Uninterpreted function declarations for abstract div/mod.
+divModAbstractDecls :: [SMTEntry]
+divModAbstractDecls =
+  [ SMTComment "abstract division/modulo (uninterpreted functions)"
+  , SMTCommand "(declare-fun abst_evm_bvsdiv ((_ BitVec 256) (_ BitVec 256)) (_ BitVec 256))"
+  , SMTCommand "(declare-fun abst_evm_bvsrem ((_ BitVec 256) (_ BitVec 256)) (_ BitVec 256))"
+  , SMTCommand "(declare-fun abst_evm_bvudiv ((_ BitVec 256) (_ BitVec 256)) (_ BitVec 256))"
+  , SMTCommand "(declare-fun abst_evm_bvurem ((_ BitVec 256) (_ BitVec 256)) (_ BitVec 256))"
+  , SMTComment "abstract multiplication (kept fully uninterpreted: no ground truth)"
```
msooseth: Let's have a single comment at the top.

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+  , SMTCommand "(declare-fun abst_evm_bvmul ((_ BitVec 256) (_ BitVec 256)) (_ BitVec 256))"
+  ]
+
+-- | Local helper for trivial SMT constructs
+sp :: Builder -> Builder -> Builder
+a `sp` b = a <> " " <> b
```
msooseth: Wait, we have this sp already somewhere? Yeah, in SMT:

```
sp :: Builder -> Builder -> Builder
a `sp` b = a <> (fromText " ") <> b
```

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+
+zero :: Builder
+zero = "(_ bv0 256)"
+
+one :: Builder
+one = "(_ bv1 256)"
+
```
msooseth: We have these in SMT.hs

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+-- The cost is completeness — a product of operands above 2^128 that happens not
+-- to overflow will not receive the lemma. Callers bound operands (e.g. a
+-- Solidity @require(x < 2**128)@) so the guard discharges cheaply.
+mulNoOverflow :: Builder -> Builder -> Builder
+mulNoOverflow x y =
+  "(and (bvule " <> x <> " " <> maxU128 <> ") (bvule " <> y <> " " <> maxU128 <> "))"
+
+wordAsBV :: forall a. Integral a => a -> Builder
+wordAsBV w = "(_ bv" <> Data.Text.Lazy.Builder.Int.decimal w <> " 256)"
+
```
msooseth: This already exists:

```
wordAsBV :: forall a. Integral a => a -> Builder
wordAsBV w = "(_ bv" <> Data.Text.Lazy.Builder.Int.decimal w <> " 256)"
```

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+-- | Collect all div/mod operations from an expression.
+collectDivMods :: Expr a -> [DivModOp]
+collectDivMods = \case
+  T.SDiv a b -> [(IsSDiv, a, b)]
+  T.SMod a b -> [(IsSMod, a, b)]
+  T.Div  a b -> [(IsUDiv, a, b)]
+  T.Mod  a b -> [(IsUMod, a, b)]
+  _          -> []
+
+-- | Collect symbolic*symbolic multiplications (neither operand a literal).
```
msooseth: No need for obvious comments. Let's ask the LLM to trim all comments down/remove the simple ones :)

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,286 @@
...
+
+-- | Collect multiplications by a non-trivial literal constant @(c, x)@.
+collectConstMuls :: Expr a -> [(W256, Expr EWord)]
+collectConstMuls = maybe [] pure . asConstMul
+
+-- | True if any prop contains a symbolic*symbolic multiplication, i.e. the
+-- multiplication abstraction is in play. Because abst_evm_bvmul is left
+-- uninterpreted (no ground truth), a satisfying model may assign it values
+-- inconsistent with real multiplication, so a counterexample cannot be
+-- trusted; callers must downgrade SAT results to Unknown to stay SOUND.
```
msooseth: Hmmmm so we return Unknown? But we could run it concretely to get a real value? Or the idea is that we should get UNSAT for correct contracts,  and then it's fine, otherwise, it's Unknown and we warn the user that there MAY be an issue?

src/EVM/SMT/AbstractLemmas.hs
```diff
@@ -0,0 +1,275 @@
...
+  | DivMono       (Expr EWord) (Expr EWord) (Expr EWord)               -- ^ x<=y => x/z<=y/z
+  | DivisorMono   (Expr EWord) (Expr EWord) (Expr EWord)               -- ^ y1<=y2 => x/y2<=x/y1
+  | MulDivBound   (Expr EWord) (Expr EWord) (Expr EWord) (Expr EWord)  -- ^ (x*y)/z <= x or y
+  | ConstMulMono  W256 (Expr EWord) (Expr EWord)                       -- ^ x<=y => c*x<=c*y
+  | ConstCancel   (Expr EWord) W256 W256 (Expr EWord)                  -- ^ (c1*x)/c2 == (c1/c2)*x
+  | NestedDiv     (Expr EWord) W256 W256                               -- ^ (A/c1)/c2 == A/(c1*c2)
+  | FracReduce    (Expr EWord) W256 W256 (Expr EWord)                  -- ^ (c1*x)/c2 == x/(c2/c1)
+  | CeilDivCancel (Expr EWord) W256 W256 (Expr EWord)                  -- ^ ceilDiv(c1*x,c2) inner divide
+  | Telescope     (Expr EWord) (Expr EWord) W256 W256                  -- ^ (a*b)/c - (a*(b-k))/c == a*(k/c)
+  deriving (Eq, Ord)
```
msooseth, line 43: Are we sure we need all of these? I'd rather not review all if we don't need all :) The LLMs often add more than what's needed. I've had to fight this quite a bit. Have we tried removing some to see if the contract in question (Spark’s PSM3) would still validate? For the `src/EVM/SMT/DivModEncoding.hs` Martin and I cleaned up a lot of these, as it turned out that quite a few were not needed.

msooseth (changes requested):
My light review :) First of all, this is very-very cool! I just wanna make sure I can read it all and understand it all, so we can merge!

 I think most of the things I highlighted could be fixed via LLM relatively easily? I hope at least! Basically, stuff I think can be maybe trimmed down? Perhaps the SMT helper functions could be extracted to another module, and shared? And the lemmas should be checked if they are all needed. I'd like to make sure only to review the ones that are needed :) They are hard to check so I wanna economise on my time :) Perhaps an LLM can figure out which ones are absolutely necessary and then we can review those only :)

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Fixed

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Fixed

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Fixed

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Claude says:
> Exactly your second reading: the abstraction is for the *proof* direction. For a correct contract the query is UNSAT and that's a real proof (the lemmas are over-approximating, so UNSAT is sound). If the solver says SAT while an abstract product is present, the model may assign `abst_evm_bvmul` values real multiplication never produces, so we downgrade it to `Unknown` with an explicit message ("counterexample unsound under multiplication abstraction", Solvers.hs) — i.e. "there MAY be an issue, rerun without `--abstract-arith` to get a trustworthy counterexample". The downgrade is precise: it only kicks in when a symbolic*symbolic product is actually in the query (`hasAbstractMul`); div/mod-only counterexamples keep their ground truth and are still reported as real cexes (there's a regression test for that: `div-cex-preserved`).

src/EVM/SMT/AbstractLemmas.hs
gustavo-grieco: Claude response after running some tests:

> Fair concern :) Some background: the catalogue wasn't designed up-front — each family was added when a concrete property refused to prove without it (the core mul families came from the vault patterns now in test.hs, the constant-scaling families from specific spark-psm swap legs), so none of them is speculative. But "was added for a reason at the time" isn't the same as "still needed given all the others", so we've started exactly the experiment you suggest: a local build with a switch that disables one lemma family at a time, re-running the spark-psm suite (60 prove_* properties, all passing at baseline) per family.

> Preliminary results so far: every removal we've tried loses properties. For example, disabling just commutativity breaks (at least) five — prove_shares_monotonic, prove_roundtrip_shares, prove_roundtrip_no_inflation, prove_assetValue_backing_monotonic all come back unknown, and prove_getSUsdsValue_roundUp_monotonic fails in a more interesting way: the solver finds a satisfying model in which abst_evm_bvmul(a,b) != abst_evm_bvmul(b,a). That second failure mode is definitive — SAT means the remaining lemmas provably cannot prove the property at any timeout (the soundness guard correctly reports it as Unknown rather than as a counterexample). With the mul lemmas off wholesale, even the basic monotonicity properties degrade the same way.

> We're still running the remaining families; happy to post the full property × lemma-family matrix here once it's done, and if any family turns out unnecessary for both spark-psm and the targeted tests in test.hs, I'll drop it from the PR.

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Fixed

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Fixed

src/EVM/SMT/AbstractLemmas.hs
msooseth: OK, that's great feedback, thank you! Resolving.

src/EVM/SMT/AbstractBase.hs
msooseth: OK, nice, thank you!

src/EVM/SMT/AbstractBase.hs
```diff
@@ -0,0 +1,228 @@
...
+-- are handled natively, so only genuinely symbolic products are abstracted.
+asMul :: Expr a -> Maybe (Expr EWord, Expr EWord)
+asMul (T.Mul x y) | notLit x, notLit y = Just (x, y)
+asMul _ = Nothing
+
+-- | A product by a non-trivial literal constant: c*x or x*c (excluding 0, 1).
+-- These stay native @bvmul@; the const-mul lemmas range over them.
+asConstMul :: Expr a -> Maybe (W256, Expr EWord)
+asConstMul (T.Mul (Lit c) x) | notLit x, c /= 0, c /= 1 = Just (c, x)
+asConstMul (T.Mul x (Lit c)) | notLit x, c /= 0, c /= 1 = Just (c, x)
```
msooseth: I thought we normalized these and Lit is ALWAYS the first one? We should spit out `error` in case it's not I think. I'd rather do that. Maybe ask the AI to do that for all of them?

src/EVM/Solvers.hs
```diff
@@ -128,9 +128,30 @@ checkSatWithProps sg props = do
...
+        let x = getNonError smt2Abstract <> SMT2 (SMTScript (getNonError refinement)) mempty mempty
+        res <- checkSat sg (Just props) (Right x)
+        -- SOUNDNESS: multiplication is abstracted as an uninterpreted function
+        -- with no ground truth, so a satisfying model may use product values
+        -- inconsistent with real multiplication. A QED stays sound (the lemmas
+        -- over-approximate), but a counterexample may be spurious, so downgrade
+        -- it to Unknown.
+        pure $ case res of
+          Cex _ | hasAbstractMul allProps ->
+            Unknown "counterexample unsound under multiplication abstraction (abst_evm_bvmul is uninterpreted)"
```
msooseth: Shouldn't we refine here instead of giving up? Like, we could inject the ground truth lemmas and then wait for a Cex?

src/EVM/Solvers.hs
```diff
@@ -128,9 +128,30 @@ checkSatWithProps sg props = do
...
+      if isLeft smt2Abstract then pure $ Error $ getError smt2Abstract
+      else if isLeft refinement then pure $ Error $ getError refinement
+      else do
+        let x = getNonError smt2Abstract <> SMT2 (SMTScript (getNonError refinement)) mempty mempty
+        res <- checkSat sg (Just props) (Right x)
+        -- SOUNDNESS: multiplication is abstracted as an uninterpreted function
+        -- with no ground truth, so a satisfying model may use product values
+        -- inconsistent with real multiplication. A QED stays sound (the lemmas
+        -- over-approximate), but a counterexample may be spurious, so downgrade
+        -- it to Unknown.
```
msooseth: Too verbose, let's cut it :)

src/EVM/Solvers.hs
```diff
@@ -264,6 +285,9 @@ getMultiSol solver timeout maxMemory smt2@(SMT2 cmds cexvars _) multiSol r sem f
     )
 
 getOneSol :: (MonadIO m, ReadConfig m) => Solver -> Maybe Natural -> Natural -> SMT2 -> Maybe [Prop] -> Chan SMTResult -> TChan CacheEntry -> QSem -> Int -> m ()
+-- the empty solver answers every query "unknown" without spawning a process
+getOneSol EmptySolver _ _ _ _ r _ _ _ =
+  liftIO $ writeChan r (Unknown "Result unknown by SMT solver")
```
msooseth, line 287: Ah, this is great, I like it :)

msooseth (changes requested):
I am OK with accepting this, but we need a fuzzer for this specific shape of arithmetic problems. Can you make the AI write a version of our Expr fuzzer that is specific for the shapes that this thing rewrites? I am pretty sure the one that we are using right now is barely firing, as our generation is very generic and not tuned towards arithmetic things specifically. It should take no more than 1h to create a fuzz test that does exactly that, using a different distribution of generative elements in the Expr, so the shape that this generates can be more effectively fuzzed. I am worried that there are bugs hiding here, but we can't find them. It's hard to review this.

src/EVM/Solvers.hs
gustavo-grieco: Done!

src/EVM/Solvers.hs
gustavo-grieco: Done!

src/EVM/SMT/AbstractBase.hs
gustavo-grieco: Done!

gustavo-grieco:
@msooseth the new fuzzing code is done already and found a real signed div/mod-by-zero soundness bug. It's fixed now and the input was added as regression test. I have run the test for 10K cases without failure.

msooseth:
I think this is getting there. I'd like to have @blishko take one more look. It's a bit exotic and quite a bit of change. I know we are holding this up, but it's also not my codebase, and I don't want to be responsible for some serious mistake here, and it's quite a bit of change.

gustavo-grieco:
@msooseth thanks for rebasing this into the new SMT solver branch

PR #1080, opened 2026-07-21
Decode cheatcode string args leniently instead of crashing on invalid UTF-8

### LINKED ISSUE #1076: Cheatcode string arguments crash on invalid UTF-8
## Summary

A Solidity `string` argument passed to hevm cheatcodes is not guaranteed to be valid UTF-8. Passing an invalid byte through a cheatcode string path can crash vanilla Echidna/hevm with a partial UTF-8 decode exception.

## Reproducer

Minimal repro: https://github.com/forkforkdog/echidna-invalid-utf8-trace-repro/blob/main/README.md

The repro passes an invalid Solidity `string` containing byte `0xbf` into `vm.setEnv`, then reads the same key with `vm.envString`.

Run command from the repro README:

```sh
docker run --rm --entrypoint bash \
  -v "$PWD:/work" \
  -v "$PWD/bin/echidna:/usr/local/bin/echidna:ro" \
  -w /work guardiansentry-rpc-e2e-task-worker-slow:latest \
  -lc 'echidna --version && echidna src/InvalidUtf8Trace.sol --contract InvalidUtf8Trace --config echidna.yaml'
```

## Actual behavior

```text
[Worker 0] Crashed:
Cannot decode byte '\xbf': Data.Text.Encoding: Invalid UTF-8 stream
```

## Expected behavior

hevm should not crash on arbitrary ABI string bytes supplied to cheatcodes. Reasonable outcomes would be preserving the raw bytes for lookup, returning a controlled cheatcode error, or rendering invalid bytes with escaping.

## Notes

Solidity ABI `string` values are byte sequences at runtime; the EVM does not enforce UTF-8 validity. This means cheatcode string argument handling should avoid partial `decodeUtf8` paths when handling user-provided `string` calldata.

## Related fix

A downstream Echidna patch fixes this by making cheatcode string rendering tolerate invalid UTF-8: https://github.com/GuardianAudits/echidna_custom/commit/77d8a300891c82034d4bce665c0298f685eb1614

gustavo-grieco:
Fixes #1076.

## Problem

Solidity ABI `string` values are raw byte sequences at runtime — the EVM does not enforce UTF-8 validity. The cheatcode handlers decoded string arguments with the partial `Data.Text.Encoding.decodeUtf8`, so passing a string containing an invalid byte (e.g. `0xbf`) through `vm.setEnv` / `vm.envString` / `vm.label` crashed the entire run:

```
Cannot decode byte '\xbf': Data.Text.Encoding: Invalid UTF-8 stream
```

This is reachable from fuzzers driving hevm (the report came from an Echidna campaign generating arbitrary string calldata).

## Fix

Two call sites, three lines of logic: replace `decodeUtf8` with `decodeUtf8With lenientDecode` in

- `toString` — the shared helper all `env*` cheatcodes, `setEnv` and `ffi` funnel through, and
- the `label(address,string)` handler.

Invalid bytes decode to U+FFFD instead of throwing. `setEnv`/`envString` stay consistent because both sides decode the key the same way, so lookups still round-trip.

These are the only partial-decode sites reachable from cheatcode string arguments: `createFork` already uses `Char8.unpack`, and assert messages / trace rendering go through the total `decodeUtf8'`-with-hex-fallback path in `EVM.Format`.

## Performance / scope

`decodeUtf8With lenientDecode` is the same single validating pass as `decodeUtf8` — no extra cost for valid strings, no behaviour change for valid UTF-8.

## Testing

- New `prove_invalidUTF8` in `test/contracts/pass/cheatCodes.sol` (runs in the `Cheat-Codes-Pass` group), mirroring the issue reproducer: `setEnv` with a `0xbf` key, `envString` round-trip, `label` with invalid bytes.
- Verified on the reproducer project: pre-fix binary crashes with the exact error above; post-fix it reports `[PASS]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Fixed
+- Cheatcode string arguments are now decoded leniently instead of crashing on
+  invalid UTF-8: ABI `string` values are raw bytes, so e.g. `vm.setEnv`,
+  `vm.envString` and `vm.label` no longer abort the whole run when handed a
+  byte sequence that is not valid UTF-8 ([#1076](https://github.com/argotorg/hevm/issues/1076))
+
 ## [0.58.0] - 2026-06-26
 
 ## Added
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -52,7 +52,8 @@ import Data.Set (insert, member, fromList)
 import Data.Sequence (Seq)
 import Data.Sequence qualified as Seq
 import Data.Text (unpack, pack)
-import Data.Text.Encoding (decodeUtf8)
+import Data.Text.Encoding (decodeUtf8With)
+import Data.Text.Encoding.Error (lenientDecode)
 import Data.Tree
 import Data.Tree.Zipper qualified as Zipper
 import Data.Typeable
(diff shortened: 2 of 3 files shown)


blishko (approved):
LGTM!

PR #1083, opened 2026-07-23
Fix #1082: resolve symbolic-key SLOAD over a ConcreteStore

### LINKED ISSUE #1082: Symbolic-key `SLOAD` over a `ConcreteStore` is left unresolved, causing non-termination in symbolic execution
## Summary

`EVM.Expr.readStorage` does not resolve a read at a **symbolic slot** from a **`ConcreteStore`**. A `ConcreteStore` is a *total* map — every slot that was never written reads `0` — but `readStorage` only exploits this for **literal** slots. For a symbolic slot it returns an unresolved abstract `SLoad`, discarding the fact that the store's contents (and its `0` default) are fully known.

Downstream, this makes the symbolic interpreter **fork on an effectively-known value**. In the degenerate case of an *empty* `ConcreteStore`, `s[x]` for symbolic `x` is provably `0`, yet the interpreter treats it as unconstrained, explores an infeasible branch, and **fails to terminate** (no SMT query is ever emitted — it hangs in exploration).

## Root cause

`src/EVM/Expr.hs`, `readStorage`:

```haskell
readStorage w st = go (simplifyNoLitToKeccak w) (simplifyNoLitToKeccak st)
  where
    go slot s@(AbstractStore _ _) = Just $ SLoad slot s
    go (Lit l) (ConcreteStore s)  = Lit <$> Map.lookup l s      -- literal slot: resolved (absent ⇒ Nothing ⇒ readStorage' returns Lit 0)
    go slot store@(ConcreteStore _) = Just $ SLoad slot store   -- SYMBOLIC slot: returned UNRESOLVED  ← bug
    ...
```

A `ConcreteStore` read is total: unwritten slots are `0`. The literal-slot case reflects that (an absent key yields `Nothing`, and `readStorage'` turns that into `Lit 0`). The symbolic-slot case does not — it returns `SLoad slot (ConcreteStore …)`, which is unconstrained.

The comment above `readStorage` even notes the related deferral:

> This does not strip writes that cannot possibly match a read … in case things in between cannot be stripped … Note that decomposition solves this, though late in the simplification lifecycle (just before SMT generation, **which can be too late**).

It is too late: the interpreter forks and diverges *before* SMT generation / decomposition ever runs.

## Minimal reproduction (Expr level, hevm-native)

```haskell
import EVM.Expr (readStorage')
import EVM.Types
import qualified Data.Map as Map
import qualified Data.ByteString as BS

-- mapping slot keccak(abi.encode(x, 0)) with a symbolic key x:
let slot = Keccak (WriteWord (Lit 0) (Var "x") (ConcreteBuf (BS.replicate 64 0)))

-- literal slot over an empty ConcreteStore — correctly 0:
readStorage' (Lit 42) (ConcreteStore mempty)
-- ==> Lit 0x0                                   ✅

-- symbolic slot over the SAME empty ConcreteStore — should also be 0:
readStorage' slot (ConcreteStore mempty)
-- ==> SLoad (Keccak (WriteWord (Lit 0x0) (Var "x") …)) (ConcreteStore (fromList []))   ❌
--     expected: Lit 0x0
```

The two reads are over the same (empty) store and must both be `0`; only the literal one is resolved.

## End-to-end manifestation (non-termination)

A tool that runs a contract's constructor concretely and then symbolically explores a method (e.g. Echidna's symbolic-execution mode, which `resetState`s the concrete fuzzing VM and hands it to `verify`/`verifyInputsWithHandler`) supplies the contract's storage as a `ConcreteStore`. Any function that does a symbolic-key mapping read then hangs:

```solidity
pragma solidity ^0.8.20;

contract SymKeyStuck {
    mapping(uint256 => uint256) s;

    // symbolic key over an (empty) ConcreteStore  -> hangs: no opcode after the
    // SLOAD branch ever steps again, no SMT query is emitted, coverage frozen.
    function prove_symbolic_key(uint256 x) external view { assert(s[x] == 0); }

    // literal key -> resolves instantly (control).
    function prove_concrete_key(uint256 x) external view { x; assert(s[5] == 0); }
}
```

Instrumenting the interpreter shows it reaches the `SLOAD` (`SLoad (Keccak (WriteWord 0 (Var "arg1") …)) (ConcreteStore (fromList []))`), forks on `s[x] == 0`, and then never makes progress — `0` SMT queries, frozen coverage.

Note: hevm's own CLI does **not** exhibit this, because `hevm symbolic` seeds storage with `AbstractStore` (`--initial-storage Abstract`, the default) or `EmptyBase` (`--initial-storage Empty`) — both of which `readStorage` handles correctly (`AbstractStore` ⇒ genuinely symbolic value; `EmptyBase` ⇒ `0`). The bug only surfaces when a `ConcreteStore` is fed into symbolic execution, which the CLI never does but embedding tools do.

## Expected vs. actual

| slot | store | expected | actual |
|---|---|---|---|
| `Lit 42` | `ConcreteStore {}` | `Lit 0` | `Lit 0` ✅ |
| `Keccak(sym)` | `ConcreteStore {}` | `Lit 0` | `SLoad …` ❌ |
| `Keccak(sym)` | `ConcreteStore {k↦v, …}` | `ITE (sym=k) v (… 0)` | `SLoad …` ❌ |

## Suggested fix

Resolve symbolic-slot reads over a `ConcreteStore` the same way literal reads are resolved — an empty store reads `0`; a populated one is an `ITE` chain over its entries defaulting to `0`:

```haskell
    go slot s@(AbstractStore _ _) = Just $ SLoad slot s
    go (Lit l) (ConcreteStore s) = Lit <$> Map.lookup l s
    -- A ConcreteStore is total: unwritten slots read 0. Reflect that for symbolic
    -- slots too, otherwise readStorage returns an unresolved SLoad and the symbolic
    -- interpreter forks on an effectively-known value and can fail to terminate.
    go slot (ConcreteStore m)
      | Map.null m = Just $ Lit 0
      | otherwise  = Just $ Map.foldrWithKey (\k v acc -> ITE (Eq slot (Lit k)) (Lit v) acc) (Lit 0) m
```

With this, the reproduction above verifies in <1 s. (`readStorage'` for external/RPC-backed contracts is unaffected — those go through `accessStorage`/the fetcher, not `readStorage'`, and the literal-slot `Nothing`-for-RPC behavior on line for `Lit` slots is left unchanged.)

## Environment

- hevm `0.57.0` (git `46bcd0da`)
- Surfaced via Echidna symbolic-execution / verification mode.

gustavo-grieco:
Fixes #1082.

## Problem

`readStorage` returns an unresolved abstract `SLoad` for a symbolic slot over a `ConcreteStore`, even though the store's contents are fully known. The symbolic interpreter then forks on an effectively-known value. Two symptoms:

- **Non-termination**: observed through echidna's verification mode (concrete deployment populates a `ConcreteStore`, then a symbolic mapping read never resolves: 0 SMT queries, frozen coverage).
- **Blowup on populated stores**: standalone `hevm test`, a symbolic-key read of a *never-written* mapping over a store holding 600 entries of an unrelated mapping does not complete within 120s. All those entries are dragged into the fork and the SMT encoding even though none of them can alias the read.

## Fix

Two pieces. Notably this is **not** the ITE-over-all-entries fold proposed in the issue: that would inline the whole store into every symbolic read and trades non-termination for OOM on populated stores.

1. **An empty `ConcreteStore` reads 0 at every slot, including symbolic ones.** A `ConcreteStore` is total (unwritten slots are 0), so with no entries every read is 0. This lives in `readStorage'` (the total variant used by the interpreter's internal-storage path and by expression simplification), **not** in `readStorage`: `accessStorage` relies on `readStorage`'s `Nothing`/`SLoad` results to drive lazy RPC slot fetching for `external` contracts, whose `ConcreteStore` is a partial cache rather than the full storage. Folding the rule into `readStorage` would make every first read of an unfetched slot on a forked contract silently return 0 instead of issuing `PleaseFetchSlot`. A new unit test (`readStorage-keeps-nothing-on-concrete-miss`) pins that contract.

2. **At read time (`OpSload`), filter the `ConcreteStore` to entries that can belong to the read's mapping**, using the keccak preimages recorded during concrete execution: a concrete key whose known preimage identifies a *different* mapping slot cannot equal the symbolic read slot, by keccak injectivity — the same assumption family the store simplifier already uses (`idsDontMatch`, `Lit < 256` vs `Keccak`). A read of an unwritten or unrelated mapping filters to empty → `Lit 0`, before any fork. Keys with unknown preimage are kept (conservative, sound). The filtered store is only used for the one read expression and is never written back.

This incidentally fulfills the existing `word-eq-bug` test's `TODO: This should be simplified to (Lit 0)` — its expectation is updated accordingly.

## Tests

- **Expr unit tests** (`test/EVM/Expr/ExprTests.hs`): empty-store rule for symbolic slots, preserved `Nothing`-on-concrete-miss contract of `readStorage`, write-stripping down to an empty base, and `filterStoreByReadSlot` drop/keep/ignore cases (using a real `keccak'` preimage).
- **Foundry tests** (`test/EVM/Test/FoundryTests.hs`):
  - `pass/symbolicKeySload.sol`: the issue's literal empty-store repro, plus a 600-entry populated mapping with a symbolic read of an untouched mapping. Without the fix the latter does not complete (this is the regression signal); with it, ~1s. Both contracts use `bool public constant IS_TEST` deliberately — a scalar state var would keep a preimage-less slot in the store and route the read through the solver instead of resolving to 0.
  - `fail/symbolicKeySload.sol`: soundness guard — a symbolic read of the *same* mapping that holds concrete entries (10, enough to be representative while keeping the SMT query small) must keep them, so the assert is falsifiable. hevm finds and concretely validates the counterexample. If the filter ever unsoundly dropped same-mapping entries, this test would start "passing".

## Measurements (arm64 macOS, z3)

| Scenario | before | after |
|---|---|---|
| issue repro `prove_symbolic_key` (empty store) | hangs via echidna verification mode | 0.08s |
| unrelated-mapping read, 60 entries | 1.44s | 0.07s |
| unrelated-mapping read, 600 entries | >120s (killed) | ~1s |
| same-mapping read (must falsify) | cex validated | same cex, same time |

## Notes

- `keccakPreImgs` are only recorded during *concrete* execution (`OpSha3`), so the filter benefits exactly the flows that hit this bug: concrete deployment/`setUp()` followed by symbolic execution (`hevm test`, echidna verification mode). During pure symbolic execution the preimage map is empty and the filter keeps everything (status quo).
- The filter only applies to a bare `ConcreteStore`; reads over `SStore` write-chains are left to the existing simplification/decomposition machinery.
- Entries at scalar slots (no recorded preimage) are always kept. If that ever shows up as a bottleneck, dropping `k < 256` keys for keccak-shaped reads would mirror the existing small-lit-vs-keccak assumption; not needed for the cases here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ## Fixed
+- A symbolic-key `SLOAD` over a `ConcreteStore` is now resolved during execution
+  instead of being carried into a fork as an unresolved read: an empty store
+  reads 0 at every slot, and a symbolic mapping read filters out concrete
+  entries whose recorded keccak preimage identifies a different mapping
+  (keccak injectivity). Fixes non-termination/blowup on symbolic mapping reads
+  over concretely-initialized storage ([#1082](https://github.com/argotorg/hevm/issues/1082))
 - Cheatcode string arguments are now decoded leniently instead of crashing on
   invalid UTF-8: ABI `string` values are raw bytes, so e.g. `vm.setEnv`,
   `vm.envString` and `vm.label` no longer abort the whole run when handed a
--- a/src/EVM.hs
+++ b/src/EVM.hs
@@ -795,7 +795,9 @@ exec1 conf = do
 
                 symbolicRead :: EVM t () = if this.external
                   then accessStorage self x finalizeLoad
-                  else finalizeLoad $ Expr.readStorage' (Expr.concKeccakOnePass x) this.storage
+                  else let slot = Expr.concKeccakOnePass x
+                           preMap = Map.fromList [ (h, bs) | (bs, h) <- toList vm.keccakPreImgs ]
+                       in finalizeLoad $ Expr.readStorage' slot (Expr.filterStoreByReadSlot preMap slot this.storage)
 
                 concreteRead :: EVM t () = do
                   acc <- accessStorageForGas self (forceLit x)
--- a/src/EVM/Expr.hs
+++ b/src/EVM/Expr.hs
@@ -74,6 +74,7 @@ module EVM.Expr
     -- * Storage
   , readStorage'
   , readStorage
+  , filterStoreByReadSlot
   , writeStorage
   , concStoreContains
   , getAddr
@@ -809,8 +810,32 @@ simplifyReads = \case
 
 readStorage' :: Expr EWord -> Expr Storage -> Expr EWord
 readStorage' loc store = case readStorage loc store of
+                           -- an empty ConcreteStore is total: every slot, even a symbolic one,
+                           -- reads 0. Not in readStorage: accessStorage needs the Nothing/SLoad
+                           -- results there to drive lazy RPC slot fetching (partial caches).
+                           Just (SLoad _ (ConcreteStore m)) | Map.null m -> Lit 0
                            Just v -> v
                            Nothing -> Lit 0
+
+-- | The mapping slot (idx) of a symbolic mapping read keccak(key . idx), for both shapes hevm produces
+readMappingIdx :: Expr EWord -> Maybe W256
+readMappingIdx (MappingSlot idx _)
+  | BS.length idx == 64 = Just (W256 (word256 (BS.takeEnd 32 idx)))
+readMappingIdx (Keccak (CopySlice (Lit 0) (Lit 0) (Lit 0x40) (WriteWord (Lit 0) _ (ConcreteBuf idx)) _))
+  | BS.length idx >= 64 = Just (W256 (word256 (BS.take 32 (BS.drop 32 idx))))
+readMappingIdx _ = Nothing
+
+-- | Drop ConcreteStore entries that cannot alias a symbolic mapping read: a concrete key whose
+-- recorded keccak preimage identifies a different mapping slot cannot equal keccak(key . idx),
+-- by injectivity. Keys with unknown preimage are kept. `preMap` maps hash -> preimage.
+filterStoreByReadSlot :: Map.Map W256 ByteString -> Expr EWord -> Expr Storage -> Expr Storage
+filterStoreByReadSlot preMap slot (ConcreteStore m)
+  | Just readSlot <- readMappingIdx slot =
+      let keep k = case Map.lookup k preMap of
+                     Just bs | BS.length bs == 64 -> W256 (word256 (BS.takeEnd 32 bs)) == readSlot
+                     _ -> True
+      in ConcreteStore (Map.filterWithKey (\k _ -> keep k) m)
+filterStoreByReadSlot _ _ st = st
 
 -- | Reads the word at the given slot from the given storage expression.
 --
--- a/test/EVM/Expr/ExprTests.hs
+++ b/test/EVM/Expr/ExprTests.hs
@@ -63,6 +63,36 @@ storageTests = testGroup "Storage tests"
     , testCase "read-past-write" $ assertEqual errorMsg
         (Lit 0xab)
         (Expr.readStorage' (Lit 0x0) (SStore (Lit 0x1) (Var "b") (ConcreteStore $ Map.fromList [(0x0, 0xab)])))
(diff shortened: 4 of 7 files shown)


msooseth (approved):
LGTM

PR #1088, opened 2026-09-04
Test Foundry methods independently

elopez:
## Description

runForgeTest combines per-method outcomes with and. This is suitable for suite-level CLI reporting, but too weak for tests that expect every selected method to have a specific result. An aggregate False only proves that at least one method failed, while an empty selection yields (True, True).

Expose raw unit-test results to test helpers and assert non-empty, per-method outcomes for homogeneous regex and prefix selections. Use exact-name, single-result checks for individual and custom-option tests so similarly prefixed methods cannot affect results. Keep aggregate reduction only for the mixed-warning suite where it is intentional.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


msooseth (approved):
LGTM. Let's merge if you are happy with this!

PR #1089, opened 2026-09-04
perf: decode word256 with direct indexing instead of a cereal parser

elopez:
## Description

word256 sits on the PUSH decoding hot path (one call per executed PUSH). It ran a cereal Get parser over a padLeft copy of the immediate: two allocations plus parser state per PUSH. Assemble the four limbs by plain byte indexing instead, handling the left zero-padding arithmetically. Reads the first 32 bytes for oversized inputs, matching runGet's behavior on the padLeft result.

Old implementation is kept around for differential fuzzing.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):

msooseth:
Wow, very nice!

PR #1092, opened 2026-09-08
perf: decide JUMPDEST validity from opIxMap and drop the codeOps disassembly

elopez:
## Description

`isValidJumpDest` checked `codeOps ! (opIxMap ! x) == OpJumpdest`, which forces the
contract's `codeOps` vector on the first JUMP/JUMPI it executes. `mkCodeOps` decodes the
whole code into a lazy list of `LitByte`s, tuples and `Op`s, and each PUSH operand was a
thunk over the entire remaining list, so every contract that ran a jump paid for a decode and
then kept the whole decode list alive for as long as the `Contract` lived: the whole run for
the main contract, and via the trace for every created contract's init code.

`opIxMap` already answers the question. PUSH data bytes share the op index of their PUSH,
so byte `x` starts an instruction iff its op index differs from byte `x-1`'s, and a `0x5b`
at an instruction start is a JUMPDEST. With that, `codeOps` has no readers left so the second
commit deletes them.

## Checklist

- [ ] tested locally
- [ ] added automated tests
- [ ] updated the docs
- [ ] updated the changelog


blishko (approved):
Seems good to me! Maybe @msooseth can confirm?

Is this going to be true also with the upcoming instructions, like DUPN/SWAPN?

src/EVM.hs
```diff
@@ -3360,9 +3356,12 @@ isValidJumpDest vm x = let
...
        Nothing -> False
-       Just b -> 0x5b == b && (isJust $ contract.opIxMap VS.!? x)
-         && (isJust $ contract.codeOps V.!? (contract.opIxMap VS.! x))
-         && OpJumpdest == snd (contract.codeOps V.! (contract.opIxMap VS.! x))
+       Just b -> 0x5b == b && case contract.opIxMap VS.!? x of
+         Nothing -> False
+         -- PUSH data bytes share the op index of their PUSH, so byte x begins
+         -- an op (and is thus a real JUMPDEST) iff its op index differs from
+         -- the previous byte's
+         Just opIdx -> x == 0 || VS.unsafeIndex contract.opIxMap (x - 1) /= opIdx
```
msooseth: This is an obscure way of figuring this out? Can't we just check that it's  JUMPDEST? Why do this very  complicated comparison? I'm confused.

msooseth (changes requested):
Also, the comment should be shorter, once it's a simpler comparison :) Sorry, I'm confused as hell. Seem like it's trying to solve the Rubik's cube, but it's just a == JUMPDEST so it's kinda weird?

src/EVM.hs
blishko: I am not sure I understand the question correctly. Are you asking if just comparing the byte value with `0x5b` would be enough?

msooseth: OK, I pushed a change. Let me know if it makes sense. I think it's cleaner! I think it is :)

msooseth (approved):

PR #1096, opened 2026-09-22
Kill running SMT solvers on early abort

### LINKED ISSUE #938: Stop exploration as soon as a CEX is found, as per comment in #934 and as a follow-up to it

msooseth:
## Description

With `--early-abort`, SMT solvers that are still running are now killed once a counterexample is found, instead of running to completion. Fixes #938.

Also fixes a deadlock ("thread blocked indefinitely in an STM transaction") when paths are still queued at abort time, and a spurious "all branches reverted" warning after an early abort.

The tests now have a "hanging solver" fixture to be able to test early abort. This solver always hangs, so we can properly check without using hacks like a very complicated query.

## Checklist

- [x] tested locally
- [x] added automated tests
- [ ] updated the docs
- [x] updated the changelog


gustavo-grieco (commented):
Any reason not to have this as default? Maybe we could even remove the flag and always use it?

msooseth:
I was going to ask you if you are interested to have this as default :) Yes, I think that would be great. We can make it as default for the CLI. Not for API, I think, users can choose it? What do you think? And then we can have the `--no-early-abort` in CLI. What do you think?

gustavo-grieco:
I think making this the default makes a lot of sense, but allowing to use `--no-early-abort` mostly for regression testing.

